PR #4, opened 2018-07-05
Multiple product images

LukaszDygon:


.gitignore
```diff
@@ -1,26 +1,234 @@
...
+# Gemfile.lock, .ruby-version, .ruby-gemset
+
+# unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
+.rvmrc
 
-# Ignore all logfiles and tempfiles.
-/log/*
-/tmp/*
-!/log/.keep
-!/tmp/.keep
```
AnthonyPengelly: How come you've changed the gitignore so much? I think that it looks like the new one looks sensible. The only thing missing from the old one is un-ignoring these .keep files. I think there is some benefit in keeping them, just so everyone has the same folder structure

Gemfile.lock
```diff
@@ -44,27 +44,13 @@ GEM
       tzinfo (~> 1.1)
```
AnthonyPengelly: It looks like quite a few things have been removed from this lock file. If you try reinstalling all of the dependencies does it change it at all?

app/assets/stylesheets/products.scss
```diff
@@ -1,3 +1,11 @@
 // Place all the styles related to the products controller here.
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: http://sass-lang.com/
+@import 'slick';
+@import 'slick-theme';
+
+img {
```
AnthonyPengelly: These styles are applied very generally.. i.e. to all images. Could you make this more specific. e.g. give the image a class like .product__image

app/controllers/products_controller.rb
```diff
@@ -20,6 +20,7 @@ def show; end
   # GET /products/new
   def new
     @product = Product.new
+    @product.images.build
```
AnthonyPengelly: What does this line do? Provide an empty array of images or something?

app/views/products/show.html.erb
```diff
@@ -18,7 +18,11 @@
   Product
 </h1>
 
-<%= image_tag(@product.image_url, alt: @product.name) unless @product.image_url.nil? %>
+<div class="slider-nav">
+  <% @product.images.each do |image| %>
+    <img src="<%= image.url %>">
```
AnthonyPengelly: We should put the title as an "alt" attribute. Also, rails provides nice helpers for tags. e.g.
`<%= image_tag(image.url, alt: image.title) %>`

lib/tasks/rapex_import.rake
```diff
@@ -43,7 +43,7 @@ def create_product(notification)
     model: field_from_notification(notification, "type_numberOfModel"),
     batch_number: field_from_notification(notification, "batchNumber_barcode"),
     brand: brand(notification),
-    image_url: first_picture_url(notification),
```
AnthonyPengelly: Can probably bin "first_picture_url" now

lib/tasks/rapex_import.rake
```diff
@@ -80,6 +80,16 @@ def first_picture_url(notification)
   field_from_notification(notification, "pictures/picture")
 end
 
+def all_pictures(notification)
+  images = []
+  urls = notification.xpath("pictures/picture")
+  urls.each { |url|
```
AnthonyPengelly: I think rails likes you more if you avoid curly brackets over multiple lines for this sort of thing. e.g.
```
urls.each do |url|
  ...
end
```

lib/tasks/rapex_import.rake
```diff
@@ -80,6 +79,16 @@ def first_picture_url(notification)
...
+      clean_url = url.text.delete("\n") unless url.nil?
+      images.push(Image.create(url: clean_url))
```
AnthonyPengelly: Perhaps you could set the title to the name of the product or something?

db/migrate/20180704100621_create_versions.rb
```diff
@@ -0,0 +1,36 @@
+# This migration creates the `versions` table, the only schema PT requires.
```
AnthonyPengelly: Where have these migrations come from? are they for the audit stuff?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,68 @@
+<h3 class="heading-medium">
```
AnthonyPengelly: It looks like this change is from the audit table task

.gitignore
LukaszDygon: It was generated through gitignore.io, as per David's recommendation. I will ignore those

db/migrate/20180704100621_create_versions.rb
LukaszDygon: Yes they are for admin stuff. I realised I committed the next ticket once on this branch quite late. I could try to revert the changes but I think it will be a mess at this point

app/views/application/_history.html.erb
LukaszDygon: yes. As above

.gitignore
AnthonyPengelly: Awesome 👍

db/migrate/20180704100621_create_versions.rb
AnthonyPengelly: In theory it's probably best to try and revert it, so that we don't get stuff on int that hasn't been reviewed. And so that each piece of work is reasonably self contained. But if it's gonna get too tangled up then fair enough 😛

.gitignore
```diff
@@ -33,8 +33,10 @@ Temporary Items
 *.rbc
 capybara-*.html
 .rspec
-/log
-/tmp
+-/log/*
+-/tmp/*
+-!/log/.keep
+-!/tmp/.keep
```
AnthonyPengelly: It looks like "-" characters have crept their way into these lines of code 😄

LukaszDygon:
Moving to #6 .

PR #5, opened 2018-07-06
Mspsds 85 investigate audit options

LukaszDygon:
Auditing added to Products, Investigations and Actions.

Currently, changing photos doesn't register as a change. It's something to investigate further. I've tried adding a meta column but seemed to have no desired effect. May be good to revisit this once we understand the requirements better


app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    <div class="panel">
+        <% version.changeset.select { |k, v| !(v[0].blank? && v[1].blank?) }.each do |k, v| %>
+            <p><%= k %>: <%= v[0].to_s %> => <%= v[1].to_s %></p>
+        <% end %>
+    </div>
+  <% end %>
+</div>
+
+
+<style>
```
AnthonyPengelly: Could you put all of the styles and scripts in scss and js files respectively?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+<div>
+  <% versions.each do |version| %>
+    <button class="heading-small accordion">
+        <%= version.event.upcase %>
+        on <i><%= version.created_at.to_formatted_s(:db) %></i>
+        by <u><%= User.find(version.whodunnit).email ||= "anonymous" %></u>
+    </button>
+
+    <div class="panel">
+        <% version.changeset.select { |k, v| !(v[0].blank? && v[1].blank?) }.each do |k, v| %>
```
AnthonyPengelly: This is quite hard to follow. Perhaps renaming k and v would make it more obvious?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
+<h3 class="heading-medium">
+  History
+</h3>
+<div>
+  <% versions.each do |version| %>
+    <button class="heading-small accordion">
+        <%= version.event.upcase %>
+        on <i><%= version.created_at.to_formatted_s(:db) %></i>
+        by <u><%= User.find(version.whodunnit).email ||= "anonymous" %></u>
```
AnthonyPengelly: i and u tags seem quite unusual, are they necessary here? There's also an example of hidden text on this GovUK doc: http://govuk-elements.herokuapp.com/typography/

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    </div>
+  <% end %>
+</div>
+
+
+<style>
+/* Style the buttons that are used to open and close the accordion panel */
+.accordion {
+    background-color: #eee;
+    color: #444;
```
AnthonyPengelly: Once these are in SASS, we can use the GovUK colour variables: http://govuk-elements.herokuapp.com/colour/
(See buttons.scss) for an example

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    cursor: pointer;
```
AnthonyPengelly: I think if you try using the <details> and <summary> elements from the link I sent above, you might be able to strip out a lot of this functionality and styling?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    padding: 18px;
+    width: 100%;
+    text-align: left;
+    border: none;
+    outline: none;
+    transition: 0.4s;
+}
+
+/* Add a background color to the button if it is clicked on (add the .active class with JS), and when you move the mouse over it (hover) */
+.active, .accordion:hover {
```
AnthonyPengelly: In general, be careful with your class names, these styles are going to be applied site-wide, so it could have knock-on effects. I think we should consider following BEM conventions if you agree? http://getbem.com/naming/

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,70 @@
...
+/* Style the accordion panel. Note: hidden by default */
+.panel {
+    padding: 0 18px;
+    background-color: white;
+    display: none;
+    overflow: hidden;
+}
+</style>
+
+<script>
```
AnthonyPengelly: I think that JQuery provides helpers for most of things that are done in this section of JS. We should consider re-writing it (in a JS file). I'm happy to help out with that if you like? It could be that the <details> & <summary> stuff mentioned above will mean that we don't need this.

app/views/application/_history.html.erb
AnthonyPengelly: It's worth checking that we have the polyfill installed that it mentions at the bottom. Hopefully we've already got it 😄

app/views/products/show.html.erb
```diff
@@ -101,3 +105,10 @@
   <%= link_to 'Edit', edit_product_path(@product), :class => "button" %>
 </div>
 <%= link_to 'Back', products_path, :class => "link-back" %>
+
+
+<%= render 'history', versions: @product.versions.reverse %>
+
+<script>
+    productImagesToCarousel();
```
AnthonyPengelly: We should avoid script tags where possible and put them in js files. I assume that this is here so that it gets run after the elements have rendered? JQuery has a smart way of doing this. You can use `$(document).ready(function(){ ...do some stuff ... })`. So that it only fires once the document has done rendering 

https://api.jquery.com/ready/

db/migrate/20180704100621_create_versions.rb
```diff
@@ -0,0 +1,36 @@
+# This migration creates the `versions` table, the only schema PT requires.
```
AnthonyPengelly: Is this file auto generated by paper_trail?

.gitignore
```diff
@@ -1,26 +1,236 @@
...
+.AppleDesktop
+Network Trash Folder
+Temporary Items
+.apdisk
+
+### Rails ###
+*.rbc
+capybara-*.html
+.rspec
+-/log/*
```
AnthonyPengelly: The log files are currently not being ignored. I think once you remove the "-" signs from these four lines then that should be fixed 😄

db/migrate/20180704100621_create_versions.rb
LukaszDygon: yes

app/views/products/show.html.erb
LukaszDygon: I've tried it before and tested it now. For some reason the expected behaviour must be overwritten somewhere because a second after navigating to the view it will destroy the carousel and just display images.

app/views/application/_history.html.erb
LukaszDygon: Thank you. I probably should read throught the site. It would have saved me some time int his case

app/views/products/show.html.erb
AnthonyPengelly: Ah my apologies! Rails is using a framework called turbolinks https://thoughtbot.com/upcase/videos/turbolinks
What this does, is make all of the HTML calls using AJAX to make it a sort of SPA. Therefore, the "ready" event is only fired when you first load the site, not when you navigate to a new page.

There is another event that is setup to handle these turbolink navigations called turbolinks:load e.g.
`$(document).on("turbolinks:load", function() { ... });`

There is an example of this in investigations.js

LukaszDygon: Thank you for explaining this. Makes sense why now. This fixed the problem

broder:
I'll put off reviewing this one until #6 is merged if that's okay.

app/models/product.rb
```diff
@@ -11,6 +12,9 @@ class Product < ApplicationRecord
   has_many :images, dependent: :destroy, inverse_of: :product
 
   accepts_nested_attributes_for :images, reject_if: :all_blank, allow_destroy: true
+
+  has_paper_trail
+  # has_paper_trail meta: { images: Proc.new { |product| product.images.count } }
```
broder: Do you think that we'll uncomment this line of code at some point?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,19 @@
+<h3 class="heading-medium">
+    History
+</h3>
+
+<details>
+  <% versions.each do |version| %>
+    <summary><span class="summary">
+        <%= version.event.upcase %>
+        on <%= version.created_at.to_formatted_s(:db) %>
+        by <%= User.find(version.whodunnit).email ||= "anonymous" %>
```
broder: Is `version.whodunnit` the user's GUID?

app/views/application/_history.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    </span></summary>
+
+    <div class="panel panel-border-narrow">
+        <% version.changeset.select { |_, change| !(change[0].blank? && change[1].blank?) }.each do |field, change| %>
```
broder: Is there a way that we could make it more obvious that `change[0]` is the old version and `change[1]` is the new version?

app/views/activities/show.html.erb
```diff
@@ -49,3 +49,5 @@
   </div>
 <% end %>
 <%= link_to 'Back', investigation_activities_path(@activity.investigation), :class => "link-back" %>
+
+<%= render 'history', versions: @activity.versions.reverse %>
```
broder: Good use of partials! I wonder if we should do the `.reverse` in the partial rather than in each of the views?

db/migrate/20180706160653_create_versions.rb
```diff
@@ -0,0 +1,36 @@
+# This migration creates the `versions` table, the only schema PT requires.
+# All other migrations PT provides are optional.
```
broder: Are there any other migrations than the 2 we have?

LukaszDygon: maybe in the future. This comment was auto-generated. Is it sensible to leave it here or better to remove it?

app/views/application/_history.html.erb
LukaszDygon: Yes. It's set by Paper Trail, hence the weird name instead of user_id.

app/views/application/_history.html.erb
```diff
@@ -2,18 +2,19 @@
...
     <summary><span class="summary">
         <%= version.event.upcase %>
         on <%= version.created_at.to_formatted_s(:db) %>
         by <%= User.find(version.whodunnit).email ||= "anonymous" %>
     </span></summary>
 
     <div class="panel panel-border-narrow">
-        <% version.changeset.select { |_, change| !(change[0].blank? && change[1].blank?) }.each do |field, change| %>
-            <p><%= field %>: <%= change[0].to_s %> => <%= change[1].to_s %></p>
+        <% version.changeset.select { |_, (old, new)| !(old.blank? && new.blank?) }.each do |field, (old, new)| %>
```
broder: Nice!

PR #6, opened 2018-07-06
MSPSDS-90: multiple product images

LukaszDygon:
Added multiple image URLs for products.


app/assets/javascripts/products.js
```diff
@@ -0,0 +1,11 @@
+
+$(window).ready( function() {
```
broder: Did Ant mention that this should be done with `$( document ).on('turbolinks:load')`?

PR #7, opened 2018-07-09
MSPSDS-137: RAPEX case imports

LukaszDygon:
Rake task now creates the products once if they have a barcode ('gtin').
The investigation and actions are generated from the RAPEX reports.
The data_import:delete_rapex task deletes the content generated by the data_import:rapex task.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -2,5 +2,5 @@ class Activity < ApplicationRecord
   default_scope { order(created_at: :desc) }
   belongs_to :investigation
   belongs_to :activity_type
-  belongs_to :user
+  belongs_to :user, optional: true
 end
--- a/lib/tasks/rapex_import.rake
+++ b/lib/tasks/rapex_import.rake
@@ -16,14 +16,13 @@ namespace :data_import do
     end
   end
 
+  # this will delete products even if they are used by an investigation not from RAPEX
   task delete_rapex: :environment do
-    pod_product_count = 9000
     RapexImport.all.destroy_all
-    pod_products = Product.last(pod_product_count)
-    Product.where.not(id: pod_products.collect(&:id)).destroy_all
-    InvestigationProduct.destroy_all
-    Investigation.destroy_all
-    Activity.destroy_all
+    Product.where(source: "Imported from RAPEX").destroy_all
+    destroyed_investigation_ids = Investigation.where(source: "Imported from RAPEX").destroy_all.collect(&:id)
+    InvestigationProduct.where(investigation_id: destroyed_investigation_ids).destroy_all
+    Activity.where(investigation_id: destroyed_investigation_ids).destroy_all
   end
 end
 
--- a/lib/tasks/rapex_import.rake
+++ b/lib/tasks/rapex_import.rake
@@ -28,14 +28,15 @@ namespace :data_import do
 end
 
 def import_report(report)
+  date = DateTime.strptime(report.xpath("publicationDate").text, "%e/%m/%Y")
   reference = report.xpath("reference").text
   puts "Importing #{reference}"
   url = report.xpath("URL").text.delete("\n")
   notifications(url).each do |notification|
-    investigation = create_investigation notification
+    investigation = create_investigation notification, date
     product = create_product notification
     create_investigation_product investigation, product
-    create_activity notification, investigation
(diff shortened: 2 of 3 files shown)


app/models/activity.rb
```diff
@@ -2,5 +2,5 @@ class Activity < ApplicationRecord
   default_scope { order(created_at: :desc) }
   belongs_to :investigation
   belongs_to :activity_type
-  belongs_to :user
+  belongs_to :user, optional: true
```
broder: Maybe we should change this to source (as part of your next ticket) - what do you think?

lib/tasks/rapex_import.rake
```diff
@@ -16,26 +16,36 @@ namespace :data_import do
     end
   end
 
+  # this will delete products even if they are used by an investigation not from RAPEX
   task delete_rapex: :environment do
-    pod_product_count = 9000
     RapexImport.all.destroy_all
-    pod_products = Product.last(pod_product_count)
-    Product.where.not(id: pod_products.collect(&:id)).destroy_all
+    Product.where(source: "Imported from RAPEX").destroy_all
```
broder: Does the ordering of this actually work? I'd expect that you couldn't delete all `Product`s before you delete all `InvestigationProduct`s

lib/tasks/rapex_import.rake
```diff
@@ -47,6 +57,34 @@ def create_product(notification)
   )
 end
 
+def create_investigation(notification, date)
+  Investigation.create(
+    description: field_from_notification(notification, "description"),
+    is_closed: true,
+    severity: 1,
```
broder: Should we map this to `level` in the RAPEX fields? We should maybe put a `TODO` in all the mapping code relating to MSPSDS-131 to update this mapping when we update the case attributes etc.

lib/tasks/rapex_import.rake
```diff
@@ -47,6 +57,34 @@ def create_product(notification)
...
+    created_at: date,
+    updated_at: date,
```
broder: I think it's a great idea to set `created_at` to the RAPEX `publicationDate` - should we set `updated_at` to now though?

app/models/activity.rb
LukaszDygon: yes definitely. I put it off since it seemed to be outside of the ticket's scope

lib/tasks/rapex_import.rake
LukaszDygon: Maybe the restrictions in place are wrong, but I seemed to be able to delete the imported products like this. I will double check.

lib/tasks/rapex_import.rake
LukaszDygon: I set it like so with regards to the description on MSPSDS-137 JIRA ticket. It seems a little confusing to set those two fields, since they are usually just autogenerated, so probably worth defining what they are meant to represent in the system. (i.e. is it more important to know when entry was added/edited in the system or in the external system/report). Perhaps it should belong in Source as well.

lib/tasks/rapex_import.rake
LukaszDygon: I believe the level is either Serious Risk or Other Risk. Is it sensible to attach a numeric values to those? Changing the severity to something other like an enum may not be what we want, right?

lib/tasks/rapex_import.rake
broder: Fair enough

lib/tasks/rapex_import.rake
broder: We could do something like `level == "Serious Risk" ? 1 : 2` for now and then put a `TODO` for the correct mapping.

lib/tasks/rapex_import.rake
broder: I would guess that it's more important to know when the entry was added to the external system (as opposed to when we imported it). Fine to leave this as is for now 👍

lib/tasks/rapex_import.rake
```diff
@@ -57,13 +57,14 @@ def create_product(notification)
   )
 end
 
+# TODO: change 'severity' to something more sensible based on requirements
```
broder: It's generally helpful to reference a JIRA ticket with `TODO`s - in this case it would be `# TODO MSPSDS-131: change...`. Sorry for another minor change!

PR #8, opened 2018-07-10
MSPSDS-154: Improved source in investigation and product

LukaszDygon:
Investigations and Products now use a more appropriate concept of a source


LukaszDygon:
Closingmoved to #9

PR #9, opened 2018-07-11
MSPSDS-154: Improved source in investigation and product

LukaszDygon:


app/views/investigations/show.html.erb
```diff
@@ -66,7 +66,7 @@
   Source
 </h4>
 <p>
-  <%= @investigation.source %>
+  <%= @investigation.source.show unless !@investigation.source %>
```
broder: Is there a case where `@investigation.source` wouldn't be set?

app/models/investigation.rb
```diff
@@ -5,6 +5,7 @@ class Investigation < ApplicationRecord
   has_many :activities, dependent: :destroy
   belongs_to :assignee, class_name: "User", optional: true
   has_one_attached :image
+  belongs_to :source
```
broder: I wonder if this should instead be something like `has_one :source, as: :sourceable`. Then the source would have `belongs_to :sourceable, polymorphic: true`.

You can read more about it here: https://launchschool.com/blog/understanding-polymorphic-associations-in-rails or https://launchschool.com/blog/understanding-polymorphic-associations-in-rails

LukaszDygon:
I will update the rake tasks as well before merging

app/views/investigations/show.html.erb
LukaszDygon: I think I changed it because the tests were failing and it fixed some of them, but after the change that defines the product source in .yml, it should always have a source, so I will revert it

PR #11, opened 2018-07-12
MSPSDS-156: investigate exports

LukaszDygon:
Added spreadsheet export of investigations.
Added ability to generate investigation and product pdfs


app/controllers/investigations_controller.rb
```diff
@@ -55,6 +63,16 @@ def update_assignee
     end
   end
 
+  # POST /investigations/report
```
broder: Should we make this a GET request? Also, we could merge this functionality with the `index` route, then we'd get pagination as well - what do you think?

.rubocop.yml
```diff
@@ -34,6 +34,9 @@ Metrics/MethodLength:
 Layout/EndOfLine:
   EnforcedStyle: lf
 
+Metrics/ClassLength:
```
broder: Do we still need this?

app/views/investigations/index.html.erb
```diff
@@ -40,3 +40,5 @@
 </table>
 <br />
 <%= will_paginate @investigations %>
+
+<%= link_to "Export as spreadsheet", investigations_path(format: :xlsx), method: :get %>
```
broder: I think it'll be `method: :get` by default?

.rubocop.yml
LukaszDygon: Yes

app/assets/stylesheets/investigations.scss
```diff
@@ -1,3 +1,30 @@
 // Place all the styles related to the Investigations controller here.
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: http://sass-lang.com/
+@media print
+{
+    #pager,
+    form,
```
AnthonyPengelly: I would be careful applying such general styles, such as _all_ forms are hidden when printing. IMO it would be better to say all forms within the investigations "show" page or something

PR #13, opened 2018-07-16
MSPSDS-97 anti virus

AnthonyPengelly:


app/analyzers/anti_virus_analyzer.rb
```diff
@@ -0,0 +1,20 @@
+class AntiVirusAnalyzer < ActiveStorage::Analyzer
+  def self.accept?(_blob)
+    true
+  end
+
+  def metadata
+    download_blob_to_tempfile do |file|
+      is_safe = Clamby.safe? file.path
+      purge_blob unless is_safe
+      { safe: is_safe }
```
broder: Neat! When we have more analysers, I wonder if it would make sense for the `MasterAnalyzer` to mark the file as `processed` or something, rather than having to write code to deal with the result of every analyser individually.

AnthonyPengelly: As it happens, the base analyzer adds "analyzed: true" to the metadata :)

PR #14, opened 2018-07-17
MSPSDS-89 Use active storage to generate thumbnails for uploaded images

AnthonyPengelly:


AnthonyPengelly:
Active storage lazily creates a thumbnail when it is first requested. It then stores this in S3 so next time it doesn't need to

app/views/investigations/show.html.erb
```diff
@@ -22,7 +22,7 @@
   <% if @investigation.image.metadata["safe"].nil? %>
     <p>Still checking upload for viruses</p>
   <% elsif @investigation.image.metadata["safe"] %>
-    <%= image_tag(url_for(@investigation.image), alt: "Investigation") %>
+    <%= image_tag @investigation.image.variant(resize: "100x100"), alt: "Investigation" %>
```
broder: So does this block the first time that it's requested?

AnthonyPengelly: It only blocks the image loading. So when it runs this code, it doesn't do anything, it just puts in an image src in the html. When the browser tries to load the image, _then_ it builds the variant. (on my PC it only too 24ms to convert it)

broder: That is really cool!

PR #15, opened 2018-07-20
Mspsds 190 show duplicate products

AnthonyPengelly:


app/assets/javascripts/products.js
```diff
@@ -8,4 +8,49 @@ $(document).on("turbolinks:load", function() {
...
+}
+
+function buildQuery() {
+    const query = {};
+    const q = $('.new-product-page .search-term:not(#gtin-input)')
+        .map(function() {
+            return $(this).val();
+        })
+        .get()
+        .filter(function(searchTerm) {
```
broder: Is this filter doing anything?

app/assets/javascripts/products.js
```diff
@@ -8,4 +8,49 @@ $(document).on("turbolinks:load", function() {
...
         adaptiveHeight: true,
     });
+
+    handleSearchInput();
 });
+
+function handleSearchInput() {
+    let debounceTimeout = null;
+    let searchRequest = null;
+    $('.new-product-page .search-term').on('keyup change', function() {
```
broder: Should we use a library for the debounce functionality?

app/controllers/products_controller.rb
```diff
@@ -72,6 +75,27 @@ def destroy
 
   private
 
+  # If the user supplies a barcode and it matches, then just return that.
+  # Otherwise use the general query param
+  def advanced_product_search
+    products = search_for_gtin
```
broder: Would something like
```
if params[:gtin].present?
  # search with gtin
else
  search_for_products
```
be neater? Defining `products` multiple times is a bit confusing.

app/controllers/products_controller.rb
```diff
@@ -6,11 +6,14 @@ class ProductsController < ApplicationController
...
-    @products = if params[:q].blank?
-                  Product.paginate(page: params[:page], per_page: 20)
-                else
-                  Product.search(params[:q]).paginate(page: params[:page], per_page: 20).records
-                end
+    @products = search_for_products
+  end
+
+  # GET /products/table
+  def table
```
broder: Did you have any thoughts about if / how pagination could work with the AJAX that we've got?

config/routes.rb
```diff
@@ -8,6 +8,7 @@
     end
     resources :activities, shallow: true
   end
+  get "/products/table" => "products#table"
```
broder: Does something like
```
resources :products do
  collection do
    get :table
  end
end
```
work?

app/assets/javascripts/products.js
AnthonyPengelly: We could do, it felt fairly straight forward so I just rolled it myself. The best suggestions for libraries were lodash and underscore which are both quite chunky. Do you think I should go for it anyway?

app/assets/javascripts/products.js
AnthonyPengelly: It's filtering out any undefined/empty values. (I will remove the log though!)

app/controllers/products_controller.rb
AnthonyPengelly: Passing in a page number query param would actually, so we could fairly easily add that on the frontend. However, my gut feel is that it's unlikely that any more than the top few are going to be relevant to the user

app/controllers/products_controller.rb
AnthonyPengelly: I agree that it is currently a bit confusing. However, the functionality is actually a bit different. it's actually more like
```
if params[:gtin].present?
  # try and search with gtin

if params[:q].present? && (there were no results for gtin || !params[:gtin].present?)
  # Search using the query string

else
  # Don't search
```
Does that make sense? I'll have a think about if I can improve the readability though

config/routes.rb
AnthonyPengelly: Works beautifully - thanks!

app/assets/javascripts/products.js
broder: I'm happy to leave it for now. I was hoping that jQuery would have functionality to do this but looks like you need a plugin (https://github.com/cowboy/jquery-throttle-debounce) :/

app/controllers/products_controller.rb
broder: Yeah, hopefully they'll only need the first few results so we should leave it for now

app/assets/javascripts/products.js
broder: Cool!

app/controllers/products_controller.rb
```diff
@@ -78,9 +78,10 @@ def destroy
   # If the user supplies a barcode and it matches, then just return that.
   # Otherwise use the general query param
   def advanced_product_search
-    products = search_for_gtin
-    products = search_for_products if products.blank? && params[:q].present?
-    products
+    gtin_search_results = search_for_gtin if params[:gtin].present?
+    # if there was no GTIN param or there were no results for the GTIN search
+    basic_search_results = search_for_products if gtin_search_results.blank? && params[:q].present?
+    basic_search_results || gtin_search_results || []
```
broder: Nice!

PR #16, opened 2018-07-24
Mspsds 181 case details

AnthonyPengelly:
Improve the case details page. Still would require some work to get it to exactly match the wireframes

### DIFF
--- a/db/migrate/20180723103012_update_investigation_from_design_sprint.rb
+++ b/db/migrate/20180723103012_update_investigation_from_design_sprint.rb
@@ -1,7 +1,10 @@
 class UpdateInvestigationFromDesignSprint < ActiveRecord::Migration[5.2]
   def change
-    add_column :investigations, :title, :string, null: false, default: "Investigation"
+    add_column :investigations, :title, :string
     add_column :investigations, :risk_notes, :text
     remove_column :investigations, :severity, :string
+
+    Investigation.update_all(title: "Investigation")
+    change_column :investigations, :title, :string, null: false
   end
 end
--- a/db/migrate/20180723103012_update_investigation_from_design_sprint.rb
+++ b/db/migrate/20180723103012_update_investigation_from_design_sprint.rb
@@ -1,0 +1,7 @@
+class UpdateInvestigationFromDesignSprint < ActiveRecord::Migration[5.2]
+  def change
+    add_column :investigations, :title, :string, null: false, default: "Investigation"
+    add_column :investigations, :risk_notes, :text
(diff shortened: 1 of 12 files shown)


db/migrate/20180723103012_update_investigation_from_design_sprint.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateInvestigationFromDesignSprint < ActiveRecord::Migration[5.2]
+  def change
+    add_column :investigations, :title, :string, null: false, default: "Investigation"
```
broder: I wonder if we should add a nullable `title` column, set the title to all of the existing investigations to "Investigation" and the make the column non-nullable. Rather than having a database default.

lib/tasks/rapex_import.rake
```diff
@@ -7,7 +7,7 @@ namespace :data_import do
   task rapex: :environment do
     weekly_reports = rapex_weekly_reports
     previously_imported_reports = RapexImport.all
-    weekly_reports.reverse_each do |report|
+    weekly_reports[0..2].reverse_each do |report|
```
broder: Did you mean to commit this? It might actually make the int site less crowded but maybe we need lots of data to check search properly...

db/schema.rb
```diff
@@ -73,10 +73,11 @@
   create_table "investigations", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
     t.text "description"
     t.boolean "is_closed"
-    t.integer "severity"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "assignee_id"
+    t.string "title", default: "Investigation", null: false
```
broder: Should we update the investigations list to display this title rather than the GUID?

AnthonyPengelly: I wondered that, but the GUID was used in the wireframes, so I left it

lib/tasks/rapex_import.rake
AnthonyPengelly: I didn't mean to! Great catch

db/schema.rb
broder: Hmm, does the investigation definitely have a title or is it pulled in from the related product? Might be worth checking with Paul?

AnthonyPengelly: I've just spoken to the guys - for now they want a title. Though that may get removed when they test it

broder: :+1:

broder:
LGTM 👍

PR #18, opened 2018-07-27
Mspsds 193 extra business features

AnthonyPengelly:
Mostly finished, still need to look into postcode APIs

### DIFF
--- a/app/controllers/businesses_controller.rb
+++ b/app/controllers/businesses_controller.rb
@@ -1,5 +1,3 @@
-require_relative "../clients/companies_house_client"
-
 class BusinessesController < ApplicationController
   include BusinessesHelper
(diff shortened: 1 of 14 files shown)


app/controllers/businesses_controller.rb
```diff
@@ -92,10 +108,13 @@ def respond_to_business_creation
     end
   end
 
+  def filter_out_existing_businesses(businesses)
+    businesses.reject { |business| Business.exists?(company_number: business[:company_number]) }
```
broder: Would adding an index on `company_number` improve the performance of this?

AnthonyPengelly: There is in fact an index on company_number :) (The migration file is in this review)

PR #21, opened 2018-08-03
[WIP] MSPSDS-209 potential duplicates

AnthonyPengelly:
Currently doesn't work for RAPEX import script. I'm investigating this at the moment


app/models/potential_product_duplicate.rb
```diff
@@ -8,9 +8,9 @@ class PotentialProductDuplicate < ApplicationRecord
   # The elasticsearch score does not have a hard upper limit, but anything over
   # 25 is going to be very similar
```
broder: Is this comment still correct?

AnthonyPengelly: ... no

PR #22, opened 2018-08-07
MSPSDS-131 update product model

AnthonyPengelly:


db/migrate/20180807110434_change_product_data_structure.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeProductDataStructure < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :products, :image_url, :string
+    remove_column :products, :mpn, :string
+    rename_column :products, :purchase_url, :url_reference
+
+    add_column :products, :serial_number, :string
+    add_column :products, :manufacturer, :string
+    add_column :products, :country_of_origin, :string
```
broder: Should we try and link this to the GOV.UK Register for countries somehow?

AnthonyPengelly: Good plan, I will do this on a new branch

PR #23, opened 2018-08-08
MSPSDS-219 consume countries register

AnthonyPengelly:


app/helpers/countries_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    "node_modules/govuk-country-and-territory-autocomplete/dist/location-autocomplete-canonical-list.json".freeze
+
+  # JSON is of the form [["Abu Dhabi", "territory:AE-AZ"], ["Afghanistan", "country:AF"]]
+  def country_from_code(code)
+    country = all_countries.find { |c| c[1] == code }
+    country && country[0]
+  end
+
+  def all_countries
+    JSON.parse(File.read(PATH_TO_COUNTRIES_LIST))
```
broder: Should we cache this so we don't have to read the file every time?

PR #24, opened 2018-08-08
MSPSDS-217 Use GOVUK elasticsearch instead of AWS

AnthonyPengelly:


config/elasticsearch.yml
```diff
@@ -0,0 +1,15 @@
...
+      :timeout: 5
+
+development:
+  <<: *default
+
+test:
+  <<: *default
+
+production:
+  :url: <%= ENV['VCAP_SERVICES'] && JSON.parse(ENV['VCAP_SERVICES'])["elasticsearch"][0]["credentials"]["uri"] %>
```
broder: Ew - would it be worth sending an email to PaaS support to ask if there's another way? Feels like our experience might be useful to them (whether or not there's a better way of doing this).

docs/Deploy from scratch.md
```diff
@@ -35,8 +35,10 @@ Create an S3 bucket named `int-mspsds`. This bucket needs public _read_ access.
 
 ## Elastic Search
 
-Create an AWS Elasticsearch domain called `int-mspsds`, choosing a suitable instance size.
-For access, choose public, and then allow your int-mspsds user to access the resource (by pasting their ARN).
+Create an empty elasticsearch instance in the `int` space
+
+    cf marketplace -s elasticsearch
+    cf create-service elasticsearch small-ha-6.x mspsds-elasticsearch
```
broder: Is there a non-HA option? I feel like our service probably doesn't need high availability for now.

AnthonyPengelly: No, this is the only option (or small-ha-5.x). It is on the free tier though

config/elasticsearch.yml
AnthonyPengelly: Will do!

PR #26, opened 2018-08-10
MSPSDS-180 Redis & Sidekiq

AnthonyPengelly:


.travis.yml
```diff
@@ -19,6 +19,7 @@ cache:
 jobs:
   include:
   - stage: test
+    if: branch != MSPSDS-180_redis
```
broder: I'm guessing that we won't need this anymore?

.travis.yml
```diff
@@ -35,13 +36,19 @@ jobs:
       - docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
   - stage: deploy-to-int
     if: branch = master AND type != pull_request
-    script: SPACE=int ./infrastructure/scripts/deploy.sh
+    script: SPACE=int ./infrastructure/scripts/deploy-app.sh
+  - if: branch = master AND type != pull_request
```
broder: Do we need to have this conditional twice?

config/redis.yml
```diff
@@ -0,0 +1,12 @@
...
+  host: 'redis'
+
+development:
+  <<: *default
+
+test:
+  <<: *default
+
+production:
+  url: <%= ENV['VCAP_SERVICES'] && JSON.parse(ENV['VCAP_SERVICES'])["redis"][0]["credentials"]["uri"] %>
```
broder: Did you think about using https://github.com/cloudfoundry/cf-app-utils-ruby?

docker-compose.yml
```diff
@@ -11,22 +11,43 @@ services:
       - "ES_JAVA_OPTS=-Xms750m -Xmx750m"
     volumes:
       - elasticsearch-volume:/usr/share/elasticsearch/data
+  redis:
+    image: redis
+    command: ["redis-server", "--appendonly", "yes"]
```
broder: Does `appendonly` mean that the task list never gets cleared?

.travis.yml
AnthonyPengelly: I believe so. They are actually different stages (distinguished by the "-" on this line). When a stage isn't given a name, it takes the name of the previous one - i.e. deploy-to-int

PR #31, opened 2018-08-16
MSPSDS-224: Autocomplete email address

LukaszDygon:


app/assets/javascripts/investigations.js
```diff
@@ -1,6 +1,9 @@
 $(document).on("turbolinks:load", function() {
     addSelect2AjaxSearchToElement($(".js-investigation-products"), "/products", mapAjaxDataToProduct);
     addSelect2AjaxSearchToElement($(".js-investigation-businesses"), "/businesses", mapAjaxDataToBusiness);
+    addSelect2AjaxSearchToElement($(".js-assignee-email"), "/businesses", mapAjaxDataToEmail);
```
tom-bell-softwire: Is this AJAX call needed if you are already populating the select options on the form?

broder: I wonder if we should do it via AJAX rather that calling `User.all` - would there be performance issues if we rolled out to 1000s of users?

PR #32, opened 2018-08-16
MSPSDS-225: added automatic activity types and activity on assignment

LukaszDygon:


app/controllers/investigations_controller.rb
```diff
@@ -110,6 +110,7 @@ def destroy
   def save_and_respond(notice)
```
tom-bell-softwire: This method is called by a few actions (close/re-open case, in addition to update assignee). Do we want to record the "Assign" activity for all of them?

app/controllers/investigations_controller.rb
```diff
@@ -110,6 +110,7 @@ def destroy
...
     respond_to do |format|
       if @investigation.save
+        record_assignment(@investigation)
```
tom-bell-softwire: The `save_and_respond` method is called by a few actions (close/re-open case, in addition to update assignee). Do we want to record an "Assign" activity for all of them?

LukaszDygon: My bad. I fixed it now 👍

app/controllers/investigations_controller.rb
```diff
@@ -128,6 +129,15 @@ def set_investigation
     @investigation = Investigation.find(params[:id])
   end
 
+  def record_assignment(investigation)
+    Activity.create(
+      source: UserSource.new(user: current_user),
+      investigation: investigation,
+      activity_type: ActivityType.find(name: "assign"),
```
broder: I wonder if the `activity_type` field should just be an enum (https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) which would make this a little less flaky - @AnthonyPengelly, was there a specific reason for having it as a separate DB table?

AnthonyPengelly: No, I guess the original though process was to avoid coupling of numbers in the db and our code, and allow adding new types dynamically. But an enum would probably be better now by the look of it - sounds good! 👍

app/controllers/investigations_controller.rb
```diff
@@ -128,6 +129,15 @@ def set_investigation
     @investigation = Investigation.find(params[:id])
   end
 
+  def record_assignment(investigation)
+    Activity.create(
```
AnthonyPengelly: You should be able to do something like `investigation.activities.create(...` and then you don't need to set the "investigation" field of the activity

db/migrate/20180817135353_drop_activity_types.rb
```diff
@@ -1,7 +1,7 @@
 class DropActivityTypes < ActiveRecord::Migration[5.2]
   def change
     remove_reference :activities, :activity_type, index: true
-    add_column :activities, :activity_type, :integer, default: 0
+    add_column :activities, :activity_type, :integer, null: false
```
broder: Should there be a `schema.db` change from this? Also, I think if you ran this migration on a database with existing activities, it would fail as you would be adding a non-null column. I think something like
```
rename_column :activities, :activity_type, :activity_type_temp
add_column :activities, :activity_type, :integer
Activity.all.each do |account|
  activity.activity_type_temp = based on existing ActivityType from activity_type_temp (or just set it to whatever
  activity.save!
end

remove_reference :activities, :activity_type_temp, index: true 
change_column_null :activities, :activity_type, false
```
might work?

PR #33, opened 2018-08-17
MSPSDS-222 UX for adding products to investigations

AnthonyPengelly:


AnthonyPengelly:
Yeah, the investigations/products_controller has made a lot of sense. There's slightly more duplication (not too much), but way fewer weirdnesses to get it to work

app/assets/javascripts/products_search.js
```diff
@@ -0,0 +1,22 @@
...
+    var q = $('.products-search-form .search-term:not(#gtin-input)')
+        .map(function() {
+            return $(this).val();
+        })
+        .get()
+        .filter(function(searchTerm) {
+            return searchTerm;
+        })
+        .map(function(searchTerm) {
+            return searchTerm + "*"
```
broder: Lukasz talked about adding fuzziness (https://www.elastic.co/guide/en/elasticsearch/reference/6.3/common-options.html#fuzziness) to the elasticsearch options - would that work here or do we need the `*`?

app/views/investigations/products/index.html.erb
```diff
@@ -0,0 +1,40 @@
...
+                </li>
+        </ol>
+    </div>
+
+    <%= link_to 'Back', @investigation, class: "govuk-back-link link-back" %>
+
+    <h1 class="heading-xlarge">Add product to investigation</h1>
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-one-half">
+            <%= form_with(model: @product, url: new_investigation_product_path(@investigation), local: true, method: :get) do |form| %>
```
broder: What does `local: true` do?

app/assets/javascripts/products_search.js
AnthonyPengelly: This is doing more of a look-ahead. i.e. they've not finished typing the word, which is slightly different to fuzzy searching. I think that the product searching will need to be refined anyway when we sort out advanced search, so this should probably be considered at that point. I'll add a TODO

app/views/investigations/products/index.html.erb
AnthonyPengelly: "By default form submits are remote and unobstrusive XHRs. Disable remote submits with local: true."
When you rails g scaffold something it sets local: true on all of the forms, which is how it has ended up as the default for us. I guess if we removed it, it would tie into the turbolinks stuff. Any strong preference?

broder: No, I was just curious!

PR #36, opened 2018-08-22
MSPSDS-226 Update case creation page

AnthonyPengelly:


lib/tasks/rapex_import.rake
```diff
@@ -114,7 +114,7 @@ def brand(notification)
   brand
 end
 
-def risk_notes(notification)
+def risk_overview(notification)
   [
     field_from_notification(notification, "level"),
```
broder: Could this populate our `risk_level` field? We could maybe put a `TODO` in so we remember to think about it later.

AnthonyPengelly: I've left a TODO :) I don't want to spend time on it now, cause I think that whether we will have a risk level etc. is still an open question

PR #37, opened 2018-08-22
MSPSDS-231: Image management functionality for investigations and products

tom-bell-softwire:


app/assets/stylesheets/images.scss
```diff
@@ -0,0 +1,61 @@
+// Place all the styles related to the Images controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
+@import "buttons";
+@import "colours";
+
+.image-with-preview {
+  background-color: $grey-1;
```
broder: Are these styles common between the three image states? Could we move them into the ```.image-list {
  .image-with-preview,
  .image-without-preview,
  .image-processing-placeholder {``` bit below?

app/models/rapex_image.rb
```diff
@@ -1,4 +1,4 @@
-class Image < ApplicationRecord
+class RapexImage < ApplicationRecord
```
broder: Should we put a `TODO` to remove this and upload the RAPEX images into our image storage or do you think we should just keep hotlinking to RAPEX?

app/views/investigations/show.html.erb
```diff
@@ -68,9 +68,9 @@
         <h2 class="govuk-heading-m">
           Images
         </h2>
-        <% if product.images.any? %>
```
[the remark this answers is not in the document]
broder: Yeah, I think so - might be related to the comment about removing `RapexImage`s above as well

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 # rubocop:disable Metrics/BlockLength
 Rails.application.routes.draw do
-  resources :investigations do
+  concern :image_attachable do
+    resources :images
```
broder: Neat!

app/models/rapex_image.rb
tom-bell-softwire: Good call – I'll add a TODO (and related JIRA ticket).

I'd prefer to replace this with using stored images across the board. It just makes the logic in the views simpler.

app/views/investigations/show.html.erb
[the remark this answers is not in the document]
tom-bell-softwire: Cool. I'd add a TODO.

app/assets/stylesheets/images.scss
tom-bell-softwire: These are used in the `_image` and `_image_tag` partials, which are used in several places (investigation and product pages, plus the various manage image pages).

They're then extended within `.image-list` and `investigations.scss` to specify the thumbnail size for the relevant page. However, there might be a neater way to achieve that.

I think I need to tweak some of the styling slightly anyway, so will have another look as part of that.

PR #39, opened 2018-08-22
MSPSDS-262 update case page design

AnthonyPengelly:
Next I plan to pick up:
* Manage products page
* Manage businesses page
* Page to list all activities you can do


app/controllers/activities_controller.rb
```diff
@@ -28,7 +28,10 @@ def edit; end
   def create
     respond_to do |format|
       if @activity.save
-        format.html { redirect_to @activity, notice: "Activity was successfully created." }
+        format.html do
+          redirect_to investigation_activities_path(@activity.investigation),
```
broder: Do we need the `show` route anymore if we're not redirecting to it here?

app/assets/stylesheets/shared.scss
```diff
@@ -0,0 +1,37 @@
+.aspect-ratio-3-2-container {
+  position: relative;
+  width: 100%;
+  padding-top: 66.66%; /* 3:2 Aspect Ratio */
+
+  .aspect-ratio-contents {
```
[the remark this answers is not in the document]
AnthonyPengelly: Alongisde the `.product-image` stuff below, I think this is working reasonbly well. It doesn't stretch, it just doesn't always go up to the edges of the container.

app/controllers/activities_controller.rb
AnthonyPengelly: I think you're right, we don't need the show route. We also don't really need the edit, update or destroy routes either? I wasn't sure whether nuking them all was a good idea at this stage, but shall I just go for it?

app/assets/stylesheets/shared.scss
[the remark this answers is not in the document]
broder: Cool!

app/controllers/activities_controller.rb
broder: Yeah, sounds good

PR #40, opened 2018-08-22
MSPSDS-261: GDSify

LukaszDygon:
I've managed to create migration that does not nuke the db, but it is missing the rollback implementation


db/migrate/20180822105208_change_uuids_to_ids.rb
```diff
@@ -0,0 +1,113 @@
...
+      execute "ALTER TABLE #{table} ADD PRIMARY KEY (id);"
+      execute "ALTER SEQUENCE #{table}_new_id_seq RENAME TO #{table}_id_seq;"
+    end
+
+    tables.each do |table|
+
+    end
+  end
+
+  def uuid_to_id(table_name, relation_name, relation_class)
```
broder: Are `relation_name` and `relation_class` always the same?

db/migrate/20180822105208_change_uuids_to_ids.rb
```diff
@@ -0,0 +1,113 @@
+class ChangeUuidsToIds < ActiveRecord::Migration[5.2]
+
+  def up
```
broder: This is crazy but good job!

db/migrate/20180822105208_change_uuids_to_ids.rb
```diff
@@ -0,0 +1,113 @@
...
+    uuid_to_id("investigation_products", "investigation", "investigation")
+    uuid_to_id("investigation_products", "product", "product")
+    uuid_to_id_for_sourceable("sources", "sourceable")
+    uuid_to_id_for_versions("versions", "item")
+
+    tables.each do |table|
+      remove_column table, :id
+      rename_column table, :new_id, :id
+      execute "ALTER TABLE #{table} ADD PRIMARY KEY (id);"
+      execute "ALTER SEQUENCE #{table}_new_id_seq RENAME TO #{table}_id_seq;"
```
broder: I think you can make the IDs start at a certain number using `RESTART 10000` - do you think that would make the UI look a bit cleaner? Presumably the IDs for the existing items have already been set at this stage though...

db/migrate/20180822105208_change_uuids_to_ids.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    tables.each do |table|
```
broder: Probably don't need this loop

db/schema.rb
```diff
@@ -37,17 +37,15 @@
...
 
-  create_table "activities", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
-    t.uuid "investigation_id"
+  create_table "activities", id: :serial, force: :cascade do |t|
     t.text "notes"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.integer "activity_type", null: false
-    t.index ["investigation_id"], name: "index_activities_on_investigation_id"
+    t.integer "investigation_id"
```
broder: Do we need to re-add the indices for all these relationships?

app/assets/stylesheets/govuk_elements.scss
```diff
@@ -35,6 +35,8 @@
 }
 
 .not-implemented {
+  // to comply with the GDS standards, components missing functionality are hidden
```
broder: I've added https://regulatorydelivery.atlassian.net/browse/MSPSDS-265 so we don't forget about this - could you add a `TODO` here?

db/migrate/20180822105208_change_uuids_to_ids.rb
LukaszDygon: They are in the migration. It's not always the case. For example "assignee" is a "user" class. I suppose I should remove it here?

PR #43, opened 2018-08-23
MSPSDS-264 product and business suggestions design

AnthonyPengelly:


app/views/application/_product_summary.html.erb
```diff
@@ -28,15 +28,6 @@
       </tbody>
     </table>
     <p><%= link_to "Full product details", product %></p>
-    <% if allow_removing %>
```
broder: I think Sam was setting this variable in `businesses/confirm_merge.html.erb` and `products/confirm_merge.html.erb` - do you want to update those files as well?

PR #45, opened 2018-08-24
MSPSDS-230: Document management functionality for investigations and products

tom-bell-softwire:


tom-bell-softwire:
Yes, there's a fair amount of common functionality between the two that could be extracted. I've raised MSPSDS-288 to look at refactoring it.

PR #46, opened 2018-08-24
MSPSDS-231: Remove RAPEX images from new/edit product form

tom-bell-softwire:
RAPEX images should only be added to products when importing data directly from RAPEX, not for manually created products.


app/views/products/_form.html.erb
```diff
@@ -38,13 +38,6 @@
     <%= form.text_field :associated_parts, :class => "form-control" %>
   </div>
 
-  <div class="form-group">
```
broder: Could we also remove the `accept_nested_attributes_for :rapex_images` in `models/product.rb`

tom-bell-softwire: Good point!

PR #49, opened 2018-08-24
MSPSDS-280: Fix investigations search & merge styling

broder:


app/views/businesses/index.html.erb
```diff
@@ -13,8 +13,9 @@
 <br/>
 
 <%= form_with(url: confirm_merge_businesses_path, method: :get) do |form| %>
-  <%= form.submit "Merge selected", name: nil %>
+  <%= form.submit "Merge selected", name: nil, class: "govuk-button" %>
```
AnthonyPengelly: I wonder if this should just be a link? However, styling a submit button as a link won't be fun :joy:

broder: Yeah... I feel they should be buttons so that people don't miss them - maybe I'll raise it with Tim this afternoon.

broder: And now we're hiding it...

AnthonyPengelly:
Fair enough!

PR #50, opened 2018-08-24
MSPSDS-285: Rename investigations to cases in the UI

broder:


app/views/activities/show.html.erb
```diff
@@ -1,53 +0,0 @@
-<% content_for :page_title, "View activity - Market Surveillance & Product Safety" %>
```
AnthonyPengelly: Did I not delete this already? Oops

app/views/investigations/edit.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :page_title, "Edit investigation - Market Surveillance & Product Safety" %>
```
AnthonyPengelly: I think you've missed a few of these page titles

broder: Good spot, thanks!

tom-bell-softwire:
Thanks for this! I think there's another reference buried in `images/index.html.erb`:

```ruby
<%= link_to "Back to #{@parent.class.name.downcase}" ...
```

Think the only way to handle it is with an `if-else` for Investigation, unfortunately...

PR #51, opened 2018-08-24
MSPSDS-279: Fix PDF creation for cases and products

tom-bell-softwire:


app/views/investigations/show.pdf.erb
```diff
@@ -2,7 +2,9 @@
   Investigation
 </h2>
 
-<%= image_tag(url_for(@investigation.image), alt: "Investigation") if @investigation.image.attached? %>
+<% if @investigation.images.attached? && @investigation.images.first.variable? %>
```
broder: Do we need to check the first image has `safe` in the metadata?

PR #53, opened 2018-08-24
MSPSDS-230: Remove broken image asset

tom-bell-softwire:
Something weird happened with the placeholder image used on the document management page, and it was showing as corrupted.

Git doesn't recognise the change if the file is just replaced, so I think I'll have to remove and re-add it as separate PRs :(


broder:
We'll just push this straight to master

PR #54, opened 2018-08-24
MSPSDS-271 update product styling

AnthonyPengelly:


app/helpers/products_helper.rb
```diff
@@ -2,8 +2,8 @@ module ProductsHelper
   # Never trust parameters from the scary internet, only allow the white list through.
   def product_params
     params.require(:product).permit(
-      :gtin, :name, :description, :model, :batch_number, :url_reference, :brand, :serial_number,
-      :manufacturer, :country_of_origin, :date_placed_on_market, :associated_parts,
+      :gtin, :name, :description, :model, :batch_number, :brand,
+      :country_of_origin, :date_placed_on_market,
       images_attributes: %i[id title url _destroy]
```
tom-bell-softwire: Can the image attributes also go, now that we aren't allowing RAPEX images to be added when adding/editing products?

AnthonyPengelly: Sounds reasonable to me!

app/views/application/_documents_list_summary.html.erb
```diff
@@ -0,0 +1,19 @@
+<% if parent.documents.attached? %>
```
tom-bell-softwire: Thanks for creating these! :)

PR #60, opened 2018-08-24
MSPSDS-294: Fix links on PDFs created from cases and products

tom-bell-softwire:


app/views/investigations/show.pdf.erb
```diff
@@ -49,9 +46,9 @@
     </tr>
     <% @investigation.activities.each do |activity| %>
       <tr>
-        <td><%= activity.created_at %></td>
+        <td><%= activity.created_at.to_s :db %></td>
```
broder: In other places, we're using `.to_s :long_ordinal` - not sure which is preferable!

tom-bell-softwire: I opted for the shorter format here so that each entry would fit on one line, but perhaps it's better to be consistent with the longer format used on the HTML pages. What do you think?

PR #61, opened 2018-08-28
MSPSDS-210: WIP: Sorting index

LukaszDygon:
There are still few things missing or working improperly:
- sorting is case sensitive
- sorting considers non-alphanumeric characters
- not all the columns in the index pages are index by elastic search
- filtering still has to be added

Additionally, perhaps it would be beneficial to refactor the code, as this is still work in progress


app/assets/stylesheets/application.css
```diff
@@ -18,4 +18,21 @@
 
 * {
     box-sizing: border-box;
-}
\ No newline at end of file
+}
+
+table th a {
+    color: black!important;
```
tom-bell-softwire: This will make all table header links black -- is that what we want?

If it's just needed for the specific case of sortable columns, could you add a dedicated class to those cells? This would probably allow you to get rid of the `!important` declaration as well, which is generally best to avoid, since it makes it hard to re-style related elements in the future.

docker-compose.yml
```diff
@@ -9,6 +9,8 @@ services:
     image: elasticsearch
     environment:
       - "ES_JAVA_OPTS=-Xms750m -Xmx750m"
+    ports:
+      - "9200:9200"
```
tom-bell-softwire: Should this change be committed?

app/controllers/investigations_controller.rb
```diff
@@ -99,6 +95,7 @@ def destroy
   def save_and_respond(notice)
     respond_to do |format|
       if @investigation.save
+        @investigation.__elasticsearch__.index_document
```
broder: Do we need to do this explicitly for all the models?

app/models/investigation.rb
```diff
@@ -33,6 +33,10 @@ class Investigation < ApplicationRecord
 
   enum sensitivity: %i[low medium high], _suffix: true
 
+  def as_indexed_json(*)
+    as_json.merge(assignee: assignee&.email)
```
broder: If we were to update the assigned `User` model, would it trigger an update to this index?

app/controllers/investigations_controller.rb
tom-bell-softwire: Ah, I hadn't spotted this was here! I might check with Lukasz...

I'd be surprised if we need to explicitly re-index when saving a record, but there could be something special about this model.

app/models/investigation.rb
tom-bell-softwire: Good point. I'll check what happens if we modify the user's email address.

tom-bell-softwire: Changes to `belongs_to` associations don't trigger a re-index, so I've removed the ability to sort by assignee for now (and therefore removed this index).

I think it _*is*_ possible to have changes to associations trigger the appropriate index updates, but it needs more time to investigate.

app/controllers/investigations_controller.rb
tom-bell-softwire: This was here to handle changing the assignee, but I've removed it for now until we can implement a better way to manage re-indexing when associations change.

PR #65, opened 2018-08-28
MSPSDS-286 accessible autocomplete

AnthonyPengelly:


app/views/investigations/assign.html.erb
```diff
@@ -1,33 +1,24 @@
...
 
-<h1 class="heading-xlarge">Update assignee</h1>
 <h4 class="heading-medium">Current assignee</h4>
 <p><%= @investigation.assignee.nil? ? "Unassigned" : @investigation.assignee.email %></p>
 
 <%= form_with(url: update_assignee_investigation_path(@investigation), local: true) do |form| %>
   <div class="form-group">
     <%= form.label :email, :class => "form-label" %>
     <%= form.select :email,
+                    # When there are many users, we should consider loading them with ajax using the autocomplete
```
broder: Should we do this now?

package.json
```diff
@@ -13,6 +13,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "accessible-autocomplete": "^1.6.1",
```
broder: Should we put this on "Accessibility Discovery of the Day"?

app/views/investigations/assign.html.erb
AnthonyPengelly: I'm happy to pick this up as my next task. I thought due to time restrictions it's better to get this through the door, and then we can make it ajax-y as another task. (I had a quick go, but it didn't Just Work, so I moved on)

AnthonyPengelly: I'm happy to pick this up as my next task. I thought due to time restrictions it's better to get this through the door, and then we can make it ajax-y as another task. (I had a quick go, but it didn't Just Work, so I moved on)

AnthonyPengelly: Also, due to disabling prefix search, your autocomplete experience will be strictly worse once we move it to ajax 😛 though I think it's worth doing anyway

broder: Don't worry about doing it now - I was hoping that it would Just Work 😐

PR #70, opened 2018-08-31
MSPSDS-269: Remove govuk_template dependency

broder:
There are a few outstanding styling issues but I'll pick them up as part of another PR - I don't want this to get any bigger!


app/views/layouts/application.html.erb
```diff
@@ -1,25 +1,136 @@
```
tom-bell-softwire: Is most of this file copied across from `govuk_template`?

broder: Mostly from https://design-system.service.gov.uk/styles/page-template/ with sprinklings of `govuk_template` (which has been deprecated in favour of the design system)

PR #71, opened 2018-09-03
MSPSDS-212: Add case filtering by status

tom-bell-softwire:
This just adds open/closed status filtering for cases, but the basic functionality should be fairly easy to extend to allow more filter options in the future.

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -78,25 +78,48 @@ module Searchable
       __elasticsearch__.search(query)
     end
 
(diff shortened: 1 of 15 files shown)


app/models/concerns/searchable.rb
```diff
@@ -78,22 +82,6 @@ def self.prefix_search(params, field)
       __elasticsearch__.search(query)
     end
 
-    # "multi_match" searches across all fields, applying fuzzy matching to any text and keyword fields
-    def self.fuzzy_search(params)
```
tom-bell-softwire: These methods used to construct the Elasticsearch query have been moved to the `ElasticsearchQuery` class.

PR #73, opened 2018-09-04
MSPSDS-109: Add Slim, JS & Sass linting

broder:


app/views/addresses/show.html.slim
```diff
@@ -0,0 +1,49 @@
...
+            | Line 1
+          td.govuk-table__cell
+            = @address.line_1
+        tr.govuk-table__row
+          th.govuk-table__header[scope="row"]
+            | Line 2
+          td.govuk-table__cell
+            = @address.line_2
+        tr.govuk-table__row
+          th.govuk-table__header[scope="row"]
```
tom-bell-softwire: If we end up using these elements a lot, might be good to define [Slim shortcuts](https://github.com/slim-template/slim#shortcuts) for them, so we could replace this with something like:

```
th-govuk
  | Locality
```

PR #77, opened 2018-09-06
MSPSDS-109: Further frontend fixes

broder:


.rubocop.yml
```diff
@@ -5,7 +5,8 @@ AllCops:
   Include:
     - 'app/**/*.rb'
     - 'lib/**/*.rb'
-    - 'spec/**/*.rb'
+    - 'lib/**/*.rake'
+    - 'test/**/*.rb'
```
tom-bell-softwire: Is it worth linting DB migrations as well?
What about YAML files (though perhaps Rubocop isn't best for those)?

broder: I think linting the migrations is a good idea - I've added it to MSPSDS-110 which involves more migration tweaks.

Can rubocop lint YAML?

tom-bell-softwire: Ah, looks like no. There's [yaml-lint](https://github.com/Pryz/yaml-lint), which only seems to check for correctness rather than style, or [YamlLint](https://github.com/shortdudey123/yamllint), which seems to also check for style but is perhaps not actively maintained...

broder: Interesting! I think we can leave it for now and then add yaml linting later if we need it

tom-bell-softwire: Yep, sounds good!

PR #79, opened 2018-09-07
MSPSDS-304: Create customised Keycloak docker image for local development

tom-bell-softwire:
This uses the same package script as for deploying to Cloud Foundry and installs onto a standard JBoss docker image.

Default `master` and `mspsds` realms are imported on first launch, which sets up the relevant client credentials and test users.

Unfortunately, devs will need to add an entry to their `/etc/hosts` file to allow the docker containers and host to connect to Keycloak on the same hostname (I'll add that requirement to the README as part of a separate PR).


.env-template
```diff
@@ -1,9 +1,9 @@
 # Copy this file to .env and fill in the values with your own env variables
 
 # REQUIRED: Keycloak client credentials
-KEYCLOAK_AUTH_URL=
-KEYCLOAK_CLIENT_ID=
-KEYCLOAK_CLIENT_SECRET=
+KEYCLOAK_AUTH_URL=http://keycloak:8080/auth
+KEYCLOAK_CLIENT_ID=mspsds-app
+KEYCLOAK_CLIENT_SECRET=932677f2-55c0-45c5-901a-f4beddb85e17
```
tom-bell-softwire: Do you think it's ok to include this here, given that it's a dev instance of Keycloak? Alternatively, I can add instructions to the README to generate a new secret and update the `.env` file as part of dev setup.

keycloak/deploy.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
tom-bell-softwire: This has just been moved across from the old `/docker/keycloak` directory and updated to match the style of the other deploy scripts.

.env-template
broder: I think it's fine to leave it here but we should probably put a comment to make it obvious that it's only used for development environments.

keycloak/deploy.sh
broder: :+1:

keycloak/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+###################
+
+COPY ./docker-entrypoint.sh /
+RUN chmod +x /docker-entrypoint.sh
+
+USER 1000
+
+EXPOSE 8080
+
+ENTRYPOINT ["/docker-entrypoint.sh"]
```
broder: Are we likely to want to change the arguments that we pass to keycloak? Maybe we should just put the stuff from this file into `CMD` below.

postgres/setup-keycloak.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+# Create the database and user required by Keycloak
+
+psql -v ON_ERROR_STOP=1 --username postgres <<-EOSQL
+  CREATE DATABASE keycloak;
```
broder: Could we parameterise this to use environment variables (like the keycloak container does)? Then we could do something like https://github.com/docker/compose/issues/210 to share it between both containers in the `docker-compose.yml`

postgres/setup-keycloak.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -ex
+
+# Create the database and user required by Keycloak
```
broder: Do we need to do something like this on GOV.UK PaaS as well?

.env-template
tom-bell-softwire: Sounds good.

keycloak/Dockerfile
tom-bell-softwire: Good point -- I think it'll always be run with the same arguments.

postgres/setup-keycloak.sh
tom-bell-softwire: Currently, an empty Keycloak database is created on first deploy, with setup then done within the Keycloak admin console. However, it would be neater to start with a common "default" setup that includes an admin user and realm configuration.

I'll look to do that as a separate PR.

docker-compose.yml
```diff
@@ -2,9 +2,16 @@ version: '3'
...
+    build:
+      context: ./postgres
+      dockerfile: Dockerfile
     volumes:
       - postgres-volume:/var/lib/postgresql/data
+    environment: &keycloak
+      - KEYCLOAK_HOSTNAME=db
+      - KEYCLOAK_DATABASE=keycloak
+      - KEYCLOAK_USERNAME=keycloak
+      - KEYCLOAK_PASSWORD=password
```
tom-bell-softwire: I'm not entirely happy with these variable names. Do you think it's worth being clearer and calling them `KEYCLOAK_DB_...` instead?

broder: I think `KEYCLOAK_DATABASE_USERNAME` `KEYCLOAK_DATABASE_PASSWORD` would be fine.

PR #94, opened 2018-09-11
MSPSDS-289: Use open source version of Elasticsearch Docker image

tom-bell-softwire:
This version does not come bundled with the commercial X-Pack plugin or trial licence.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,6 @@ services:
       context: .
       dockerfile: ./elasticsearch/Dockerfile
     environment:
-      - xpack.security.enabled=false
       - "ES_JAVA_OPTS=-Xms750m -Xmx750m"
     volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -10,7 +10,6 @@ services:
       context: .
       dockerfile: ./elasticsearch/Dockerfile
     environment:
-      - xpack.security.enabled=false
       - "ES_JAVA_OPTS=-Xms750m -Xmx750m"
```
broder: https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-cli-run-dev-mode suggests that `discovery.type=single-node` would be good for development as well.

tom-bell-softwire: Yes, that looks like it would be worth adding.

PR #96, opened 2018-09-12
MSPSDS-266: Remove RAPEX images

broder:


db/migrate/20180828082107_init_schema.rb
```diff
@@ -2,7 +2,7 @@ class InitSchema < ActiveRecord::Migration[5.0]
   def up
     # These are extensions that must be enabled in order to support this database
     enable_extension "plpgsql"
-    create_table "active_storage_attachments", force: :cascade do |t|
+    create_table "active_storage_attachments" do |t|
```
broder: Strong migrations was complaining about the `force: :cascade` stuff :/

lib/tasks/rapex_import.rake
```diff
@@ -67,11 +67,22 @@ def create_investigation(notification, date, name)
...
-    updated_at: date,  # TODO MSPSDS-131: confirm this is what we want instead of the current Date
+    updated_at: date,
     source: ReportSource.new(name: "RAPEX")
   )
 end
 
+def create_product_images(notification, product)
+  urls = notification.xpath("pictures/picture")
+  urls.each do |url|
+    clean_url = url.text.delete("\n") unless url.nil?
```
tom-bell-softwire: Ed mentioned on [MSPSDS-287](https://regulatorydelivery.atlassian.net/browse/MSPSDS-287) that larger images appear to be available on RAPEX:

> It looks like larger images are available (e.g. compare https://mspsds-int.cloudapps.digital/investigations/6890 and https://ec.europa.eu/consumers/consumers_safety/safety_products/rapex/alerts/?event=viewProduct&reference=0972/10).
> Looks like maybe we can use 'f' at the end of the image URL rather than 's'?

Is it worth downloading the full size images here?

db/migrate/20180828082107_init_schema.rb
tom-bell-softwire: Ah, ok - thanks!

lib/tasks/rapex_import.rake
broder: Good catch!

tom-bell-softwire: Great - thanks!

PR #100, opened 2018-09-14
MSPSDS-352: Add first system tests

broder:


README.md
```diff
@@ -1,6 +1,7 @@
 # Market Surveillance & Product Safety Digital Service
 
 [![Build Status](https://travis-ci.org/UKGovernmentBEIS/beis-mspsds.svg?branch=master)](https://travis-ci.org/UKGovernmentBEIS/beis-mspsds)
+[![Coverage Status](https://coveralls.io/repos/github/UKGovernmentBEIS/beis-mspsds/badge.svg?branch=master)](https://coveralls.io/github/UKGovernmentBEIS/beis-mspsds?branch=master)
```
tom-bell-softwire: Nice!

app/controllers/activities_controller.rb
```diff
@@ -19,8 +19,8 @@ def create
     respond_to do |format|
       if @activity.save
         format.html do
-          redirect_to investigation_activities_path(@investigation),
-                      notice: "Activity was successfully created."
+          redirect_to investigation_activities_url(@investigation),
```
tom-bell-softwire: Just out of interest, why this change?

broder: It's something to do with changing the `config.action_controller.default_url_options` to `localhost` - by default it seems to be `example.com` so we when we just had the path, it was saying "expected `localhost/whatever` to equal `example.com/whatever`". I couldn't figure out how to override it... :/

tom-bell-softwire: Huh, weird!

PR #101, opened 2018-09-14
MSPSDS-470: Add custom Keycloak email provider to use GOV.UK Notify service

tom-bell-softwire:
Use the [GOV.UK Notify](https://www.notifications.service.gov.uk) service to send account verification and password reset emails from Keycloak.

For now, system event emails (e.g. for notifying admins about failed logins) are not supported, but these would be simple to add if needed in the future, by creating a new email template and implementing the relevant method in `NotifyEmailTemplateProvider`.


tom-bell-softwire:
Looks like the Notify service supports integration tests by using ["test" API keys](https://docs.notifications.service.gov.uk/java.html#test), so we should be able to use one of these.

keycloak/providers/.gitignore
```diff
@@ -0,0 +1,38 @@
+# Taken from: https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore
```
broder: We could probably just update the root `.gitignore` to include the Java stuff. We created that using gitignore.io

keycloak/providers/README.md
```diff
@@ -0,0 +1,54 @@
+# GOV.UK Notify Email Provider for Keycloak
+
+
+## Getting Setup
```
broder: It might be worth adding something brief here.

tom-bell-softwire: Good call.

PR #106, opened 2018-09-17
Mspsds 311 images in new tab

samlionheart:


app/views/images/_image.html.slim
```diff
@@ -1,4 +1,4 @@
 - if image.metadata["safe"]
-  = link_to render("images/image_tag", local_assigns), image, class: "image-with-preview"
+  = link_to render("images/image_tag", local_assigns), image, class: "image-with-preview", target: "_blank"
```
broder: Do we need to do something similar for documents as well?

PR #110, opened 2018-09-18
MSPSDS-378: Fix issue with ClamAV running in Docker for Windows

broder:
@raamSoftwire and @jasiekmiko FYI.

I think it's something to do with `chmod` not working on Docker for Windows? https://docs.docker.com/docker-for-windows/troubleshoot/#volumes


.dockerignore
```diff
@@ -0,0 +1,330 @@
+# This file should be mostly kept up to date with the .gitignore (with exceptions at the bottom after "# End of ...")
+# These are the files which will be ignored when creating a Docker container
```
tom-bell-softwire: I didn't know this was a thing. Cool!

PR #113, opened 2018-09-20
MSPSDS-309 Add company status to businesses

jasiekmiko:


app/models/business.rb
```diff
@@ -23,20 +24,56 @@ class Business < ApplicationRecord
   has_paper_trail
 
   def nature_of_business
+    # noinspection RubyResolve
```
tom-bell-softwire: It feels like we might end up with a lot of these entries throughout the codebase (RubyMine doesn't seem to recognise methods from quite a few Gems, either).

Is it worth just disabling this inspection? Or leaving it on and living with the false-positive warnings?

I think maybe we should start with the latter option, because the warnings are quite useful for catching mis-typed methods. What do you think?

jasiekmiko: Very much agree with not turning it off, so if anything I'd prefer false positives - but the more you have, the less useful warnings become.

All the ones in this file are actually the same thing, not recognising `companies_house_constants` - as I understand, magically appearing from a `yml` file, so i reckon little chance of making ruby understand where its coming from. 

Perhaps pulling it out into a variable and ignoring the warning just once?

tom-bell-softwire: Yep, sounds good!

tom-bell-softwire:
Looks good – thanks!

PR #117, opened 2018-09-21
MSPSDS-485 Surface the company status in search results

jasiekmiko:
This is what the search results look like:
image


app/views/businesses/_search_results.slim
```diff
@@ -19,12 +19,15 @@ table.govuk-table
...
       th.govuk-table__header[scope="col"]
+      th.govuk-table__header[scope="col"]
   tbody.govuk-table__body
     - if @companies_house_businesses.any?
       - @companies_house_businesses.each do |business|
         tr.govuk-table__row
           td.govuk-table__cell
             = link_to business[:company_name], business[:url], target: "_blank"
+          td.govuk-table__cell
+            = Rails.application.config.companies_house_constants["company_status"][business[:company_status_code]]
```
broder: If you put the ```def companies_house_constants
    Rails.application.config.companies_house_constants
  end``` bit you added into `business_helper.rb`, I think you could use that method in this view as well as in the model?

jasiekmiko: Ah yeah, been wondering with Tom where would make most sense. I'll go for that :)

jasiekmiko:
ta David :)

PR #121, opened 2018-09-24
Mspsds 360 change user details

raamSoftwire:
It might be worth having a look at the keycloak repo that I forked.

https://github.com/UKGovernmentBEIS/keycloak-theme-govuk


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule ".\\keycloak\\govuk-theme"]
```
broder: I'd guess that the Travis errors are due to this Windows-style path :(

keycloak/build-theme.sh
```diff
@@ -0,0 +1,23 @@
...
+THEME_PATH=./keycloak/govuk-theme
+PACKAGE_PATH=./keycloak/package
+
+# Download and add the GOV.UK theme
+cd $THEME_PATH
+npm install
+npm run build
+cd ../../
+
+mkdir -p $PACKAGE_PATH/themes/govuk
```
broder: It might work without these `mkdir` commands with the way that you've done `cp` below

PR #125, opened 2018-09-26
Rename submodule to not start with a period

greysteil:
GitHub doesn't deal well with submodule names that start with a period. In particular, it causes all calls to its [create tree API](https://developer.github.com/v3/git/trees/#create-a-tree) endpoint to 422 with a `GitRPC::BadGitmodules` error.

The name of a submodule is simply for convenience, so there shouldn't be any reason to prefer starting it with a period to not.


broder:
That's weird - thanks for your help in debugging!

PR #136, opened 2018-10-04
Mspsds 486 protect search results from companies house outages

jasiekmiko:


web/app/helpers/businesses_helper.rb
```diff
@@ -53,7 +53,12 @@ def create_business
 
   def advanced_search(excluded_ids = [])
     @existing_businesses = search_for_similar_businesses(@business, excluded_ids)
-    @companies_house_businesses = search_companies_house_for_similar_businesses(@business)
+    begin
+      @companies_house_businesses = search_companies_house_for_similar_businesses(@business)
+    rescue CompaniesHouseClient::ClientException => e
+      Rails.logger.error e
```
jasiekmiko: Not sure if this is the right way to log - not much in the way of examples in the code so far ;P
`logger.error e` seemed to work, until that code run through a test at which point it complained, hence the `Rails` bit

tom-bell-softwire: What error are you seeing in the tests when you don't include `Rails` ? I wonder if we're missing some logging config in our test setup.

.env-template
```diff
@@ -3,5 +3,5 @@
 # OPTIONAL: To test email sending
 NOTIFY_API_KEY=
 
-# OPTIONAL: To test companies house business integration
+# OPTIONAL: To test companies house business integration (register at https://developer.companieshouse.gov.uk/api/docs/)
```
tom-bell-softwire: Do we need to include this here, since we already describe how to obtain accounts in the readme?

Would it be worth adding a reference to the accounts section of the readme at the top of this file instead?

tom-bell-softwire: Also, tiny thing, but is it worth changing to "Companies House" since it's a proper name?

web/app/views/businesses/_search_results.slim
```diff
@@ -13,26 +13,30 @@ h3.govuk-heading-s
...
-            = link_to business[:company_name], business[:url], target: "_blank"
-          td.govuk-table__cell
-            = companies_house_constants["company_status"][business[:company_status_code]]
-          td.govuk-table__cell
-            = form_with(url: companies_house_new_business_form_url, local: true) do |form|
-              = form.hidden_field :company_number, value: business[:company_number]
-              = form.submit "Add", class: "govuk-button govuk-!-margin-0"
-    - else
+- if @companies_house_error
+  p
```
tom-bell-softwire: Should this be styled as an error message?

jasiekmiko: I could see it both ways, but opted for not as it's nothing that the user's done wrong - it's just that some of the functionality isn't available.
Do you reckon it's worth checking with the designers?

.env-template
jasiekmiko: yeah, good point - I'll point them to the readme! :)

web/app/helpers/businesses_helper.rb
jasiekmiko:
```
NameError: undefined local variable or method `logger' for #<BusinessHelperTest:0x000055c2aaa30da0>
    app/helpers/businesses_helper.rb:59:in `rescue in advanced_search'
    app/helpers/businesses_helper.rb:56:in `advanced_search'
    test/helpers/business_helper_test.rb:107:in `block (2 levels) in <class:BusinessHelperTest>'
    test/helpers/business_helper_test.rb:105:in `block in <class:BusinessHelperTest>'
```

web/app/views/businesses/_search_results.slim
tom-bell-softwire: I think it's fine to leave it as is.

web/app/helpers/businesses_helper.rb
tom-bell-softwire: Looks like you need to explicitly call `Rails.logger` outside of controllers and models:
https://stackoverflow.com/questions/6779250/nameerror-undefined-local-variable-or-method-logger

jasiekmiko: Aaah fair, I did see that but didn't think of the fact that we don't have a controller in the test. Cheers :)

README.md
```diff
@@ -59,17 +59,23 @@ Ask someone on the team to create an account for you on the Int and Staging envi
 
 #### GOV.UK Notify
 
-If you want to send emails from your development instance, or update any API keys for the deployed instances, you'll need an account for [GOV.UK Notify](https://www.notifications.service.gov.uk) - ask someone on the team to invite you.
+If you want to send emails from your development instance, or update any API keys for the deployed instances,
```
broder: I wonder if we should install some kind of markdown linting to enforce line length... Maybe there isn't enough churn on these files.

web/app/clients/companies_house_client.rb
```diff
@@ -24,12 +30,32 @@ def companies_house_businesses(search_term)
...
+    profile = get_company(business.company_number)
     business.with_company_house_info(profile)
   end
+
+private
+
+  def company_search(search_term)
+    begin
+      response = @client.company_search(search_term)
+    rescue StandardError => e
```
broder: Should we be catching `CompaniesHouse::APIError` (https://github.com/gocardless/companies-house-rest#error-handling) instead?

PR #141, opened 2018-10-08
Update readme to include git submodule information

raamSoftwire:


README.md
```diff
@@ -6,6 +6,8 @@
 
 ## Getting Setup
 
+This project includes a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules), so when running `git clone` you need to add the `--recurse-submodules` option.
```
broder: I guess you could also add the `git submodule init` and `git submodule update --recursive` commands as well - I think people might need to run the second command if the submodule is updated because of a `pull`.

raamSoftwire: Done.

PR #144, opened 2018-10-09
MSPSDS-609: Add links to case page

kicferk1:


web/app/views/investigations/index.html.slim
```diff
@@ -27,8 +27,12 @@
     .govuk-form-group.govuk-grid-column-one-third[class="govuk-!-margin-bottom-0 govuk-!-margin-top-6"]
       = form.submit "Search", name: nil, class: "govuk-button govuk-!-margin-bottom-0"
 br
-p
-  = link_to "Create new case", new_investigation_path, class: "govuk-!-font-size-19  govuk-!-font-weight-bold"
+.govuk-grid-row
+  .govuk-grid-column-full
+    p class="govuk-body"
```
broder: This could be done with `p.govuk-body` I think. (We only need to use `class=""` if it's conditionally assigned or if it has an `!` in the class name)

web/app/views/investigations/index.html.slim
```diff
@@ -27,8 +27,12 @@
...
+      = link_to "Report an unsafe product", new_investigation_path, class: "govuk-link mspsds-block-link"
+      = link_to "Begin an investigation", "/", class: "govuk-link mspsds-block-link"
```
broder: It might be worth putting a `TODO` comment near these links so we remember to update them.

web/app/views/investigations/index.html.slim
kicferk1: It can. 
Also .govuk-grid-row and .govuk-grid-column-full don't really affect the end result, do we need them? Do they have a purpose that's not immediately obvious from display?

web/app/views/investigations/index.html.slim
```diff
@@ -29,8 +29,9 @@
 br
 .govuk-grid-row
   .govuk-grid-column-full
-    p class="govuk-body"
+    .p-govuk-body
```
broder: Should this be `p.gov-uk-body`?

Also, `.govuk-grid-row` and `.govuk-grid-column-full` look like they're ensuring that the page width is correct at various screen sizes.

kicferk1: p.govuk-body, p.gov-uk-body and just p produce the same visual result on my machine, heroku prototype uses '<p class="govuk-body">'. There seems to be no styling difference caused by swapping between gov-uk and govuk.

PR #153, opened 2018-10-11
Mspsds 526 manualy list tabs

jasiekmiko:
Move away from loops?


web/app/views/investigations/tabs/_activity_control.slim
```diff
@@ -0,0 +1 @@
+= render "tab_control", id: "activity", name: "Activity", extra_class: "govuk-tabs__tab--selected"
```
MatthewBurstein: Should this file be .html.slim rather tha .slim?

jasiekmiko:
Decided to go with original approach

PR #155, opened 2018-10-12
MSPSDS-526: update case page to tabs layout

jasiekmiko:
An extension to the MSPSDS-526 work, klling the investigation_presenter.
This allows all of our code to view logic to remain within the views


web/test/controllers/investigations_controller_test.rb
```diff
@@ -42,6 +42,11 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
+  test "should generate investigation pdf" do
+    get investigation_url(@investigation) + ".pdf"
```
broder: Does `investigation_url(@investigation, format: :pdf)` work?

web/app/presenters/tab.rb
```diff
@@ -1,9 +1,9 @@
 class Tab
-  attr_reader :id, :title, :partial
+  attr_reader :id, :title, :render
 
   def initialize(id, title, partial)
     @id = id
     @title = title
-    @partial = partial
+    @render = partial
```
broder: Maybe if you're changing the below, you could change the argument to be `render` rather than `partial`.

jasiekmiko: ah good catch thanks!

web/test/controllers/investigations_controller_test.rb
jasiekmiko: yeah! thanks! :)

PR #156, opened 2018-10-12
Mspsds 514 record report originator

kicferk1:


web/app/controllers/investigations_controller.rb
```diff
@@ -54,6 +73,8 @@ def update_assignee
   # POST /investigations
   # POST /investigations.json
   def create
+    #'Record report originator' flow doesn't require the user to specify title, so we set it to 'Untitled case'
+    @investigation.title = 'Untitled case' if !@investigation.title
```
jasiekmiko: Hmmm... perhaps the solution would be to leave it empty, stop `title` from being required and display `Untitled case` when title is missing? Feels like that should be more resilient to changes in this area?

web/app/models/investigation.rb
```diff
@@ -11,6 +11,7 @@ class Investigation < ApplicationRecord
   end
 
   validates :title, presence: true
+  validates :reporter_type, presence: true
```
jasiekmiko: Oh-oh! I don't think we want this here, as this will make creating investigations via other flows fail!
It seems like our case validation wants to be a bit more independent from the model and based on the flows. @broder do you have any tips for doing this in a railsy-enough way? :P

web/app/views/investigations/tabs/_details.html.slim
```diff
@@ -1,2 +1,9 @@
-| Such detail
+p
+  = 'Reporter type: ' + (@investigation.reporter_type || '')
```
jasiekmiko: Do we want to be displaying them when there's no reporter?
Or are we explicitly leaving the polish of this section to another ticket? If so I'd recommend tagging a TODO with it :)

web/config/application.rb
```diff
@@ -14,7 +14,9 @@ class Application < Rails::Application
     # Application configuration can go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
-
     config.eager_load_paths << Rails.root.join("presenters")
+    config.action_view.field_error_proc = Proc.new { |html_tag, instance|
```
jasiekmiko: Is this the bit that turns off rails' "smart" errors bit?
If so a comment to that effect would be helpful :)

web/app/controllers/investigations_controller.rb
```diff
@@ -30,6 +30,25 @@ def new
     @investigation = Investigation.new
   end
 
+  # GET /investigations/new_report
+  def new_report
+    @investigation = Investigation.new
+    @investigation.errors.add(:reporter_type, :invalid, message: 'Please select reporter type') if params[:error]
```
jasiekmiko: A bit worried that the error seems very generic and we're translating it into a very concrete message.
Would using a more explicit error parameter help here?

web/app/controllers/investigations_controller.rb
```diff
@@ -30,6 +30,25 @@ def new
...
+  end
+
+  # POST and GET /investigations/new_report_details
+  def new_report_details
+    return redirect_to investigations_path if request.get?
+
+    @investigation = Investigation.new(investigation_params)
+    @investigation.validate
+    if @investigation.errors[:reporter_type].any?
+      redirect_to(new_report_investigations_url(error: true))
```
broder: We could store the error on the `session` object either.

web/app/models/investigation.rb
broder: I wonder if the easiest way would be to move the reporter details (and this validation) into a separate model and then make an investigation optionally have a reference to a reporter?

(@jasiekmiko to answer your question about validation based on a certain flow, you can do something like https://guides.rubyonrails.org/active_record_validations.html#conditional-validation but I don't think we need to do that here)

web/app/views/investigations/index.html.slim
```diff
@@ -30,7 +30,7 @@ br
 .govuk-grid-row
   .govuk-grid-column-full
     .p-govuk-body
-      = link_to "Report an unsafe product", new_investigation_path, class: "govuk-link mspsds-block-link"
+      = link_to "Report an unsafe product", new_report_investigations_url, class: "govuk-link mspsds-block-link"
```
broder: I think `new_report_investigations_path` would be fine here.

web/app/views/investigations/new_report.html.slim
```diff
@@ -0,0 +1,56 @@
...
+                        "data-aria-controls": "conditional-reporter-type-other"
+                = form.label :reporter_type, "Other", class: "govuk-label govuk-radios__label"
+                  | Other
+              #conditional-reporter-type-other.govuk-radios__conditional.govuk-radios__conditional--hidden
+                .govuk-form-group
+                  span#other-reporter-hint.govuk-hint [class="govuk-!-width-two-thirds"]
+                    | Please give details
+                  = form.text_field :other_reporter, class: "govuk-input govuk-!-width-two-thirds",\
+                    "aria-describedby": "other-reporter-hint"
+        button.govuk-button type="submit"
```
broder: You could do this with `form.submit "Submit", class: "govuk-button"` instead.

web/yarn.lock
```diff
@@ -773,9 +773,9 @@ govuk-country-and-territory-autocomplete@^0.5.0:
     openregister-picker-engine "^1.2.1"
     webpack-sources "^1.3.0"
 
-govuk-frontend@^2.2.0:
-  version "2.2.0"
-  resolved "https://registry.yarnpkg.com/govuk-frontend/-/govuk-frontend-2.2.0.tgz#ef14463f8c0b4b628d5a498f8aa4f3bf6833315e"
+govuk-frontend@^2.1.0:
```
broder: I'm not sure that we want to commit this - I think if you run `yarn install` again after you've merged master, it should go away.

web/app/controllers/investigations/report_controller.rb
```diff
@@ -2,32 +2,40 @@ class Investigations::ReportController < ApplicationController
   include Wicked::Wizard
   steps :type, :details
 
+  # GET /investigations/report/new
+  def new
+    redirect_to report_index_path if request.get?
```
broder: Would the request ever not be a get?

web/app/controllers/investigations/report_controller.rb
```diff
@@ -2,32 +2,40 @@ class Investigations::ReportController < ApplicationController
...
+    redirect_to investigation_url(@investigation)
+  end
+
+  # GET /investigations/report
+  # GET /investigations/report/step
   def show
-    if step == steps.first
-      session[:reporter] = {}
-    end
+    session[:reporter] = {} if step == steps.first
```
broder: I wonder if we should remove the `session[:reporter]` at the end of the journey as well? Might help keep cookies smaller?

web/config/routes.rb
```diff
@@ -21,6 +21,9 @@
       get :assign
       post :update_assignee
     end
+    collection do
+      resources :report, controller: "investigations/report"
```
broder: Should we restrict the routes here using `only`? Maybe to `new create show update`? (We'd then maybe need to update the entrypoint to be `new_report_path` or similar)

web/app/controllers/investigations/report_controller.rb
kicferk1: Probably not, and it's not handled in here anyway.

PR #159, opened 2018-10-12
MSPSDS-604: case page header

raamSoftwire:


web/app/views/application/_investigation_heading.html.slim
```diff
@@ -1,10 +1,39 @@
...
+    p.govuk-caption-xl
+      | Case:
+      '
+      = investigation.id
+
+    = investigation.title
+    p.govuk-caption-m
+      | Created by
+      '
+      = @investigation.source.user.first_name + " " + @investigation.source.user.last_name + ", " + investigation.created_at.strftime("%d/%m/%Y")
```
raamSoftwire: Is there a better way of doing this?

web/app/views/application/_investigation_heading.html.slim
```diff
@@ -1,10 +1,39 @@
...
+            | Assigned to
+          = @investigation.assignee ? "Assigned to #{@investigation.assignee.email}" : "Unassigned"
+          td.govuk-table__cell [class="govuk-!-font-size-16"]
+            = link_to "Change"
+
+      tr.govuk-table__row
+        th.govuk-table__header[scope="row"]
+          span.govuk-caption-m[class="govuk-!-font-size-16"]
+            | Priority
+          | Not set
```
raamSoftwire: Hard coded for now. From what I could see priority doesn't exist yet and should be created as part of another ticket.

web/app/views/application/_investigation_heading.html.slim
```diff
@@ -1,10 +1,39 @@
...
-  '
-  = investigation.id
-h1.govuk-heading-xl
-  = investigation.title
-p.govuk-body
-  | Created:
-  '
-  = investigation.created_at.strftime("%d %B %Y")
+.govuk-grid-column-two-thirds
+  h1.govuk-heading-xl
```
broder: I think have other tags inside an `h1` is bad for accessibility (as people use headers to navigate). It might be worth asking Tim what should be called out when a user is navigating by header... Maybe we aren't thinking about this kind of stuff yet...

web/app/views/application/_investigation_heading.html.slim
broder: You should also use `source.show` as it's not always created by a user (e.g. if we import from RAPEX)

web/app/views/investigations/show.html.slim
```diff
@@ -3,45 +3,17 @@
...
-    p.govuk-body class="govuk-!-font-size-48 govuk-!-font-weight-bold"
-      = @investigation.sensitivity&.titleize
+
+/ TODO: MSPSDS-532
+/ These exist on the design but don't lead any where yet. Add with links, or delete if changed
+/ https://mspsds-design-prototype.herokuapp.com/case
+/ p
+/   a href="" Edit risk details
+/   a style="margin-left: 20px" href="" View assessment
+
```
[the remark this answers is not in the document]
raamSoftwire: No, I think that falls under ticket 528

web/app/views/application/_investigation_heading.html.slim
raamSoftwire: I did think it was weird, but I was following Nick's pattern from the prototype. Happy to chat about this with Nick on Monday.

web/app/views/investigations/show.html.slim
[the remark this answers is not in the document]
broder: Ah, fair enough.

web/app/views/application/_investigation_heading.html.slim
broder: Sounds good, thanks :+1:

raamSoftwire: It looks like this is the desired pattern from the [GOV guidelines](https://design-system.service.gov.uk/styles/typography/)

web/app/views/application/_investigation_heading.html.slim
raamSoftwire: Done.

web/app/views/application/_investigation_heading.html.slim
broder: It says that it can be in or out of the `h1` tag depending on whether you want it to be a part of the heading (and read out on screen readers) https://design-system.service.gov.uk/styles/typography/#headings-with-captions. So I guess it's fine for screen readers to read the whole thing?

PR #160, opened 2018-10-12
MSPSDS-623: prompt to associate product with case

raamSoftwire:


web/app/views/investigations/show.html.slim
```diff
@@ -42,6 +42,17 @@
       | Sensitivity
     p.govuk-body class="govuk-!-font-size-48 govuk-!-font-weight-bold"
       = @investigation.sensitivity&.titleize
+
+-if @investigation.investigation_products.length == 0
```
broder: Does `@investigation.products.any?` work?

PR #164, opened 2018-10-15
MSPSDS-629 and MSPSDS-630 business and product tabs

MatthewBurstein:
Content has been moved into appropriate tabs where it exists, tabs are empty where it does not.


web/app/views/businesses/tabs/_activity.html.slim
```diff
@@ -0,0 +1,4 @@
+#history.govuk-grid-row[class="govuk-!-padding-bottom-6"]
```
broder: Are these IDs required any more?

web/app/views/businesses/show.html.slim
```diff
@@ -12,21 +12,7 @@
...
-        li
-          a[class="govuk-!-font-size-19" href="#locations"]
-            | Locations
-        li
-          a[class="govuk-!-font-size-19" href="#cases"]
-            | Cases
-        li
-          a[class="govuk-!-font-size-19" href="#history"]
-            | History
+      // in the prototype there is a placeholder for the product image here.
```
broder: Should we add a TODO so we don't forget this? Is there another JIRA card to add this image?

web/app/views/products/show.html.slim
```diff
@@ -11,24 +11,7 @@
...
-      li
-        a[class="govuk-!-font-size-19" href="#documents"]
-          | Documents
-      li
-        a[class="govuk-!-font-size-19" href="#cases"]
-          | Cases
-      li
-        a[class="govuk-!-font-size-19" href="#history"]
-          | History
+  // in the prototype there is a placeholder for the product image here
```
broder: We do have images on products already - could we use one of those or is it part of a separate ticket?

web/app/views/businesses/tabs/_activity.html.slim
```diff
@@ -0,0 +1,4 @@
...
+  .govuk-grid-column-full
+    h2.govuk-heading-l[class="govuk-!-margin-bottom-1"]
```
broder: Is this tag adding anything? Should we have `| History` on the next line?

broder:
Looks like we'll need to add a test to cover the `get_first_image` method that you added - maybe you could add an image to investigation and check that `investigation#show` still renders correctly? (https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/test/controllers/investigations_controller_test.rb#L41 already tests that it renders successfully, it might just be a case of adding an image to `@investigation`)

web/app/views/businesses/show.html.slim
```diff
@@ -12,23 +12,9 @@
...
-        li
-          a[class="govuk-!-font-size-19" href="#locations"]
-            | Locations
-        li
-          a[class="govuk-!-font-size-19" href="#cases"]
-            | Cases
-        li
-          a[class="govuk-!-font-size-19" href="#history"]
-            | History
+      //  TODO in the prototype there is a placeholder for the product image here.
```
jasiekmiko: Does this still need resolving as part of this ticket? If not, then would be great to tag this TODO with a ticket number!

MatthewBurstein: David commented about this [here](https://github.com/UKGovernmentBEIS/beis-mspsds/pull/164#discussion_r225438723). Since Products already have images I added them to the page as part of this PR but Businesses do not so I haven't yet.

Having said that, I've had a look and I don't think I can see another ticket for it...

PR #171, opened 2018-10-17
Mspsds 522 add incident data to cases

jasiekmiko:
The ticket is not finished yet, but I'd like some early feedback/help with the structure of the date.
I really don't like the way the controller is dealing with it rn, it's messy and patchy - it's the end result of trying to use rails' magic date handling, followed by changes on top of it to make the govuk component cooperate with it well.
I guess I'm looking for a pointer on how to structure it so it's cleaner but still agreeing with rails conventions/style.


web/db/schema.rb
```diff
@@ -70,6 +70,18 @@
...
   end
 
+  create_table "incidents", id: :serial, force: :cascade do |t|
+    t.string "affected_party"
+    t.datetime "created_at", null: false
+    t.date "date"
+    t.text "description"
+    t.integer "investigation_id"
+    t.string "location"
+    t.string "type"
```
broder: I think this is causing the Travis errors. Running `docker-compose exec web bin/rake db:drop db:create db:migrate` should regenerate this file

web/config/routes.rb
```diff
@@ -35,6 +35,9 @@
         post :companies_house
       end
     end
+
+    resources :incidents, only: %i[new create edit] do
```
broder: Are we using `#edit`?

web/app/views/incidents/new.html.slim
```diff
@@ -0,0 +1,81 @@
...
+          legend.govuk-fieldset__legend.govuk-fieldset__legend--l
+            | When did it happen?
+          span#incident-date-hint.govuk-hint
+            | For example, 31 3 2018
+          - if form.object.errors[:date].present?
+            span#date-error.govuk-error-message
+              = form.object.errors[:date].first
+          #incident-date.govuk-date-input
+            .govuk-date-input__item
+              / The ids for these fields follow a special convention to work with Active Record dates
```
broder: Should we pull this (and the other date specific code in the controller / model) out into another partial / other files?

web/app/views/incidents/new.html.slim
```diff
@@ -0,0 +1,81 @@
+- content_for :page_title, "Add Incident - Market Surveillance & Product Safety"
+- content_for :after_header do
+  = link_to "Back to case", investigation_path(@investigation), class: "govuk-back-link"
+h1.govuk-heading-xl
+  span.govuk-caption-xl
+    / TODO MSPSDS-522 / MSPSDS-524 Get the case Id @investigation.caseId ??
+    = "Case: 1810-0001"
```
broder: I think this is just going to be `@investigation.id` for now

web/app/assets/javascripts/incidents.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
broder: I think we can delete these assets if they're going to be unused.

web/test/system/investigation_incident_test.rb
```diff
@@ -0,0 +1,108 @@
...
+    assert_current_path(/investigations\/\d+/)
+    click_on "Full detail"
+    assert_text "Bad Stuff TM"
+    assert_text "Oh, it was horrible"
+    assert_text "07/12/1984"
+    assert_text "Unsuspecting citizens"
+    assert_text "Kingston Falls"
+  end
+
+  test "can go back to the editing page from the confirmation page and not loose data" do
```
raamSoftwire: lose*

web/test/system/investigation_incident_test.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  test "wizard data doesn't persist between reloads" do
+    fill_in "Incident type", with: "Bad Stuff TM"
+    click_on "Continue"
+    visit new_investigation_incident_path(investigations(:one))
+
+    assert_no_field with: "Bad Stuff TM"
+  end
+
+  test "wizard data gets clear after completion" do
```
raamSoftwire: cleared*

PR #173, opened 2018-10-17
COS-16 cosmetics setup

DWRendell:
Sets up a new rails app for the Cosmetics Portal (name pending) web application.

Also updates the build process to push built containers to DockerHub, and pull those containers into separate test jobs that run in parallel.


.travis.yml
```diff
@@ -50,7 +58,9 @@ jobs:
   - name: Deploy keycloak
     if: branch = master AND type != pull_request
     script: SPACE=int COMPONENT=keycloak ./ci/deploy-if-changed.sh
-
+  - name: Deploy cosmetics
+    if: branch = master AND type != pull_request
+    script: SPACE=int COMPONENT=cosmetics ./ci/deploy-if-changed.sh
```
broder: Should this be `cosmetics-web`?

ci/build-docker-images.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+# Builds Docker images for components that have changes, and then pushes all current images 
+# tagged with the build ID to DockerHub.
+set -ex
+
+# Base name of the respositories for each component.
+REPOSITORY_BASE=davidrendell/beis-mspsds
```
broder: We should probably setup a team account before this goes in

broder: Also, does this need to be set in other files as well?

ci/deploy-if-changed.sh
```diff
@@ -3,6 +3,10 @@ set -ex
 
 if [[ $(./ci/get-changed-components.sh) =~ ((^| )$COMPONENT($| )) ]]; then
     ./$COMPONENT/deploy.sh $SPACE
+    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+    docker pull $REPOSITORY_BASE-$COMPONENT:$REPOSITORY_BASE-$COMPONENT
```
broder: Will this also get triggered for staging and prod builds? Do we need to tag the images differently?

cosmetics-web/deploy.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# USERNAME: cloudfoundry username
+# PASSWORD: cloudfoundry password
+# SPACE: the space to which you want to deploy
+
+./ci/install-cf.sh
+
+cf login -a api.cloud.service.gov.uk -u $USERNAME -p $PASSWORD -o 'beis-mspsds' -s $SPACE
+
+cf push -f ./web/manifest.yml --route-path cosmetics-$SPACE
```
broder: `--route-path` needs to be updated to `--hostname`

ci/build-docker-images.sh
DWRendell: No, it isn't needed to pull from public DockerHub repos.

I'll look into setting up an account using the Softwire DL, unless there's some other email we can use.

broder: I think we can use a BEIS team address - we can chat about it tomorrow if you're at 1VS

cosmetics-web/config/secrets.yml
```diff
@@ -0,0 +1,22 @@
+# Be sure to restart your server when you modify this file.
```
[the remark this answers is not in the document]
DWRendell: Done

ci/deploy-if-changed.sh
DWRendell: I think so, good point though I hadn't really though about that. Probably best we talk over the different environments and build types and decide how best to do this

.travis.yml
```diff
@@ -18,25 +18,29 @@ jobs:
...
   - name: Static Analysis
     install: docker-compose -f docker-compose.yml -f docker-compose.ci.yml pull
-    script: 
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web bin/rubocop
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web bin/slim-lint app/views
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web yarn eslint app/assets/javascripts
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web yarn sass-lint -vq -c .sasslint.yml 'app/assets/stylesheets/**/*.scss'
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web bin/brakeman --no-pager
+    script:
+    - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run web bin/rubocop
```
broder: I think it'll be faster if you do a single `docker-compose -f docker-compose.yml -f docker-compose.ci.yml up -d web` in the `install` bit and then use `docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec web` for the various linting commands.

ci/build-docker-images.sh
```diff
@@ -0,0 +1,38 @@
...
+
+function docker_tag_exists {
+    curl --silent -f -lSL https://index.docker.io/v1/repositories/$1/tags/$2 > /dev/null
+}
+
+echo "Logging into DockerHub"
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+
+for COMPONENT in "${COMPONENTS[@]}"; do
+    if [[ $(./ci/get-changed-components.sh) =~ ((^| )werwerwe($| )) ]]; then
```
broder: Should this be `$COMPONENT` rather than `werwerwe`? Also, maybe we could cache the result of `$(./ci/get-changed-components.sh)` rather than running it for each component?

ci/build-docker-images.sh
```diff
@@ -0,0 +1,38 @@
...
+        echo "Building component $COMPONENT"
+        docker-compose -f docker-compose.yml -f docker-compose.ci.yml build $COMPONENT
+    elif docker_tag_exists $DOCKER_USERNAME/$COMPONENT $BRANCH; then
+        echo "No changes for component $COMPONENT, pulling $BRANCH image"
+        docker pull $DOCKER_USERNAME/$COMPONENT:$BRANCH
+        docker tag $DOCKER_USERNAME/$COMPONENT:$BRANCH $DOCKER_USERNAME/$COMPONENT:$TRAVIS_BUILD_NUMBER
```
broder: I wonder if we should replace `$TRAVIS_BUILD_NUMBER` with `$BUILD_ID` in these scripts. (In the same way that we use `$BRANCH` rather than `$TRAVIS_BRANCH`). I wonder how that would play with the values in `docker-compose.ci.yml` though

ci/deploy-if-changed.sh
```diff
@@ -3,6 +3,10 @@ set -ex
...
+    docker tag $REPOSITORY_BASE-$COMPONENT:$TRAVIS_BUILD_NUMBER $REPOSITORY_BASE-$COMPONENT:master
+    docker push $REPOSITORY_BASE-$COMPONENT:master
```
broder: Does this need to be `$BRANCH` rather than `master`?

ci/deploy-if-changed.sh
```diff
@@ -3,6 +3,10 @@ set -ex
 
 if [[ $(./ci/get-changed-components.sh) =~ ((^| )$COMPONENT($| )) ]]; then
     ./$COMPONENT/deploy.sh $SPACE
+    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
```
broder: I think you could just do `docker login -u $DOCKER_USERNAME -p $DOCKER_PASSWORD` - Travis will correctly mask the variables.

cosmetics-web/config/secrets.yml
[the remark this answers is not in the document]
broder: You've not actually removed this but we actually have this file in the MSPSDS folder anyway so there's no need to get rid of it! The `.gitignore` comment says "Only include if you have production secrets in this file, which is no longer a Rails default".

cosmetics-web/.gitignore
```diff
@@ -0,0 +1,189 @@
...
+
+
+# End of https://www.gitignore.io/api/ruby,node,rails
+
+# These files are copied as part of the worker deploy script
+apt.yml
+clamav/
+public/assets/.sprockets-manifest-qq.json
+
+config/secrets.yml
```
broder: We don't need this line. (Sorry for the confusion!)

cosmetics-web/node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
broder: I don't think we want this file however! It should have been ignored.

ci/build-docker-images.sh
DWRendell: Argh yes, thanks, forgot to fix that. Will try and cache that result too.

ci/deploy-if-changed.sh
DWRendell: Oh yes, good catch

.travis.yml
```diff
@@ -1,80 +1,91 @@
...
     after_script:
-      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
-
+    - docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
+  - name: Static Analysis
+    install: 
+      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml pull
+      - docker-compose -f docker-compose.yml -f docker-compose.ci.yml up -d web
+    script:
+    - docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec web bin/rubocop
```
broder: These commands are failing because it takes a while for the ruby and node dependencies to be installed - adding `docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm --no-deps web echo 'Gems pre-installed'` back into the `install` step might fix this?

PR #174, opened 2018-10-18
MSPSDS-524 provide report reference number

MatthewBurstein:
Add confirmation page and report number to investigations/report wizard.


web/app/controllers/investigations/report_controller.rb
```diff
@@ -31,8 +31,8 @@ def update
     if !@reporter.valid?(step)
       render step
     else
-      redirect_to next_wizard_path if step != steps.last
-      create if step == steps.last
+      create if step == :details
```
jasiekmiko: Hmm, the thing that's special in this case isn't intrinsically linked to the details page, but to the fact that the next page is the confirmation page?
Might be worth using one of the [wizard helpers](https://github.com/schneems/wicked#quick-reference), say
`if next_step?(:confirmation)`

web/app/views/investigations/new_report_confirmation.html.slim
```diff
@@ -0,0 +1,39 @@
+- content_for :page_title, "New product safety report - Case created"
```
jasiekmiko: reminder that this file can go :)

web/app/views/investigations/report/confirmation.html.slim
```diff
@@ -0,0 +1,39 @@
...
+        | You have been sent a confirmation email.
+      hr.govuk-section-break.govuk-section-break--m.govuk-section-break--visible
+  .govuk-grid-row.mspsds-panel
+    .govuk-grid-column-two-thirds
+      h2.govuk-heading-l
+        | Do you have report correspondance?
+      p.govuk-body
+        | If the report came by email, or you've notes / transcript from a call, add them now.
+      p.govuk-body
+        // TODO implement add reporter correspondance functionality
```
jasiekmiko: I think this belongs to MSPSDS-516 :)

web/test/controllers/investigations_controller_test.rb
```diff
@@ -22,19 +22,21 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
...
+  test "should create investigation and redirect to investigation page" do
+    new_investigation_title = "new_investigation_title"
     assert_difference("Investigation.count") do
       post investigations_url, params: {
         investigation: {
-          title: @investigation.title,
+          title: new_investigation_title,
           description: @investigation.description,
           is_closed: @investigation.is_closed,
           source: @investigation.source,
```
jasiekmiko: I'ts a little confusing that we're mixing data from an existing investigation and new data (when the previous data isn't relevant to the test).
You can probably cut the unnecessary fields, since they're not being checked for anyways! (or put ad-hoc data into them)

PR #175, opened 2018-10-18
MSPSDS-514: Remove validation and style text in reporter flow

kicferk1:
In line with comments in here: https://regulatorydelivery.atlassian.net/browse/MSPSDS-514

### DIFF
--- a/web/app/models/reporter.rb
+++ b/web/app/models/reporter.rb
@@ -1,6 +1,6 @@
 class Reporter < ApplicationRecord
   belongs_to :investigation, required: false
   validates :investigation, presence: true, on: %i[create update]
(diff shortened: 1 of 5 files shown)


web/test/controllers/investigations/report_controller_test.rb
```diff
@@ -35,4 +35,16 @@ class ReportControllerTest < ActionDispatch::IntegrationTest
     get new_report_url
     assert_redirected_to report_index_path + '/type'
   end
+
+  test "should pass the data from other type field correctly" do
+    other_type = "Very very other"
```
raamSoftwire: I think the code would be more readable if you used feasible names, eg. Product Regulator.

PR #178, opened 2018-10-22
fix: display pretty id in investigations#index and investigations#show

MatthewBurstein:


web/app/views/application/_investigation_heading.html.slim
```diff
@@ -3,7 +3,7 @@
     h1.govuk-heading-xl
       p.govuk-caption-xl
         'Case:
-        = investigation.id
+        = get_pretty_id(investigation.id)
```
broder: We should maybe put this method in `investigation.rb` (i.e. we'd do `investigation.pretty_id`) but it's not massively important.

PR #183, opened 2018-10-22
Mspsds 570 capture product safety question

kicferk1:
This adds the 'ask a question' flow and refactors various related elements in code to remove code duplication. Namely: all views in report flow, as well as its controller. I also put all standard error lists into a single file while I was at removing duplication.


web/app/views/investigations/show.html.slim
```diff
@@ -12,10 +12,29 @@
 / p
 /   a href="" Edit risk details
 /   a style="margin-left: 20px" href="" View assessment
+
+-if !@investigation.is_case
+  .govuk-grid-row
+    .govuk-grid-column-one-third
+      h2.govuk-heading-l
+        span.govuk-caption-m = 'Asked by ' + (@investigation.reporter&.reporter_type&.presence || 'Anonymous')
```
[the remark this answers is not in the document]
kicferk1: The one thing I'm grateful to linters for is teaching me ruby tricks :D

web/app/views/investigations/show.html.slim
```diff
@@ -12,10 +12,29 @@
...
+        - if @investigation.reporter
```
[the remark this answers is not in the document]
kicferk1: MSPSDS-639 is asking for this section, and I did some exposure anyway to see if my code was working. I buggered Tim and Nick and hopefully I'll get an answer soon

web/app/helpers/investigation_flow_helper.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def clear_session
+    session[:reporter] = {}
+    session[:investigation_id] = nil
+  end
+
+  def load_investigation
+    # TODO: add loading from params if we use the same flow to edit its details
+    @investigation = Investigation.find_by(id: session[:investigation_id]) || default_investigation
```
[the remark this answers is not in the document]
kicferk1: Actually, providing a default_investigation method in the helper works. 
If the class using it will have its own implementation, that implementation is used instead. This solves the problem of exceptions.

web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -0,0 +1,13 @@
+module Investigations::DisplayTextHelper
+  def case_question_lower_case_text(investigation)
+    investigation.is_case ? 'case' : 'question'
+  end
+
+  def case_question_upper_case_text(investigation)
```
broder: You can get this by doing `.titleize` on the method above - does that make it any simpler?

web/app/views/activities/_form.html.slim
```diff
@@ -1,14 +1,6 @@
...
-  - if activity.errors.any?
-    #error_explanation
-      h2.govuk-heading-l
-        = pluralize(activity.errors.count, "error")
-        |  prohibited this activity from being saved:
-      ul.govuk-list
-        - activity.errors.full_messages.each do |message|
-          li
-            = message
+  = render "form_validation_display", validated_object: activity
```
broder: Awesome, thanks!

web/app/views/application/_case_confirmation.html.slim
```diff
@@ -0,0 +1,37 @@
+main#main-content.govuk-main-wrapper role="main"
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      .govuk-panel.govuk-panel--confirmation
+        h2.govuk-panel__title
+          = case_question_upper_case_text(investigation) + " created"
```
broder: Would `"#{case_question_upper_case_text(investigation)} created"` be easier to read?

web/app/views/application/_case_confirmation.html.slim
```diff
@@ -0,0 +1,37 @@
...
+  .govuk-grid-row.mspsds-panel
+    .govuk-grid-column-two-thirds
+      h2.govuk-heading-l
+        = "Do you have " + report_question_lower_case_text(investigation) + "correspondance?"
+      p.govuk-body
+        = "If the " + report_question_lower_case_text(investigation)
+        = " came by email, or you've notes / transcript from a call, add them now."
+      p.govuk-body
+        // TODO MSPSDS-516 implement add reporter correspondance functionality
+        a.govuk-link.mspsds-block-link href="/flows/add-correspondance/01-add-correspondance-context"
```
broder: I don't think we need the `href` here (yet).

web/app/controllers/investigations/question_controller.rb
```diff
@@ -0,0 +1,32 @@
+class Investigations::QuestionController < ApplicationController
+  include InvestigationFlowHelper
+  include Wicked::Wizard
+  steps :questioner_type, :questioner_details, :question_details, :confirmation
```
broder: As discussed, maybe `:confirmation` shouldn't be a step for these creation flows.

PR #184, opened 2018-10-22
MSPSDS-655: fix: some small font fixes

MatthewBurstein:
fix fonts in businesses/index, investigations/index, products/_form and products_suggestions_container

### DIFF
--- a/web/app/views/investigations/businesses/index.html.slim
+++ b/web/app/views/investigations/businesses/index.html.slim
@@ -4,14 +4,16 @@
 = render "investigation_heading", investigation: @investigation
 h2.govuk-heading-l[class="govuk-!-margin-bottom-2"]
   | Manage businesses
-= link_to "Add business", new_investigation_business_path(@investigation)
+p
+  = link_to "Add business", new_investigation_business_path(@investigation)
 div[class="govuk-!-padding-top-8"]
   - if @investigation.businesses.any?
     - @investigation.businesses.each do |business|
       = render layout: "business_summary", locals: { business: business } do
         .govuk-form-group
-          = link_to "Remove", investigation_business_path(@investigation, business), method: :delete, \
-              data: { confirm: "Are you sure?" }
+          p
+            = link_to "Remove", investigation_business_path(@investigation, business), method: :delete, \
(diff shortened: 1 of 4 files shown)


web/app/views/investigations/businesses/index.html.slim
```diff
@@ -4,14 +4,14 @@
 = render "investigation_heading", investigation: @investigation
 h2.govuk-heading-l[class="govuk-!-margin-bottom-2"]
   | Manage businesses
-= link_to "Add business", new_investigation_business_path(@investigation)
+= link_to "Add business", new_investigation_business_path(@investigation), class: "govuk-link govuk-body"
```
broder: Do we need to add `.govuk-body` here too?

web/app/views/products/_suggestions_container.html.slim
```diff
@@ -4,4 +4,5 @@ h3.govuk-heading-m
   - if [@product.gtin, @product.name, @product.brand, @product.product_type].compact.any?
     = render "products/suggested"
   - else
-    | Start filling in the form for suggestions.
+    .govuk-body
```
broder: If we make this a `p` element, `.govuk-body` should be applied automatically by https://github.com/alphagov/govuk-frontend/blob/ad596de756ef51619c6e5bc8f22f3b93479773dd/src/core/_global-styles.scss#L9

web/app/views/investigations/businesses/index.html.slim
MatthewBurstein: .govuk-link does not provide the font-size property so the link looks oddly small without it. Is there a better way to add the property?

broder: Hmmmm, maybe wrap it in a `p` tag? I wonder if that would work everywhere though - might be worth having a quick play with it and checking how other gov pages deal with it (e.g. https://design-system.service.gov.uk/).

MatthewBurstein: The link "read more about what we do" in [this page](https://www.gov.uk/government/organisations/companies-house) is wrapped in a p tag so that seems reasonable.

PR #190, opened 2018-10-24
MSPSDS-559: understand case history

raamSoftwire:
**This is a WIP branch.**

I have started by only looking at Case pages, and will expand this to cover products and businesses after.

I have also started by choosing a single activity to focus on, adding a business to a case. Once the pattern has been reviewed and ironed out, then we can begin to create AuditActivities for all the things we want to track.


web/db/migrate/20181023102910_add_audit_fields_to_activity.rb
```diff
@@ -0,0 +1,10 @@
+class AddAuditFieldsToActivity < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :activities, :title, :string
+    add_column :activities, :subtitle_slug, :string
```
broder: Should we call this `subtitle`?

web/app/views/application/_activities_summary.html.slim
```diff
@@ -3,16 +3,17 @@
...
       - activities.each do |activity|
         li
-          h3.govuk-heading-s[class="govuk-!-margin-bottom-1"]
-            = activity.type.titleize
-          p.timeline-details[class="govuk-!-margin-bottom-1"]
-            ' By
+          h3.govuk-heading-s
+            = activity.title.titleize
+          p.govuk-body-s.timeline-details
+            = activity.subtitle_slug
```
jasiekmiko: would the next few lines be clearer as interpolated text?

web/db/migrate/20181023102910_add_audit_fields_to_activity.rb
jasiekmiko: We added `slug` here to help point out that this isn't the entirety of the subtitle (which is not to say this name couldn't be a bit more pithy :P)

web/app/models/audit_activity.rb
```diff
@@ -0,0 +1,3 @@
+class AuditActivity < Activity
+  validates :description, presence: true
```
jasiekmiko: Not sure how true that is (the https://mspsds-design-prototype.herokuapp.com/case page has a `consumer report` activity, which is doesn't seem to have a body).

Sidenote-ally, since this is not an entity created by hand by the user, not sure validation is the right paradigm? I'd be tempted to put a constraint in the db instead.

web/app/helpers/businesses_helper.rb
```diff
@@ -144,4 +144,14 @@ def match_additional_information(business)
       }
     }
   end
+
+  def create_audit_activity_for_adding_business (business)
+    AuditActivity.create(
+        title: business.company_name,
+        subtitle_slug: "Business added",
+        description: "Role: **Distributor**",
```
jasiekmiko: I recall you were planning on changing `description` to something like `body` or `content`, is that still the plan?

web/app/controllers/investigations/businesses_controller.rb
```diff
@@ -22,21 +22,25 @@ def suggested
 
   # POST /investigations/1/businesses/add
   def add
-    @investigation.businesses << Business.find(params[:business_id])
+    business = Business.find(params[:business_id])
+    @investigation.businesses << business
+    create_audit_activity_for_adding_business business
```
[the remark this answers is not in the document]
jasiekmiko: oooh, shiny✨!
Also looks like a really good page to read to get a feel for the `belongs_to`, `has_one` etc. business overall!

web/app/models/audit_activity.rb
MatthewBurstein: Do we need to validate this at all? I think any appropriate validation will be done in the actual records which activities represent. Have removed.

web/app/controllers/investigations/businesses_controller.rb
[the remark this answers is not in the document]
MatthewBurstein: This is non-trivial because the call to AuditActivity.create uses current_user which is not available in the model. Rails does seem to have [this](https://edgeapi.rubyonrails.org/classes/ActiveSupport/CurrentAttributes.html) but I can't get it to work. If we are to implement callbacks it would be good to pair with someone about it :)

web/app/controllers/images_controller.rb
```diff
@@ -58,8 +60,9 @@ def update
   # DELETE /images/1
   # DELETE /images/1.json
   def destroy
-    @image.purge_later
+    @parent.images.find(params[:id]).delete
```
broder: This could just be `@image.delete`. It might also be worth putting a comment here to say that we're intentionally not purging the blob (as we want a soft delete)

web/app/controllers/images_controller.rb
```diff
@@ -32,6 +32,7 @@ def edit; end
   def create
     respond_to do |format|
       if @image
+        create_audit_activity_for_add_image_to_investigation if @parent.class == Investigation
```
broder: Would `audit_add_image` be a little bit less verbose?

web/app/controllers/images_controller.rb
```diff
@@ -90,4 +97,37 @@ def update_image
   def image_params
     params.require(:image).permit(:file, :title, :description)
   end
+
+  def create_audit_activity_for_add_image_to_investigation
+    title = @image.metadata[:title] || "Untitled image"
+    activity = AddImageAuditActivity.create(
```
broder: I wonder if we should push this logic (and the logic below) into the `AddImageAuditActivity` model and have something like `AddImageAuditActivity.from(investigation: @parent, image: @image)`. We'd have to create the `self.from` method in `add_image_audity_activity.rb` (and `include UserService` in the model too).

I think that would result in a cleaner controller.

web/app/controllers/investigations/businesses_controller.rb
```diff
@@ -22,7 +22,8 @@ def suggested
 
   # POST /investigations/1/businesses/add
   def add
-    @investigation.businesses << Business.find(params[:business_id])
+    business = Business.find(params[:business_id])
```
broder: Out of curiosity, is there a functional difference here?

web/app/helpers/activities_summary_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ActivitiesSummaryHelper
+  def build_subtitle activity
+    "#{activity.subtitle_slug} by #{activity.source.show}, #{activity.created_at.strftime('%C %B %Y')}"
```
broder: Should we just put a `subtitle` method in `activity.rb`?

web/app/views/application/_activities_summary.html.slim
```diff
@@ -1,18 +1,27 @@
...
+              h3.govuk-heading-s
+                = activity.title
+              p.govuk-body-s.timeline-details
+                = build_subtitle(activity)
+              - if activity.attached_image?
+                .govuk-grid-row
+                  .govuk-grid-column-one-third
+                    = image_tag activity.image.variant(resize: "100x100"), alt: activity.image.metadata["title"]
+                  .govuk-grid-colun-two-thirds
+                    .govuk-body
```
broder: `.govuk-body` is applied globally to `p` tags here: https://github.com/alphagov/govuk-frontend/blob/ad596de756ef51619c6e5bc8f22f3b93479773dd/package/core/_global-styles.scss#L10. Should we just do `p` here instead?

web/app/views/businesses/tabs/_activity.html.slim
```diff
@@ -1,3 +1,3 @@
 .govuk-grid-row[class="govuk-!-padding-bottom-6"]
   .govuk-grid-column-full
-    = render "history", versions: @business.versions
+    = "HISTORY TIMELINE WAS HERE. WILL SOON BE MERGED WITH ACTIVITY TIMELINE"
```
broder: Should we just delete this?

web/db/migrate/20181023102910_add_audit_fields_to_activity.rb
```diff
@@ -0,0 +1,10 @@
+class AddAuditFieldsToActivity < ActiveRecord::Migration[5.2]
```
broder: We should probably squash these migrations together before this gets merged - there's no benefit to being able to migrate to a point midway between these migrations.

web/app/controllers/investigations/businesses_controller.rb
MatthewBurstein: No I think in a previous incarnation, business was used again in the method later

web/app/models/audit_activity/document.rb
```diff
@@ -0,0 +1,16 @@
+class AuditActivity::Document < AuditActivity
+  has_one_attached :document
+
+  private_class_method def self.from(document, investigation, title)
```
broder: Does defining a `private` block not work here?

web/app/models/audit_activity/document/update.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+    super(document, investigation, title)
+  end
+
+  def subtitle_slug
+    "Document details updated"
+  end
+
+  def self.no_change? (document, previous_data)
+    document.metadata[:title] == previous_data[:title] && document.metadata[:description] == previous_data[:description]
```
broder: Would https://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-changed-3F be useful here?

web/app/views/application/_activities_summary.html.slim
broder: Any thoughts on this?

web/app/models/activity.rb
```diff
@@ -3,5 +3,12 @@ class Activity < ApplicationRecord
 
   has_one :source, as: :sourceable, dependent: :destroy
 
-  has_paper_trail
+  def attached_image?
+    nil
+  end
+
+  def build_subtitle
```
broder: Should we just call this method `subtitle`?

web/app/views/application/_activities_summary.html.slim
```diff
@@ -1,18 +1,27 @@
...
+          - activities.each do |activity|
+            li
+              h3.govuk-heading-s
+                = activity.title
+              p.govuk-body-s.timeline-details
+                = activity.build_subtitle
+              - if activity.attached_image?
+                .govuk-grid-row
+                  .govuk-grid-column-one-third
+                    = image_tag activity.image.variant(resize: "100x100"), alt: activity.image.metadata["title"]
```
broder: Should we use the `_image` or `_image_tag` partials here to cope with the image being scanned for viruses?

web/db/migrate/20181023102910_add_audit_fields_to_activity.rb
broder: Could still squash these migrations

web/app/views/application/_activities_summary.html.slim
raamSoftwire: I agree, I've now removed this.

web/app/controllers/images_controller.rb
raamSoftwire: Due to moving the models into subdirectories this now reads AuditActivity::Image::Add and is agnostic about the parent type.

web/app/helpers/activities_summary_helper.rb
raamSoftwire: Done.

web/app/helpers/activities_summary_helper.rb
```diff
@@ -0,0 +1,3 @@
+module ActivitiesSummaryHelper
```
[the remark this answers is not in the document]
raamSoftwire: No, I have now removed it.

web/app/models/activity.rb
```diff
@@ -3,5 +3,12 @@ class Activity < ApplicationRecord
...
+    p source
```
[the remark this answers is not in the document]
raamSoftwire: I've removed it now.

web/app/models/activity.rb
raamSoftwire: Yes, I've now renamed it.

web/app/helpers/businesses_helper.rb
raamSoftwire: I've now renamed this `body`.

web/app/models/audit_activity/document/update.rb
raamSoftwire: We think this only works for ActiveRecord, not ActiveStorage, so we have committed a work-around in commit `5be6cb1`

web/app/models/investigation.rb
```diff
@@ -52,26 +52,29 @@ def pretty_id
...
     AddCaseAuditActivity.create(
-        source: UserSource.new(user: current_user),
-        investigation: self)
+      source: UserSource.new(user: current_user),
+      investigation: self
+)
   end
 
   def create_audit_activity_for_product product
     AddProductAuditActivity.create(
```
broder: Should we do this with a `.from()` method too to make it a bit neater?

web/app/models/investigation.rb
```diff
@@ -52,26 +52,29 @@ def pretty_id
...
   end
 
   def create_audit_activity_for_business business
     AddBusinessAuditActivity.create(
-        description: "Role: **Distributor**",
-        business: business,
-        source: UserSource.new(user: current_user),
-        investigation: self,
-        title: business.company_name)
+      body: "Role: **Distributor**",
```
broder: We've actually got the role now on `InvestigationBusiness`

web/app/models/audit_activity/document.rb
raamSoftwire: Sadly not. From what I understand the private block works for instance methods but not class methods.

PR #191, opened 2018-10-24
Mspsds 516 report initial case correspondence

kicferk1:
There might well be merge conflicts or logical inconsistencies between this and MSPSDS-570, I'd rather have 570 merged in first.

### DIFF
--- a/web/app/controllers/investigations/correspondence_controller.rb
+++ b/web/app/controllers/investigations/correspondence_controller.rb
@@ -4,7 +4,6 @@ class Investigations::CorrespondenceController < ApplicationController
 
   def new
     clear_session
-    session[:investigation_id] = params[:investigation_id]
     redirect_to wizard_path(steps.first, request.query_parameters)
   end
 
@@ -18,7 +17,6 @@ class Investigations::CorrespondenceController < ApplicationController
 
   def show
     load_investigation_and_correspondence
-    @correspondence = Correspondence.new(session[:correspondence])
     render_wizard
   end
 
@@ -59,7 +57,7 @@ private
   end
 
   def load_investigation_and_correspondence
-    @investigation = Investigation.find_by(id: session[:investigation_id])
+    @investigation = Investigation.find_by(id: params[:investigation_id])
     load_correspondence
   end
 
@@ -97,6 +95,5 @@ private
 
   def clear_session
     session[:correspondence] = nil
-    session[:investigation_id] = nil
   end
 end
--- a/web/app/controllers/investigations/correspondence_controller.rb
+++ b/web/app/controllers/investigations/correspondence_controller.rb
@@ -1,0 +1,59 @@
+class Investigations::CorrespondenceController < ApplicationController
+  include Wicked::Wizard
+  steps :surface, :content, :confirmation
+
+  # GET /investigations/report/new
+  def new
+    session[:investigation_id] = params[:investigation_id]
+    redirect_to wizard_path(steps.first, request.query_parameters)
+  end
+
+  # POST /investigations/report
+  def create
+    update_partial_correspondence
+    @investigation.correspondences << @correspondence
+    @investigation.save
+    redirect_to investigation_path(@investigation)
+  end
+
+  # GET /investigations/report
+  # GET /investigations/report/step
+  def show
+    update_partial_correspondence
+    @correspondence = Correspondence.new(session[:correspondence])
+    render_wizard
+  end
+
+  def update
+    update_partial_correspondence
+    if !@correspondence.valid?(step)
+      render step
+    else
+      create if next_step? :confirmation
+      redirect_to next_wizard_path
+    end
+  end
+
+private
+
+  def correspondence_params
+    return {} if !params[:correspondence] || params[:correspondence] == {}
+    if params[:correspondence][:correspondent_type] == 'Other'
+      params[:correspondence][:correspondent_type] = params[:correspondence][:other_correspondent_type]
+    end
+    params.require(:correspondence).permit(
+      :correspondent_name, :correspondent_type, :contact_method, :phone_number, :email_address, :correspondence_date,
+      :overview, :details
+    )
+  end
+
+  def update_partial_correspondence
+    p '========================='
+    p session[:investigation_id]
+    @investigation = Investigation.find_by(id: session[:investigation_id])
+    p @investigation
+    session[:correspondence] = (session[:correspondence] || {}).merge(correspondence_params)
+    @correspondence = Correspondence.new(session[:correspondence])
+    session[:correspondence] = @correspondence.attributes
+  end
+end
--- a/web/app/controllers/investigations/correspondence_controller.rb
+++ b/web/app/controllers/investigations/correspondence_controller.rb
@@ -2,34 +2,31 @@ class Investigations::CorrespondenceController < ApplicationController
   include Wicked::Wizard
   steps :surface, :content, :confirmation
 
-  # GET /investigations/report/new
   def new
+    clear_session
     session[:investigation_id] = params[:investigation_id]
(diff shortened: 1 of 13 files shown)


web/app/views/investigations/tabs/_details.html.slim
```diff
@@ -43,3 +43,17 @@ table.govuk-table
         td.govuk-table__cell = @investigation.reporter&.other_details
 p = link_to "Record an incident", new_investigation_incident_path(@investigation), class: "no-print"
 p = link_to "Edit details", edit_investigation_path(@investigation), class: "no-print"
+
+- if @investigation.correspondences.any?
```
jasiekmiko: This is extra to designs, isn't it? I'm a bit worried about things just being dumped in here without much of a plan (guilty of starting the trend!).
At least I'd hope for it to match the rest of the tab with the table design, but also worth checking with Tim/Nick and putting a TODO here indicating what the future plan for this is (whether that's changing it or removing it once the activities are ready)

web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -0,0 +1,100 @@
...
+        correspondent_name: @reporter.name,
+        contact_method: get_contact_method,
+        phone_number: @reporter.phone_number,
+        email_address: @reporter.email_address
+      )
+    end
+    values
+  end
+
+  def get_contact_method
```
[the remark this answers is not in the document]
jasiekmiko: also, `contact_method` is a persisted field. do we need this manual re-creation?

web/test/system/add_correspondence_from_flow_test.rb
```diff
@@ -0,0 +1,69 @@
+require "application_system_test_case"
+
+class AddCorrespondenceFromFlowTest < ApplicationSystemTestCase
+  setup do
+    sign_in_as_admin
+    visit root_path
+    click_on "Report an unsafe product"
```
jasiekmiko: Seems like this is a bit of a round-about way of creating an investigation and going to the right page.
How about:
1. adding an investigation to the db (the fixtures ones might be in the db already?)
2. visiting `new_investigation_correspondence_path(investigation)`

web/test/system/add_correspondence_from_flow_test.rb
```diff
@@ -0,0 +1,69 @@
...
+  teardown do
+    logout
+  end
+
+  test "first step should be type" do
+    assert_text("Who is the correspondance with?")
+  end
+
+  test "first step should be populated with reporter details" do
+    assert_equal('Ben', find_field('correspondence[correspondent_name]').value, '')
```
jasiekmiko: If you're not providing a message (third parameter) then let's just not specficy it - it defaults to `nil` (ditto for assertions below!)

web/test/system/add_correspondence_from_flow_test.rb
```diff
@@ -0,0 +1,69 @@
...
+  def select_type_and_continue
+    choose("reporter[reporter_type]", visible: false, match: :first)
+    click_button "Continue"
+  end
+
+  def fill_name_and_continue
+    fill_in("reporter[name]", with: "Ben")
+    click_button "Continue"
+  end
+end
```
jasiekmiko: Let's have a test that checks the included values got recorded and get surfaced, too? (not just that the wizard completes)

web/app/controllers/investigations/correspondence_controller.rb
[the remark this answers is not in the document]
kicferk1: This re-creation should only be useful for pre-populating the field. So I would say we need something like this for the first step, possibly not for latter ones.

web/app/views/investigations/tabs/_details.html.slim
kicferk1: I updated the style to follow the table design. 

MSPSDS-620 is a ticket better suited for providing final styling.

web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -0,0 +1,100 @@
+class Investigations::CorrespondenceController < ApplicationController
+  include Wicked::Wizard
+  steps :surface, :content, :confirmation
```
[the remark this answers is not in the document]
kicferk1: The part of correspondence that is not content xD
Naming is hard, I re-named it to general_info, open to suggestions for a better one

web/app/helpers/date_helper.rb
```diff
@@ -0,0 +1,39 @@
+module DateHelper
+  def set_keys(date_key)
```
broder: Maybe call this `set_date_key`?

web/app/helpers/date_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+    # TODO: remove if we enforce convention of storing all data inputs as 'date'
+    @date_key = date_key
+  end
+
+  def helper_after_initialize
```
broder: Could we instead do something like
```
included do
  after_initialize do
    # contents from helper_after_initialize
  end
end
```
?

web/app/views/investigations/correspondence/confirmation.html.slim
```diff
@@ -29,4 +29,4 @@ main#main-content.govuk-main-wrapper role="main"
               td.govuk-table__cell
         button.govuk-button type="submit"
           | Continue
-        a.govuk-link.mspsds-block-link href=wizard_path(:surface)  Edit details
+        a.govuk-link.mspsds-block-link href=wizard_path(:general_info)  Edit details
```
broder: Should we use `link_to` for this?

web/app/views/investigations/tabs/_details.html.slim
```diff
@@ -44,16 +44,29 @@ table.govuk-table
...
+        table.govuk-table
+          tbody.govuk-table__body
+            | Correspondence details:
+            tr.govuk-table__row
+              th.govuk-table__header scope="row" Correspondent name
+              td.govuk-table__cell = correspondence.correspondent_name
+            tr.govuk-table__row
+              th.govuk-table__header scope="row" Contact method
+              - if correspondence.contact_method == 'Email'
+                td.govuk-table__cell = correspondence.contact_method + ': ' + correspondence.email_address
```
broder: Could do this with string interpolation instead

web/app/models/investigation.rb
```diff
@@ -26,6 +26,8 @@ class Investigation < ApplicationRecord
 
   has_many :incidents, dependent: :destroy
 
+  has_many :correspondences, dependent: :destroy
```
broder: Should correspondence entries actually be an activity? Maybe a communication activity as defined here: https://regulatorydelivery.atlassian.net/wiki/spaces/MSPSDS/pages/538608077/Recording+investigation+activity+types+model

web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -0,0 +1,93 @@
...
+  include Wicked::Wizard
+  steps :general_info, :content, :confirmation
+
+  def new
+    clear_session
+    redirect_to wizard_path(steps.first, request.query_parameters)
+  end
+
+  def create
+    load_investigation_and_correspondence
```
broder: Also, you can do this kind of stuff with `before_action` e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/app/controllers/products_controller.rb#L7

web/app/views/application/_date_input.html.slim
```diff
@@ -0,0 +1,35 @@
+span#incident-date-hint.govuk-hint
```
broder: This would be useful for `products/_form.html.slim` too!

PR #192, opened 2018-10-24
MSPSDS 545 assign case by name

MatthewBurstein:


web/app/controllers/investigations_controller.rb
```diff
@@ -37,15 +37,21 @@ def status; end
   # GET /investigations/1/assign
   def assign
     redirect_to investigation_path(@investigation) if @investigation.is_closed
-    @assignee = @investigation.assignee
+    @investigation
```
broder: Do we need this line?

web/app/controllers/investigations_controller.rb
```diff
@@ -37,15 +37,21 @@ def status; end
...
   end
 
   # POST /investigations/1/update_assignee
   def update_assignee
-    assignee = User.find_by(email: params[:email].downcase)
+    return redirect_to assign_investigation_path if request.get?
```
broder: Would it be simpler to remove this line and remove the `get :update_assignee` from `routes.rb`?

web/app/controllers/investigations_controller.rb
```diff
@@ -37,15 +37,21 @@ def status; end
...
+
+    assignee = User.find_by(id: params[:assignee_id])
+    if !assignee
+      @investigation.errors.add(:assignee, "must not be left blank")
+      return render :assign
```
broder: I don't think we should return the result of the `render` (i.e. `render :assign` and `return` should be on separate lines). In other places, we don't use `return` and just do an `if...else` - should we do that for consistency?

Also, should we use `respond_to` here to support API users? It's not strictly necessary but might be useful in the future.

web/test/test_helper.rb
```diff
@@ -18,18 +18,20 @@ class ActiveSupport::TestCase
 
   # Setup all fixtures in test/fixtures/*.yml for all tests in alphabetical order.
   fixtures :all
-
   # Add more helper methods to be used by all tests here...
+
   def sign_in_as_admin
-    admin = { sub: "admin", email: "admin@example.com", given_name: "First", family_name: "Last" }
-    stub_user_credentials(user: admin, is_admin: true)
+    user = build_admin_user :keycloak
```
broder: Are these changes required anymore?

web/app/controllers/investigations_controller.rb
MatthewBurstein: If you remove that line then a GET request to the route shows the rails routing error page. Is this something that will eventually be handled generally for the application so doesn't need to be handled here?

web/app/controllers/investigations_controller.rb
MatthewBurstein: This makes sense. How would you use respond_to here? The case when the POST works already does when save_and_respond is called. If assignee doesn't exist how would you expect an API to respond? Should it be with "Unprocessable entity" as when the call to save fails?

web/test/test_helper.rb
MatthewBurstein: Which changes do you mean? We still need to stub the User data.

web/app/controllers/investigations_controller.rb
broder: It will eventually be handled generally.

web/app/controllers/investigations_controller.rb
broder: Maybe something like:
`  @investigation.assignee = User.find_by(id: params[:assignee_id])
    respond_to do |format|
      if @investigation.assignee && @investigation.save
        format.html { redirect_to @investigation, notice: "Assignee was successfully updated." }
        format.json { render :show, status: :ok, location: @investigation }
        NotifyMailer.assigned_investigation(@investigation, @investigation.assignee.full_name, @investigation.assignee.email).deliver_later
      else
        @investigation.errors.add(:assignee, "must not be left blank")
        format.html { render :assign }
        format.json { render json: @investigation.errors, status: :unprocessable_entity }
      end
    end`?

web/test/test_helper.rb
broder: Wasn't it being stubbed before though? I'd expect if you undid all the changes in `test_helper.rb`, your tests will still pass.

MatthewBurstein: If I don't stub KeycloakClient#all_users I get the following error, and in order to stub the method I need to create the two different kinds of mocked users:
`Error:
InvestigationsControllerTest#test_should_assign_user_to_investigation:
Errno::EADDRNOTAVAIL: Failed to open TCP connection to localhost:80 (Cannot assign requested address - connect(2) for "localhost" port 80)
    app/clients/keycloak_client.rb:23:in `block in all_users'
    app/clients/keycloak_client.rb:22:in `all_users'
    app/models/user.rb:18:in `all'
    test/controllers/investigations_controller_test.rb:80:in `block in <class:InvestigationsControllerTest>'`

web/app/controllers/investigations_controller.rb
MatthewBurstein: Yes this makes sense thanks :)

web/test/test_helper.rb
broder: Ah, fair enough. In that case, this is almost there - I'd prefer if we stubbed the `Keycloak::Internal.get_users` method rather than the method on our `KeycloakClient` class. We can also generate the UUIDs (rather than hardcoding them) using `SecureRandom.uuid`.

MatthewBurstein: Thanks, have done so. Is this because we only want to stub the module, and not code that we have written that wraps the module?

broder: Yeah!

PR #201, opened 2018-10-30
MSPSDS-521: Record case images

kicferk1:
'Soft delete' and 'history' part of this ticket are blocked by activities
Most of the work here was styling and re-writing image controller so it uses wizard.


web/app/controllers/images_controller.rb
```diff
@@ -73,21 +72,48 @@ def set_parent
...
 
   def update_image
     @image.blob.metadata.update(image_params)
     @image.blob.metadata["updated"] = Time.current
   end
 
+  def validate
+    session[:errors] = nil
+    if image_params[:title].blank? && step != :step_upload
+      session[:errors] = (session[:errors] || []).push(field: "title", message: "Title can't be blank")
```
broder: Could we store the errors on `@errors` instead? Do they need to be persisted between pages or just between here and rendering the view?

web/app/views/application/_images_summary.html.slim
```diff
@@ -1,13 +1,25 @@
 - if parent.images.attached?
-  h2.govuk-heading-l[class="govuk-!-margin-bottom-1"]
-    = "Images (#{parent.images.size})"
-  - parent.images.attachments.limit(2).each do |image|
-    = render "images/image", image: image, dimensions: "300x200"
+  / TODO: Pagination if we decide it's needed
```
broder: I don't think we need this `TODO` - it'll probably come up through usability testing so we don't need to remember it here.

web/app/views/application/_form_validation_display.html.slim
```diff
@@ -7,3 +8,14 @@
       - validated_object.errors.full_messages.each do |message|
         li
           = message
+
+/ For our own validation, active storage doesn't allow us to add errors to its objects
+- elsif defined?(errors) && errors&.any?
```
broder: If we set our errors to be something like `errors = ActiveModel::Errors.new(ActiveStorage::Blob.new)` and then add errors using `errors.add(:base, :title_not_implemented, message: "title must be implemented")` then we wouldn't need this branch. It's maybe a little bit hacky though... What do you think?

web/app/controllers/images_controller.rb
```diff
@@ -1,10 +1,12 @@
 class ImagesController < ApplicationController
   include ImagesHelper
+  include Wicked::Wizard
+  steps :step_upload, :step_metadata
+  skip_before_action :setup_wizard, only: %i[edit destroy]
+  skip_before_action :verify_authenticity_token, only: :create
```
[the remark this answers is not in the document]
kicferk1: Skip setup_wizard: because if we don't, it expects a step that matches the method. The only suggested solutions I found online were creating a separate controller or a skip. 

Skip verification was a hacky solution that is luckily no longer needed :)

web/app/controllers/images_controller.rb
```diff
@@ -19,75 +21,100 @@ def index
...
 
   def update_image
     @image.blob.metadata.update(image_params)
     @image.blob.metadata["updated"] = Time.current
   end
 
+  def validate
+    session[:errors] = nil
+    if (image_params[:title].blank? && step != :step_upload)
+      session[:errors] = (session[:errors] || []).push({field: "title", message: "Title can't be blank"})
```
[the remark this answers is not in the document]
kicferk1: I just tried it and it seems using @errors is enough to stop you from saving an image without title, but not enough to display a meaningful error message. 

@errors is empty whenever a page is rendered :(

web/app/views/application/_form_validation_display.html.slim
kicferk1: The problem might be persistence. Currently validation happens in create, and if data is invalid the user gets redirected back. The wizard related flow can probably be redesigned so that validation happens in update, and so we don't lose errors. But the editing flow that happens outside wizard makes it really hard to validate without redirecting and losing the errors in the process. 
If we can't get around that, we would need to store some trace of errors in session and re-create them each time we load a page. 
Definite drawback is that we would have a method for validating and another one for retrieving and loading the result of that validation. 
Definite advantage is that this could likely be reused in any place that for some reason needs controller-level validation. 
And we would get rid of a branch in this file. Not sure what's optimal tbh

web/app/views/application/_form_validation_display.html.slim
```diff
@@ -1,9 +1,10 @@
-- if validated_object&.errors&.any?
+/ For active record validation
```
broder: I don't think this comment is relevant anymore.

web/app/controllers/images_controller.rb
```diff
@@ -2,92 +2,54 @@ class ImagesController < ApplicationController
...
-  end
-
-  def update_image
-    @image.blob.metadata.update(image_params)
-    @image.blob.metadata["updated"] = Time.current
+    if params[:id].present?
+      @image = @parent.images.find(params[:id])
+      @image_blob = @image.blob
+      session[:image_blob_id] = @image_blob.id
+    elsif session[:image_blob_id]
```
broder: Can we hit this condition on this controller anymore?

web/app/views/images/_metadata_form.html.slim
```diff
@@ -1,10 +1,23 @@
...
-  .govuk-form-group
-    = form.label :title, "Image title", class: "govuk-label"
-    = form.text_field :title, value: image.metadata["title"], class: "govuk-input"
-  .govuk-form-group
-    = form.label :description, class: "govuk-label"
-    = form.text_area :description, value: image.metadata["description"], rows: 5, class: "govuk-textarea"
-  .govuk-form-group
-    = form.submit "Save", class: "govuk-button"
+= render "page_heading", title: edit ? "Edit image details" : "Enter image details"
+- url = edit ? associated_image_path(@parent, image) : wizard_path
```
broder: Should we just pass `url` in as the parameter from the parent views rather than the `edit` bool? Then we wouldn't need the conditional here

web/app/controllers/images_controller.rb
```diff
@@ -2,92 +2,54 @@ class ImagesController < ApplicationController
...
+  before_action :set_image, only: %i[edit update create]
 
   # GET /images/1/edit
   def edit; end
 
+  # PATCH/PUT /images/1
+  # PATCH/PUT /images/1.json
+  def update
+    validate
+    return render if @errors.present?
```
broder: If we return `render :edit` here, then we don't need both `edit.html.slim` and `update.html.slim`

web/app/views/images_flow/metadata.html.slim
```diff
@@ -0,0 +1,4 @@
+- content_for :page_title, "Edit image - Market Surveillance & Product Safety"
```
broder: We don't need the `Market Surveillance` bit in the title anymore.

web/app/views/images_flow/upload.html.slim
```diff
@@ -0,0 +1,11 @@
+- content_for :page_title, "New image"
+- content_for :after_header do
+  = link_to "Back", @parent, class: "govuk-back-link"
+= render "page_heading", title: "New image"
+br
```
broder: Do we need a break at the start? I would have assumed that the page template would do that for us

web/config/routes.rb
```diff
@@ -5,7 +5,11 @@
   end
 
   concern :image_attachable do
-    resources :images
+    resources :images, controller: "images" do
+      collection do
+        resources :image_flow, controller: "images_flow", only: %i[show new create update]
```
broder: Should we call it `wizard` or `new` or something? Then the routes would be `/investigations/1/images/wizard/step` or `/investigations/1/images/new/step` rather than `/investigations/1/images/images_flow/step`

web/app/views/images/_metadata_form.html.slim
kicferk1: edit also influences the title(which is not too important) and whether we show a delete button

I guess we could pass url to the form as a separate argument though

PR #208, opened 2018-10-31
MSPSDS-673 Upload risk assessment

samlionheart:


raamSoftwire:
I have done some work with Michal on this, and you get a hazard activity on the timeline, with a link to the risk assessment. However, work still needs to be done to update a hazard and display it correctly on the timeline.

web/app/controllers/investigations/hazards_controller.rb
```diff
@@ -1,17 +1,24 @@
...
   def new
-    session[:hazard] = nil
+    clear_session
+    initialize_file_attachment
     redirect_to wizard_path(steps.first, request.query_parameters)
   end
 
-  # GET /investigations/1/hazards/step
+  def create
+    attach_file_to_attachment_slot(@file, @hazard.risk_assessment)
```
kicferk1: Possibly not relevant to this particular ticket, but we might need designs for how to approach editing file attachments. 

For example, this will never remove a file, because 1) it should be loaded from hazard, and 2) the method checks the file exists before doing anything. Do we expect to need to delete a file? Should we also show any presently attached file and in what way?

web/app/views/investigations/hazards/risk_level.slim
```diff
@@ -7,14 +7,16 @@ h1.govuk-heading-xl
...
-        | Risk level
-        .govuk-radios
+      .govuk-radios
         - Hazard.risk_levels.each do |(level, _)|
           .govuk-radios__item
             = form.radio_button :risk_level, level, class: "govuk-radios__input"
             = form.label :risk_level, level.humanize.capitalize, class: "govuk-label govuk-radios__label"
-
+        h4.govuk-fieldset__legend--l
+          | Add risk assessment
```
kicferk1: Since this ticket has editing functionality, should we indicate around here that a file is already attached(if it is) and that uploading a new one will override it?

web/app/controllers/investigations/hazards_controller.rb
```diff
@@ -1,17 +1,24 @@
...
+    @investigation.hazard = @hazard
+    @investigation.save
+    activity = AuditActivity::Hazard::Add.from(@hazard, @investigation)
```
kicferk1: This line will be executed when hazard is edited, leading to a wrong activity being created :( 

This controller handles creating a hazard(through wizard flow), editing a hazard(through a very similar wizard flow that actually should be slightly different), and from what it seems a slightly different edit dealing just with updating risk level. It might be best to split it, we can pair up on that if you want

web/app/views/application/_activities_summary.html.slim
```diff
@@ -21,6 +21,9 @@
                   = markdown activity.body if activity.body
                   = link_to "View business details", business_url(activity.business_id) if activity.business_id
                   = link_to "View product details", product_url(activity.product_id) if activity.product_id
+                  - if activity.is_a?(AuditActivity::Hazard::Base) && activity&.risk_assessment&.attached?
```
broder: Can `activity` be nil here? Do we need `activity?.risk_assessment`?

web/app/controllers/concerns/file_concern.rb
```diff
@@ -35,6 +35,17 @@ def attach_file_to_list(file, attachment_list)
     end
   end
 
+  def attach_file_to_attachment_slot(file, attachment_slot)
+    if file
+      update_file_details(file)
+      if attachment_slot.attached?
+        attachment_slot.detach
```
broder: What's the default behaviour when we attach something? i.e. if we don't detach first, what happens?

web/app/controllers/concerns/file_concern.rb
```diff
@@ -35,6 +35,17 @@ def attach_file_to_list(file, attachment_list)
     end
   end
 
+  def attach_file_to_attachment_slot(file, attachment_slot)
+    if file
```
broder: We could reduce the amount of nesting by starting this with `return unless file`

web/app/controllers/investigations/hazards/edit_hazard_flow_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+private
+
+  def preload_hazard(investigation)
+    @hazard = investigation.hazard
+    @hazard = Hazard.new if @hazard.blank?
+    session[:hazard] = (@hazard&.attributes || {}).merge(session[:hazard] || {})
+  end
+
+  def create_hazard_audit_activity
+    AuditActivity::Hazard::Update.from(@hazard, @investigation)
```
broder: Should we do this in the hazard model?

web/app/views/investigations/hazards/_risk_level.slim
```diff
@@ -0,0 +1,16 @@
+h3.govuk-fieldset__legend--l
+  | Risk level
+.govuk-form-group
+  .govuk-radios
+    - Hazard.risk_levels.each do |(level, _)|
+      .govuk-radios__item
+        = form.radio_button :risk_level, level, class: "govuk-radios__input"
+        = form.label :risk_level, level.humanize.capitalize, class: "govuk-label govuk-radios__label"
```
broder: I think we could just use `level.titleize` here.

PR #214, opened 2018-11-01
MSPSDS-648: Fix reporter details bug

kicferk1:


web/app/controllers/investigations/flow_controller.rb
```diff
@@ -41,14 +41,21 @@ def load_reporter
...
       :name, :phone_number, :email_address, :reporter_type, :other_details
     )
   end
 
+  def handle_reporter_type
+    if params[:reporter][:reporter_type] == 'Other'
+      params[:reporter][:reporter_type] = params[:reporter][:other_reporter].presence || 'Other'
+    end
+    if params[:reporter][:reporter_type].blank?
+      params[:reporter][:reporter_type] = 'Person'
```
broder: I feel like we should just make this mandatory rather than having a hidden extra option...

kicferk1: So do I, but the ticket doesn't :( 
https://regulatorydelivery.atlassian.net/secure/RapidBoard.jspa?rapidView=18&projectKey=MSPSDS&modal=detail&selectedIssue=MSPSDS-648

broder: Fine...

PR #215, opened 2018-11-02
MSPSDS-631,MSPSDS-632,MSPSDS-633,MSPSDS-634: Update business and product details

broder:


web/app/controllers/investigations/businesses_controller.rb
```diff
@@ -17,13 +14,13 @@ def new
...
     advanced_search(excluded_business_ids)
-    render partial: "businesses/search_results"
+    render partial: "businesses/suggested"
   end
 
-  # POST /investigations/1/businesses/add
-  def add
-    @investigation.businesses << Business.find(params[:business_id])
-    redirect_to_investigation_businesses_tab "Business was successfully added."
+  # PUT /investigations/1/businesses/2/link
```
broder: Maybe I should map this route to e.g. `/investigations/1/businesses/2` given that we don't have an update on this controller

web/app/assets/stylesheets/block-link.scss
```diff
@@ -1,6 +1,7 @@
 .mspsds-block-link {
   display: inline-block;
-  @include govuk-responsive-margin(7, "bottom");
+  @include govuk-responsive-margin(2, "top");
+  @include govuk-responsive-margin(2, "bottom");
```
jasiekmiko: Is it worth keeping these in line with the design values as much as possible?
https://github.com/UKGovernmentBEIS/beis-mspsds-design-prototype/blob/726aff35d5b7ba053711dce468e9e2da45fbe7bb/app/assets/sass/_block-link.scss#L3

web/app/controllers/investigations/businesses_controller.rb
jasiekmiko: Yeah, that seems better, otherwise it loos a bit like a link entity on an investigation business :)

web/app/views/businesses/_suggested.html.slim
```diff
@@ -2,12 +2,12 @@ h3.govuk-heading-s
...
-      - if @investigation.present?
-        div[class="govuk-!-margin-top-3"]
-          = form_with(url: add_investigation_businesses_path(@investigation), local: true) do |form|
-            = form.hidden_field :business_id, value: business.id
-            = form.submit "Add", class: "govuk-button"
+    = render "business_summary", business: business do
+      p
+        = link_to "View business", business, class: "govuk-link mspsds-block-link"
+        - if @investigation.present?
+          = link_to "Add", link_investigation_business_path(@investigation, business), method: :put, \
```
jasiekmiko: The [rails magic](https://api.rubyonrails.org/v5.2.1/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) which allows you to specify the `method: :put` modifier relies on javascript to work.
For that, it's probably worth leaving the more verbose `form`-based form here, perhaps encapsulating it in a partial if needed?

broder: It will require JavaScript to work - I might leave it for now and we can pick it up again if / when we need to support JavaScript-disabled users?

web/app/assets/stylesheets/block-link.scss
broder: I think we should use it as a reference but maybe not follow it exactly.

web/app/views/businesses/_suggested.html.slim
jasiekmiko: I thought that's one of the principles we are trying to follow as the baseline? tbh i can't recall where i got this idea from, but good to know which way it is :P

PR #218, opened 2018-11-05
MSPSDS-643 see case title

MatthewBurstein:


web/app/models/investigation.rb
```diff
@@ -83,6 +95,29 @@ def send_assignee_email
...
+
+  def get_property_value_if_shared property_name
+    first_product = products.first
+    if products.all? { |product| product[property_name] == first_product[property_name] }
+      first_product[property_name]
+    end
+  end
+
+  def build_string_from_array array, separator
+    array.reject(&:blank?).join(separator)
```
broder: I think it would be easier to read if we inlined this (such that `def case_title` is something like
```
[build_title_products_portion, build_title_hazard_portion].reject(&:blank?).join(" - ").presence || "Untitled case"
```
)

web/app/models/investigation.rb
```diff
@@ -6,8 +6,9 @@ class Investigation < ApplicationRecord
   include Searchable
   include Documentable
   include UserService
+  include Investigations::DisplayTextHelper
```
broder: Are we actually using this import?

web/app/controllers/investigations_controller.rb
```diff
@@ -91,7 +91,7 @@ def set_investigation
   # Never trust parameters from the scary internet, only allow the white list through.
   def investigation_params
     params.require(:investigation).permit(
-      :title, :description, :image, :risk_level, :is_closed,
+      :question_title, :description, :image, :risk_level, :is_closed,
```
broder: Does the "Begin an investigation" link on the case list page still work? We should maybe update it as part of this ticket.

web/app/models/investigation.rb
MatthewBurstein: Methods from this module are used in `views/application/_investigation_heading.html.slim` and `views/application/_case_confirmation.html.slim`. This line came in when I rebased onto master before pushing

web/app/controllers/investigations_controller.rb
MatthewBurstein: Clicking this link takes you to a form where the only field is a title for your investigation. The information you enter here is not saved anywhere at the moment. I don't really understand what this link is supposed to be for now that case titles are generated programatically. What is the expected behaviour?

MatthewBurstein:
This branch is behaving weirdly because the tests which are failing don't exist on my local copy. I'm closing this PR and opening another on a branch where they do.

web/app/controllers/investigations_controller.rb
broder: https://mspsds-design-prototype.herokuapp.com/flows/begin-investigation/01-begin-investigation - it should ask you for a few details which are then added to the case created activity (probably as the description).

PR #220, opened 2018-11-05
MSPSDS-559: add activities for incidents and hazards

raamSoftwire:


web/app/models/audit_activity/hazard.rb
```diff
@@ -0,0 +1,17 @@
...
+     self.create(
+         body: body,
+         source: UserSource.new(user: current_user),
+         investigation: investigation,
+         title: hazard.hazard_type,
+         )
+   end
+
+  def self.build_body(hazard)
+    "#{hazard.description} <br><br>
```
broder: It's a bit horrendous that we're putting HTML in here... Looks like there are a few options (from https://gist.github.com/shaunlebron/746476e6e7a4d698b373)
* Two spaces at the end of the line (works but it's not obvious how (could we put a \ at the end of the line to show the spaces?))
* `<br/>`
* Putting two lines between the text - this turns them into `p` elements though (which adds extra spacing)
Which do you prefer?

Also, I think the indentation is actually being captured here i.e. it should be:
```
  def self.build_body(hazard)
    "#{hazard.description}<br><br>
Risk level: **#{hazard.risk_level}**<br>
Vulnerable group: **#{hazard.affected_parties}**"
  end
```

broder: I don't think `br`s are particularly well defined in HTML so either `<br>` or `<br/>` should work.

https://stackoverflow.com/a/1726103 also suggests that we shouldn't be using `br`s here at all - might be worth chatting to Nick about it

PR #221, opened 2018-11-05
MSPSDS-643 case title markups

MatthewBurstein:


web/app/models/audit_activity/investigation/add.rb
```diff
@@ -4,6 +4,6 @@ def self.from(investigation)
   end
 
   def subtitle_slug
-    nil
+    body
```
broder: Is there a chance that the `body` could be really long? We might not want that in the subtitle if it could be.

web/app/models/investigation.rb
```diff
@@ -6,8 +6,9 @@ class Investigation < ApplicationRecord
   include Searchable
   include Documentable
   include UserService
+  include Investigations::DisplayTextHelper
```
broder: Do we need this import?

web/app/views/investigations/_form.html.slim
```diff
@@ -1,11 +1,16 @@
...
-    = form.text_field :title, class: "govuk-input #{'govuk-input--error' if investigation.errors.any?}"
-  .govuk-form-group
-    = form.submit "Save", class: "govuk-button"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with(model: investigation, local: true) do |form|
+      fieldset.govuk-fieldset
+        legend class=("govuk-fieldset__legend govuk-fieldset__legend--l")
+          | Creation details
+      - if investigation.errors.any?
```
broder: I think the `form_validation_display` partial does the `if` as well - you might not need both.

web/app/controllers/investigations_controller.rb
```diff
@@ -91,7 +91,7 @@ def set_investigation
   # Never trust parameters from the scary internet, only allow the white list through.
   def investigation_params
     params.require(:investigation).permit(
-      :title, :description, :image, :risk_level, :is_closed,
+      :question_title, :description, :image, :risk_level, :is_closed,
```
broder: I think the only params that are used in this controller are `description` and `is_closed` (through the "Begin an investigation" and the status page) so we could remove the others - do you want to check it?

web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -1,13 +1,14 @@
 book = xlsx_package.workbook
 
 book.add_worksheet name: "Cases" do |sheet_investigations|
-  sheet_investigations.add_row %w[ID Status Description Assignee Source Risk_Overview Risk_Level
+  sheet_investigations.add_row %w[ID Status Title Description Assignee Source Risk_Overview Risk_Level Sensitivity
```
broder: I'm not sure we need to add `Sensitivity` here.

web/app/views/investigations/_form.html.slim
MatthewBurstein: The line below this inserts a div which even when empty creates some white space between the legend and the label. We could move the div into the _form_validation_display partial but I am hesitant to do this because that partial is used in a lot of places. What do you suggest?

web/app/models/investigation.rb
broder: They methods aren't used in this class though. Helpers are pulled into views automatically so I don't think we need it here - does anything break if you remove it?

(This line isn't on master and things seem to be working. I imagine someone removed it in the commits you rebased onto and this is a rebase conflict.)

MatthewBurstein: Ah I see makes sense. Have removed it.

web/app/views/investigations/_form.html.slim
broder: If there's a reason for the `div` then that's fine - it's just a little different to how we're doing it elsewhere e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/f52e627b761ce75e51874f1dca14c94bc3a442b5/web/app/views/activities/_form.html.slim and https://github.com/UKGovernmentBEIS/beis-mspsds/blob/f52e627b761ce75e51874f1dca14c94bc3a442b5/web/app/views/images_flow/upload.html.slim

PR #232, opened 2018-11-06
MSPSDS-692 new activity links

MatthewBurstein:


web/app/views/investigations/_record_activities.html.slim
```diff
@@ -0,0 +1,28 @@
...
+  | Add new activity
+.govuk-grid-row.mspsds-record-action-panel
+  .govuk-grid-column-two-thirds
+    span.govuk-hint
+      | Record an activity
+    .govuk-grid-row
+      .govuk-grid-column-one-third
+        p
+          = link_to "Record email", '#', class: "mspsds-block-link"
+          br
```
broder: Could we achieve the same effect with a `p` tag for each link?

web/app/assets/stylesheets/block-link.scss
```diff
@@ -1,7 +1,6 @@
 .mspsds-block-link {
   display: inline-block;
-  @include govuk-responsive-margin(2, "top");
-  @include govuk-responsive-margin(2, "bottom");
+  @include govuk-responsive-margin(4, "bottom");
```
broder: This change upsets some alignment on the products and businesses tab:
screenshot 2018-11-06 at 11 09 03

(I imagine that you got this from the design repo but I think we should just use that as a reference and be happy to deviate from it)

PR #240, opened 2018-11-07
MSPSDS-520: Record reported documents

kicferk1:
Several classes are almost empty and only differ by what helpers they include, not sure how happy we are about this vs code duplication that would occur otherwise.


web/app/controllers/concerns/file_concern.rb
```diff
@@ -40,6 +40,12 @@ def update_file_details(file)
     file.metadata["updated"] = Time.current
   end
 
+  def validate_blob_size(blob, errors, allowed_size = max_file_byte_size)
+    if blob && (blob.byte_size > allowed_size)
+      errors.add(:base, :file_too_large, message: "File is too big, allowed size is #{allowed_size / 1000000}MB")
```
broder: Rails defines https://api.rubyonrails.org/classes/Numeric.html#method-i-gigabyte and https://api.rubyonrails.org/classes/Numeric.html#method-i-megabyte which could be useful here?

web/app/controllers/files_controller.rb
```diff
@@ -0,0 +1,61 @@
+class FilesController < ApplicationController
+  before_action :set_parent
+  before_action :set_file, only: %i[edit update destroy]
+
+  # POST /documents
+  def create
+    validate
+    return redirect_to request.referer if @errors.present?
```
broder: Could `request.referer` be undefined? Would you expect us to ever do this redirect?

web/app/views/application/_documents_list_summary.html.slim
```diff
@@ -1,16 +1,23 @@
...
+            = "Updated #{Time.zone.parse(document.blob.metadata[:updated]).strftime('%d/%m/%Y')}"
+        br
+        p
+          = document.blob.metadata[:description]
+        p
+          = link_to "View PDF document", document, class: "govuk-link mspsds-block-link", target: '_blank'
+          = link_to "Edit document details or delete", edit_associated_document_path(parent, document),
+                  class: "govuk-link mspsds-block-link"
 - else
+  h2.govuk-heading-l[class="govuk-!-margin-bottom-1"]
```
broder: Do we want this header if there aren't any documents?

web/app/views/application/_documents_list_summary.html.slim
```diff
@@ -1,16 +1,18 @@
...
-          | Title
-        th.govuk-table__header[scope="col"]
-          | Upload date
-        th.govuk-table__header[scope="col"]
-    tbody.govuk-table__body
-      - parent.documents.attachments.each do |document|
-        = render "document_summary", document: document, parent: parent
-- else
-  p
-    | No documents.
```
jasiekmiko: Is this intentionally removed?

web/app/views/application/_documents_list_summary.html.slim
```diff
@@ -1,16 +1,18 @@
...
+  - parent.documents.attachments.each do |document|
+    .govuk-grid-row[class="govuk-!-padding-bottom-6"]
+      .govuk-grid-column-one-third
+        = render "documents/document_image", document: document
+      .govuk-grid-column-two-thirds
+        h2.govuk-heading-m[class="govuk-!-margin-bottom-1"]
+          = document.blob.metadata[:title]
+        span.govuk-caption-m[class="govuk-!-font-size-16"]
+          - if document.blob.metadata[:updated]
+            = "Updated #{Time.zone.parse(document.blob.metadata[:updated]).strftime('%d/%m/%Y')}"
```
jasiekmiko: Worth moving this into the helper to reduce the logic in view?

web/app/views/application/_images_summary.html.slim
```diff
@@ -7,18 +7,12 @@
         h2.govuk-heading-m[class="govuk-!-margin-bottom-1"]
           = image.blob.metadata[:title]
         span.govuk-caption-m[class="govuk-!-font-size-16"]
-          = "Updated #{image.blob.metadata[:updated]&.to_date}"
+          - if image.blob.metadata[:updated]
+            = "Updated #{Time.zone.parse(image.blob.metadata[:updated]).strftime('%d/%m/%Y')}"
```
jasiekmiko: ...especially that it's repeated! :)

web/app/views/application/_documents_list_summary.html.slim
jasiekmiko: ahh, I see you've removed that after David's comment! :)
I'm just curious if that's a pattern we're applying across the app now I guess!

kicferk1: Nick told me to have a simple paragraph saying that there are no attachments yet, I guess this is a reasonable thing for different places too.

PR #241, opened 2018-11-07
MSPSDS-652 set case priority

MatthewBurstein:


web/app/models/investigation.rb
```diff
@@ -58,6 +60,12 @@ def create_audit_activity_for_case
     ::AuditActivity::Investigation::Add.from(self)
   end
 
+  def create_audit_activity_for_priority
+    if saved_changes.key?(:priority) || saved_changes.key?(:priority_rationale)
+      ::AuditActivity::Investigation::UpdatePriority.from(self)
```
broder: Do we need the `::` at the start?

web/app/models/audit_activity/investigation/update_priority.rb
```diff
@@ -0,0 +1,10 @@
+class AuditActivity::Investigation::UpdatePriority < AuditActivity::Investigation::Base
```
broder: I wonder if in future, we should just have `AuditActivity::Investigation::Update` which could work for both assignment and updating priority.

web/app/views/investigations/priority.html.slim
```diff
@@ -0,0 +1,47 @@
+- content_for :page_title, "Update priority"
+- content_for :after_header do
+  = link_to "Back to case", @investigation, class: "govuk-back-link"
+= render "investigation_heading", investigation: @investigation
+h2.govuk-heading-l
+  | Case status
```
broder: Should this be "Priority"?

web/app/views/investigations/priority.html.slim
```diff
@@ -0,0 +1,47 @@
...
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with(url: update_priority_investigation_url(@investigation), local: true) do |form|
+      - if @investigation.errors.any?
```
broder: Could we use `form_validation_display` here?

web/db/migrate/20181107102901_add_priority_to_investigation.rb
```diff
@@ -0,0 +1,10 @@
+class AddPriorityToInvestigation < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      change_table :investigations, bulk: true do |t|
+        t.string :priority
+        t.string :priority_rationale
```
broder: Does the `priority_rationale` display anywhere other than the activity? Maybe we don't need to actually store it on the investigation... I guess it's either a comment as to why the priority is being updated (in which case would only need to live on the update activity) or it's an essential part of the priority concept (in which case it should stay on the investigation and be displayed there)

web/db/schema.rb
```diff
@@ -142,6 +142,8 @@
     t.text "description"
     t.boolean "is_case", default: true, null: false
     t.boolean "is_closed", default: false
+    t.string "priority"
```
broder: Should we have priority as an enum rather than a string?

MatthewBurstein: Should we be storing this in the database as an enum as shown here?https://dev.to/amplifr/postgres-enums-with-rails-4ld0

broder: The "Rails way" (i.e. `enum status: [ :active, :archived ]`), yes. There's an example in our codebase at https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/app/models/hazard.rb

web/app/models/audit_activity/investigation/update_priority.rb
broder: There will be soon! I think it would make sense to do this as a part of this ticket - they're quite close in terms of functionality.

MatthewBurstein: Have added this as part of this ticket

web/app/views/investigations/priority.html.slim
```diff
@@ -12,16 +12,16 @@
         .govuk-form-group
           .govuk-radios
             .govuk-radios__item
```
broder: Could we do something like:
```
- Investigation.priorities.each do |(priority, _)|
      .govuk-radios__item
        = form.radio_button :priority, priority, class: "govuk-radios__input"
        = form.label :priority, priority.titleize, class: "govuk-label govuk-radios__label"
```
?

web/app/models/investigation.rb
```diff
@@ -58,8 +58,23 @@ def pretty_id
...
+  def question_title_prefix
+    question_type && !is_case ? question_type + ' ' : ''
+  end
+
+  def title
+    self.is_case ? case_title : question_title
+  end
+
+  def display_priority
+    priority&.capitalize
```
broder: I'd almost be tempted to just use `investigation.priority&.titleize` wherever we're using this - I think it would be more readable.

PR #243, opened 2018-11-07
MSPSDS-559: Display link to document on activity timeline

raamSoftwire:


web/app/views/application/_activities_summary.html.slim
```diff
@@ -21,6 +21,9 @@
                   = markdown activity.body if activity.body
                   = link_to "View business details", business_url(activity.business_id) if activity.business_id
                   = link_to "View product details", product_url(activity.product_id) if activity.product_id
+                  - if activity.is_a?(AuditActivity::Document::Base)
```
broder: If we're going to end up with a lot of these `is_a?` checks for documents, maybe we could use a concern to call it `attachment` everywhere? Then the conditional could be `if activity.respond_to?(:attachment) && activity.attachment.attached?`.

We'd be able to customise the link text using some kind of `document_type` method which could be overridden in the various activity classes.

PR #244, opened 2018-11-08
MSPSDS-552 record case comments

MatthewBurstein:


web/app/models/comment_activity.rb
```diff
@@ -1,7 +1,11 @@
...
   validates :body, presence: true
 
-  def subtitle_slug
-    "Comment added"
+  def title
+    "Comment: #{source.show.titleize}"
+  end
+
+  def subtitle
+    pretty_date_stamp
```
broder: It feels a little weird that the title / subtitle are differently formatted for comments compared to every other activity...

web/app/assets/javascripts/investigations.js
```diff
@@ -2,3 +2,15 @@
 $(document).on('turbolinks:load', function () {
   simpleAccessibleAutocomplete('assignee-picker');
 });
+
+$(document).ready(function () {
```
broder: You'll want to do this on `turbolinks:load`. I can explain what exactly Turbolinks is if you're not sure

web/test/fixtures/sources.yml
```diff
@@ -51,3 +51,13 @@ address_one:
 corresopndence_one:
   type:
   sourceable: one (Correspondence)
+
+activity_one:
+  user_id: e54a3d29-5714-4232-b66a-1e0c0c6fdebc
```
broder: Do you need to define `user_id` here? I think the `UserSource` class is able to deal with nil `user_id`s.

MatthewBurstein: If we don't include `user_id` here then the correspondence flow test breaks. This happens because in the view we call activity.title which calls source.show.title. Without the user_id source.show is nil so source.show.title throws an error. Is there a better solution?

web/app/models/comment_activity.rb
MatthewBurstein: Agreed, but not sure what the solution is? Should I go back to the designers?

web/test/fixtures/sources.yml
broder: I think `source.show` should be returning "anonymous": https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/app/models/user_source.rb#L5

PR #246, opened 2018-11-08
MSPSDS-472: Improve technical setup

jasiekmiko:
The code changes comprise:
- De-verbalising logs via [Lograge](https://github.com/roidrage/lograge)
- Adding timeouts [Slowpoke](https://github.com/ankane/slowpoke)
- Adding db admin dashboard [pgHero](https://github.com/ankane/pghero) 
- NOT adding the [secure_headers gem](https://github.com/twitter/secure_headers) as this is already covered by rails (see https://github.com/twitter/secure_headers/issues/394)
- Preventing host header injection (following [this piece of advice](https://github.com/ankane/secure_rails)


web/config/routes.rb
```diff
@@ -83,6 +83,10 @@
 
   match "/404", to: "errors#not_found", via: :all
   match "/500", to: "errors#internal_server_error", via: :all
+  # This is the page that will show for timeouts, currently showing the same as an internal error
+  match "/503", to: "errors#internal_server_error", via: :all
```
broder: Should we make a new controller method for this? `errors#internal_server_error` returns 500 rather than 503. We can probably reuse the `internal_server_error` view however!

web/config/routes.rb
```diff
@@ -83,6 +83,10 @@
...
+
+  mount PgHero::Engine, at: "pghero"
```
broder: Do we want this in production? Also, do we need to be running https://github.com/ankane/pghero/blob/master/guides/Rails.md#historical-query-stats?

jasiekmiko:
Damn it, I've written a whole host of comments before putting into review that I've accidentally deleted now :P

web/config/database.yml
```diff
@@ -21,6 +21,9 @@ default: &default
   username: postgres
   password:
   pool: 5
+  connect_timeout: 1
```
jasiekmiko: @broder any reason we wouldn't want these in dev?

web/config/routes.rb
jasiekmiko: I think we very much do - otherwise we can't properly evaluate the performance.
And that was one of my comments, do 
https://github.com/ankane/pghero/blob/master/guides/Rails.md#historical-query-stats and
https://github.com/ankane/pghero/blob/master/guides/Rails.md#historical-space-stats
seem useful here?
I'm tempted to say yes, but also not sure what we'd want to use for the stats capture scheduling. I've had quick look at https://github.com/ondrejbartas/sidekiq-cron, am i barking up the right tree here?

jasiekmiko: Finally, there's a recommendation to use a separate user for phHero, but the required approach seems super messy:
https://github.com/ankane/pghero/blob/master/guides/Permissions.md

broder: sidekiq-cron sounds like it might work!

web/config/database.yml
broder: Should be fine in dev (and it'll be fairly easy to change it if it isn't). https://www.postgresql.org/docs/9.1/libpq-connect.html#LIBPQ-CONNECT-CONNECT-TIMEOUT says that it shouldn't be less than 2 seconds though

jasiekmiko: ah interesting - clashes with the recommendation from the other place. I'll suggest a change to them. Thanks for pointin gout!

jasiekmiko:
@broder as discussed, I'll want to push this through and leave enabling pghero historical stats to a separate PR

PR #247, opened 2018-11-08
MSPSDS-472: Enable Logit logging

broder:


docker-compose.yml
```diff
@@ -45,6 +45,7 @@ services:
     depends_on:
       - db
       - elasticsearch
+      - keycloak
```
jasiekmiko: huh? how come it was missing (ie. how come it was working despite missing?)

broder: It worked if you did `docker-compose up` (which starts everything). If you just did `docker-compose up web` (which just starts web and its dependencies) it wouldn't work.

jasiekmiko: ahhh

PR #251, opened 2018-11-09
MSPSDS-472: Pen test fixes

broder:
I'd probably review this commit-by-commit.


web/app/controllers/investigations/activities_controller.rb
```diff
@@ -1,30 +1,22 @@
...
-    @activities = @investigation.activities.paginate(page: params[:page], per_page: 20)
-  end
+class Investigations::ActivitiesController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
 
-  # GET /activities/new
-  def new
-    @activity = CommentActivity.new
-  end
+  before_action :set_investigation, only: %i[create]
```
jasiekmiko: do we need the `only`s here with `create` being the only action?

PR #252, opened 2018-11-09
MSPSDS-472 Enable collecting query and space statistics for pgHero

jasiekmiko:


postgres/Dockerfile
```diff
@@ -3,3 +3,9 @@ FROM postgres
 WORKDIR /docker-entrypoint-initdb.d
 
 COPY setup-keycloak.sh .
+
+# These allow pgHero to use pt_stat_statements in dev mode
+RUN echo "shared_preload_libraries = 'pg_stat_statements'" >> /var/lib/postgresql/data/postgresql.conf
```
broder: We could put this into a `setup-pghero.sh` script but it's not very important.

web/config/initializers/sidekiq.rb
```diff
@@ -1,5 +1,15 @@
...
   config.redis = Rails.application.config_for(:redis)
+  pg_hero_stats_capture_job = Sidekiq::Cron::Job.new(
+    name: 'pgHero stats capture - every 5min',
+    cron: '*/5 * * * *',
+    class: "DbStatsCaptureJob"
+  )
+
+  unless pg_hero_stats_capture_job.save
+    puts "***** WARNING - pgHero stats capture job was not saved! *****"
+    puts pg_hero_stats_capture_job.errors
```
broder: Should we log this with the Rails logger?

web/db/migrate/20181108172853_create_pghero_query_stats.rb
```diff
@@ -0,0 +1,15 @@
+class CreatePgheroQueryStats < ActiveRecord::Migration[5.2]
```
broder: We could merge these migrations but again, it's not too important

web/config/initializers/sidekiq.rb
jasiekmiko: oopsy! Yeah :)

web/db/migrate/20181108172853_create_pghero_query_stats.rb
jasiekmiko: hmm, yeah, could be - it doesn't impact anything though right? They came from the `gem` so keen to keep them as is without touching

broder: More migrations slow things down but yeah, it's fine to leave them

postgres/Dockerfile
jasiekmiko: seems neat especially as it removes the need for a comment and keeps it as a single RUN

postgres/Dockerfile
```diff
@@ -4,8 +4,5 @@ WORKDIR /docker-entrypoint-initdb.d
 
 COPY setup-keycloak.sh .
 
-# These allow pgHero to use pt_stat_statements in dev mode
-RUN echo "shared_preload_libraries = 'pg_stat_statements'" >> /var/lib/postgresql/data/postgresql.conf
-RUN echo "pg_stat_statements.track = all" >> /var/lib/postgresql/data/postgresql.conf
-RUN echo "pg_stat_statements.max = 10000" >> /var/lib/postgresql/data/postgresql.conf
-RUN echo "track_activity_query_size = 2048" >> /var/lib/postgresql/data/postgresql.conf
+# Adds configuration to the postgres server to allow pgHero to see query stats
+RUN setup=pgHero.sh
```
broder: It should be `setup-pgHero.sh`

jasiekmiko: woops xD

broder:
What's the plan for PaaS?

jasiekmiko:
There's not one yet, we wanted to push this out to see what, if anything is missing from the postgres setup before investigating how to change it.

PR #255, opened 2018-11-12
MSPSDS-545: Assign case to individual

kicferk1:
Need to write some tests because some behaviour is annoying to check by hand


web/app/models/investigation.rb
```diff
@@ -43,6 +43,8 @@ class Investigation < ApplicationRecord
   has_one :reporter, dependent: :destroy
   has_one :hazard, dependent: :destroy
 
+  before_create :assign_current_user_to_case
```
tom-bell-softwire: Will this cause issues for investigations that are imported from RAPEX (since there won't be a current user when the import script is run)?

web/app/assets/stylesheets/form.scss
```diff
@@ -5,3 +5,7 @@
 .govuk-button + .govuk-button {
   margin-left: 24px;
 }
+
+.autocomplete__option {
```
broder: This is already fixed by `autocomplete.scss`.

web/app/models/investigation.rb
broder: Yeah, maybe we should add a check to use `UserSource.new(user: current_user)` only if `source` is not already defined?

web/app/models/investigation.rb
```diff
@@ -93,6 +106,13 @@ def create_audit_activity_for_business business
     AuditActivity::Business::Add.from(business, self)
   end
 
+  def assign_current_user_to_case
+    return if current_user.blank?
+
+    self.source = UserSource.new(user: current_user) if self.source.blank?
+    self.assignee = current_user
```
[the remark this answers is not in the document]
kicferk1: Hey, I asked the designers on slack before I started working on it, and I got "Cases should never be unassigned – they should start off assigned to the person who created".

PR #256, opened 2018-11-12
Increase postgres statement timeout

raamSoftwire:


web/config/database.yml
```diff
@@ -23,7 +23,7 @@ default: &default
   pool: 5
   connect_timeout: 2
   variables:
-    statement_timeout: 250
+    statement_timeout: 500
```
broder: Should we do this for development only? In production we use `MIGRATION_STATEMENT_TIMEOUT`.

... in fact, that should be happening in dev too - did you down and up after moving onto master? That might also fix this issue...

jasiekmiko: Oh, I didn't think of the need to check down-and-up-ing!
My guess was that `MIGRATION_STATEMENT_TIMEOUT` might only get applied to migrations, but not to the db creation (which is the statement that I saw fail).

As for changing the value in defaults rather then just for test, I thought there's no particular tie to this specific value in production? Or do we want to keep it as small as possible?

PR #261, opened 2018-11-14
Mspsds 716 missing activity flow records

raamSoftwire:


web/test/models/investigation_test.rb
```diff
@@ -11,12 +11,63 @@ class InvestigationTest < ActiveSupport::TestCase
...
+    assert_difference "Activity.count" do
+      @investigation.is_closed = !@investigation.is_closed
       @investigation.save
     end
   end
 
-  def add_investigation
-    @investigation = Investigation.new
+  def create_investigation
+    @investigation = Investigation.create
```
broder: We could probably just inline these `.create`s - saves someone from having to scan up and down the file to figure out what's going on

web/test/models/investigation_test.rb
```diff
@@ -11,12 +11,63 @@ class InvestigationTest < ActiveSupport::TestCase
...
+  end
+
+  def create_business
+    @business = Business.new
+    @business.company_name = 'TestCompany'
```
broder: You could do this in one line using `Business.new(company_name: 'Test Company')`

PR #262, opened 2018-11-14
MSPSDS-537: Filter by status

kicferk1:
Includes UI for all filters, but only status is doing anything


web/app/models/user.rb
```diff
@@ -46,6 +46,15 @@ def self.get_assignees_select_options(except_those_users = [])
     select_options
   end
 
+  def self.get_assignees_select_options_short
```
tom-bell-softwire: It feels a bit inconsistent to use a shorter display name here. What if there are multiple people with the same name?

Since the filter is for a specific person (rather than all cases where the assignee matches the selected name), I think we need to disambiguate by showing the email address. Perhaps best to confirm this with the designers.

web/app/views/investigations/index.html.slim
```diff
@@ -2,6 +2,80 @@
...
+              .govuk-checkboxes
+                .govuk-checkboxes__item
+                  = form.check_box :assigned_to, { multiple: true, class: "govuk-checkboxes__input" }, "me", true
+                  = form.label :assigned_to, "Me", class: "govuk-checkboxes__label"
+                .govuk-checkboxes__item
+                  = form.check_box :assigned_to, { multiple: true, class: "govuk-checkboxes__input",
+                          "data-aria-controls": "conditional-Assignee-2" }, "other", true
+                  = form.label :assigned_to, "Someone else", class: "govuk-checkboxes__label"
+                #conditional-Assignee-2.govuk-checkboxes__conditional.govuk-checkboxes__conditional--hidden
+                  = form.label :assigned_to, "Name / Team", class: "govuk-label"
```
tom-bell-softwire: I think we decided to remove the "Team", since we don't support teams yet. Again, best to confirm with Tim and Ed.

web/app/views/investigations/index.html.slim
```diff
@@ -2,6 +2,80 @@
...
+                  = form.text_field :assigned_to, class: "govuk-input"
+                  = form.select :assigned_to,
+                          options_for_select(User.get_assignees_select_options_short),
```
tom-bell-softwire: Should we exclude the current user from the options?

web/app/views/investigations/index.html.slim
```diff
@@ -2,6 +2,80 @@
...
+                | Assigned to
+              .govuk-checkboxes
+                .govuk-checkboxes__item
+                  = form.check_box :assigned_to, { multiple: true, class: "govuk-checkboxes__input" }, "me", true
+                  = form.label :assigned_to, "Me", class: "govuk-checkboxes__label"
+                .govuk-checkboxes__item
+                  = form.check_box :assigned_to, { multiple: true, class: "govuk-checkboxes__input",
+                          "data-aria-controls": "conditional-Assignee-2" }, "other", true
+                  = form.label :assigned_to, "Someone else", class: "govuk-checkboxes__label"
+                #conditional-Assignee-2.govuk-checkboxes__conditional.govuk-checkboxes__conditional--hidden
```
tom-bell-softwire: Does this get shown/hidden automatically by the Design System JS? Neat!

tom-bell-softwire:
Does this also cover part of MSPSDS-722 as well?

web/app/helpers/investigations_helper.rb
```diff
@@ -16,10 +16,25 @@ def sort_direction
...
+    { status: 'closed' }
   end
 
   def query_params
-    params.permit(:q, :sort, :direction, :status)
+    handle_status_params
+    params.permit(:q, :sort, :direction, :status_open, :status_closed)
+  end
+
+  def handle_status_params
```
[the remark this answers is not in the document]
broder: Do we need this if we're checking for `blank?` in `get_status_filter`?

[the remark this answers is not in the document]
kicferk1: I think there are two flows of information that are a bit disjoint here. 
filter_params is used when we create a query for elasticsearch, whereas query_params is used for creating a model for our form. 

This check makes sure the 'default' look of the page is 'open filter checked', and get_status_filter one makes sure the query is correct for the default case.

web/app/models/user.rb
kicferk1: The field is too short to ever show both the name and email, but I can definitely ask designers about it.

kicferk1:
Yeah, I decided that there is little point styling just the tiny part that I need so I did the styling for the entire searchbox component.

web/app/models/user.rb
kicferk1: "For now just display user's name"

PR #264, opened 2018-11-14
MSPSDS-498 record case comms email

MatthewBurstein:
Not ready for merging yet but would be interested to get feedback. The bulk of the functionality is in place but there is definitely a lot of room for improvement around adding multiple attachments, especially in FileConcern, EmailsController and AuditActivity::Correspondence::AddEmail and there are some fiddly issues to fix also.

Should we have a separate activity display partial for each each kind of activity in the activities summary? Should larger refactoring like this be a separate PR?


web/app/assets/javascripts/investigations/correspondence.js
```diff
@@ -0,0 +1,12 @@
+$(document).on('turbolinks:load', function () {
+  $('#correspondence_email_attachment').on('change', function () {
+    var attachementDescription = $('#attachement-description');
```
tom-bell-softwire: This bit of functionality (attachment description field shown only after selecting a file) appears on a number of activity pages, so I'm planning to pull it out into a partial and associated JS file.

Depending on when that gets merged in, we can then remove this.

web/app/controllers/investigations/emails_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    initialize_file_attachment :email_file
+    initialize_file_attachment :email_attachment
+    redirect_to wizard_path(steps.first, request.query_parameters)
+  end
+
+  def create
+    attach_file_to_list(@email_file, :email_file, @correspondence.documents)
+    attach_file_to_list(@email_attachment, :email_attachment, @correspondence.documents)
+    @investigation.correspondences << @correspondence
+    @investigation.save
```
tom-bell-softwire: Do we need a check here to handle a failed `save` due to validation errors? Perhaps unlikely when following the wizard steps, but a direct POST request (i.e. API call) could contain invalid properties.

web/app/controllers/investigations/emails_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    AuditActivity::Correspondence::AddEmail.from(@correspondence, @investigation)
+    redirect_to investigation_path(@investigation)
```
tom-bell-softwire: I think the convention elsewhere is to include a flash success notice when redirecting to the case page. Is that explicitly _not_ wanted here?

web/db/migrate/20181109145423_add_email_subject_direction_and_attachement_description_to_correspondence.rb
```diff
@@ -0,0 +1,11 @@
+class AddEmailSubjectDirectionAndAttachementDescriptionToCorrespondence < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      change_table :correspondences do |t|
+        t.string "email_direction"
+        t.text "attachment_description"
```
tom-bell-softwire: Should we be storing the attachment description as a field on the correspondence? I think it should just be set in the metadata of the relevant attachment blob.

web/app/assets/javascripts/investigations/correspondence.js
MatthewBurstein: Should I wait for the merge to see what you have done before working on this?

tom-bell-softwire: Unless you're planning a major refactor of this, no need to wait for the other branch to be merged! Should be easy to refactor this afterwards instead.

web/db/schema.rb
```diff
@@ -10,10 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_11_08_172853) do
+ActiveRecord::Schema.define(version: 2018_11_09_145423) do
```
MatthewBurstein: The version number causes a merge conflict with master. What's the best way to resolve this?

broder: Delete the file and run `db:migrate` again locally

web/app/assets/javascripts/investigations/correspondence.js
```diff
@@ -0,0 +1,12 @@
+$(document).on('turbolinks:load', function () {
```
broder: Do we need this as well as `attachment-description.js`?

web/db/schema.rb
```diff
@@ -164,6 +166,26 @@
     t.index ["assignee_id"], name: "index_investigations_on_assignee_id"
   end
 
+  create_table "pghero_query_stats", force: :cascade do |t|
```
broder: We removed these recently - they shouldn't be in `schema.rb` anymore

MatthewBurstein: I deleted the schema and migrated again but they are still there. How should I get rid?

web/app/assets/javascripts/investigations/correspondence.js
MatthewBurstein: Yes meant to do this. Have deleted

PR #267, opened 2018-11-15
MSPSDS-553: Record test activity on cases

tom-bell-softwire:


web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+  steps :details, :confirmation
+
+  before_action :set_investigation
+  before_action :build_test_from_params, only: %i[update]
+  before_action :store_test, only: %i[update]
+  before_action :restore_test, only: %i[show create]
+  before_action :set_attachment, only: %i[show update create]
+
+  # GET /tests/1
+  # GET /tests/1.json
```
broder: Does this actually respond to a JSON request?

web/app/views/investigations/tests/details.html.slim
```diff
@@ -0,0 +1,75 @@
...
+                  class: "govuk-select", id: "test_product"
+
+        legend.govuk-fieldset__legend.govuk-fieldset__legend--l
+          | Against which legislation?
+        .govuk-form-group class=("govuk-form-group--error" if @test.errors[:legislation].any?)
+          span.govuk-hint
+            | Select the relevant legislation from the list, or enter a different one.
+          #legislation-autocomplete-container[data-options=legislation_constants["legislation"]
+                                              data-value=@test.legislation]
+            = form.text_field :legislation, class: "govuk-input"
```
broder: What was the business reason behind allowing them to enter any legislation they like?

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+  def new_result
+    clear_session
+    session[:test] = { type: Test::Result.name }
+    redirect_to wizard_path(steps.first, request.query_parameters)
+  end
+
+  # POST /tests
+  # POST /tests.json
+  def create
+    attach_files
```
broder: Should we do this after the `.save` call?

web/app/helpers/activity_helper.rb
```diff
@@ -0,0 +1,11 @@
+module ActivityHelper
+  def activity_attachment_link_text(activity)
```
broder: I think I preferred your idea for putting an `attachments` method on the base activity and then overriding it in the activities that have attachments. The method could return a hash of the activity field name and a pretty name to display it with (i.e. `{"risk assessment": :risk_assessment}` or something like that) - what do you think?

I've also mentioned this to Matt here https://github.com/UKGovernmentBEIS/beis-mspsds/pull/264/files#r233804162 but would be interested to hear your thoughts.

web/app/models/audit_activity/test/base.rb
```diff
@@ -0,0 +1,28 @@
...
+      source: UserSource.new(user: current_user),
+      investigation: investigation,
+      product: test.product,
+      test: test
+    )
+    attach_to_activity(activity, test.documents.first) if test.documents.attached?
+  end
+
+  def self.build_body(test)
+    date_label = test.requested? ? "Date requested" : "Test date"
```
broder: It's maybe a little bit confusing to have something that refers to either a request or a result in the base activity. It might be a bit of a faff to move the code into either `AuditActivity::Test::Request` or `AuditActivity::Test::Result` though...

web/app/models/test.rb
```diff
@@ -0,0 +1,33 @@
...
+  include DateConcern
+
+  belongs_to :investigation
+  belongs_to :product
+
+  enum result: { passed: "Pass", failed: "Fail" }
+
+  has_many_attached :documents
+
+  attribute :day, :integer
```
broder: I wonder should we move these fields into `DateConcern`. @jasiekmiko was there any particular reason to not put it in the concern?

web/app/models/test.rb
```diff
@@ -0,0 +1,33 @@
...
+  attribute :month, :integer
+  attribute :year, :integer
+  validate :date_from_components
+
+  validates :investigation, presence: true
+  validates :legislation, presence: true
+  validates :product, presence: true
+
+  def initialize(*args)
+    raise "Cannot directly instantiate a Test record" if self.class == Test
```
broder: Does `self.abstract_class = true` work instead of this? We should maybe put that on a lot of the activity classes too...

web/app/models/test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    super
+  end
+
+  def pretty_name
+    "test record"
+  end
+
+  def requested?
+    self.class == Test::Request
```
broder: Maybe we should set this in the child classes and raise an error here instead?

web/test/controllers/investigations/tests_controller_test.rb
```diff
@@ -0,0 +1,70 @@
...
+      }
+    end
+
+    assert_redirected_to investigation_url(@investigation)
+  end
+
+  test "should add test record to investigation" do
+    post investigation_tests_path(@investigation), params: {
+        test: {
+            type: "Test::Request",
```
broder: What would happen if someone POSTed something that wasn't a type. Or if they posted something like `ActiveRecord::Base`?

web/app/controllers/investigations/tests_controller.rb
tom-bell-softwire: Oops. Nope.

web/app/controllers/investigations/tests_controller.rb
tom-bell-softwire: I'll check, but I seem to remember there being some issue about the attachment associations not being persisted properly. Perhaps an additional `save` call after attaching the files would solve that.

web/app/helpers/activity_helper.rb
tom-bell-softwire: Good idea!

web/app/models/test.rb
tom-bell-softwire: Yeah, that would be nicer.

web/app/models/test.rb
tom-bell-softwire: Unfortunately, that seems to disable the Single Table Inheritance functionality: https://github.com/rails/rails/issues/23783
https://stackoverflow.com/questions/2850418/rails-sti-prevent-base-class-from-instantiation

web/app/models/test.rb
tom-bell-softwire: Good call.

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+      redirect_to investigation_url(@investigation), notice: "#{@test.pretty_name.capitalize} was successfully recorded."
+    else
+      render step
+    end
+  end
+
+  # PATCH/PUT /tests/1
+  # PATCH/PUT /tests/1.json
+  def update
+    if @test.valid?(step)
```
MatthewBurstein: should we be validating the size of the attachment here?

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,118 @@
...
+    else
+      render step
+    end
+  end
+
+private
+
+  def clear_session
+    session[:test] = nil
+    initialize_file_attachment
```
MatthewBurstein: If you override get_file_session_key in this controller then I don't think you need to call `session[:test] = nil` as well as `initialize_file_attachment`

web/app/controllers/investigations/tests_controller.rb
tom-bell-softwire: Ah yes, good catch!

web/test/controllers/investigations/tests_controller_test.rb
tom-bell-softwire: It returns an error. I've added a test case to cover that scenario.

web/app/controllers/investigations/tests_controller.rb
tom-bell-softwire: I think they are separate keys in the session: one for the `Test` model properties and one for the file `Blob`.

web/test/controllers/investigations/tests_controller_test.rb
broder: Would it be better to have a bool parameter like `result`? That way we aren't really revealing anything about how it works under the hood.

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -43,7 +42,7 @@ def create
   # PATCH/PUT /tests/1
   # PATCH/PUT /tests/1.json
```
broder: There are some other `json` comments

web/app/models/concerns/activity_attachable.rb
```diff
@@ -1,13 +1,44 @@
...
+        self.class_eval do
+          has_one_attached key
+        end
+      end
+    end
+  end
+
+  def attachment_names
+    klass = self.class
+    while klass.respond_to? :attachment_names do
```
broder: Is this for when we define `with_attachments` on a parent class and call `attachments` from the child or if we have `with_attachments` on both parent and child?

web/test/controllers/investigations/tests_controller_test.rb
tom-bell-softwire: I've added an `is_result` boolean param instead of `type` (`result` is already used to specify the test result).

web/app/models/concerns/activity_attachable.rb
tom-bell-softwire: The former -- to allow child classes to access the `attachment_names` property on the parent.
Declaring `with_attachments` on the child will override the parent value.

web/app/models/test/result.rb
```diff
@@ -0,0 +1,19 @@
+class Test::Result < Test
```
broder: Will this correctly inherit from the `Test` class rather than the `Test` module? We had issues with activities which lead to us needing to do things like `AuditActivity::Base`.

web/app/views/application/_form_validation_display.html.slim
```diff
@@ -1,8 +1,8 @@
+- model_name = "case" if local_assigns[:model_name].blank?
```
broder: Could we do something like "prohibited this item from being saved"?

web/test/models/test_test.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  test "test result requires result to be specified" do
+    test_request = Test::Result.create(investigation: @investigation, product: @product, date: "2018-11-08")
+    assert_not test_request.save
+    test_request.result = "Fail"
+    assert test_request.save
+  end
+
+  test "should create activity when test request is created" do
+    assert_difference"Activity.count" do
```
broder: Do we need a space between `assert_difference` and `"Activity.count"`?

web/db/schema.rb
```diff
@@ -192,13 +194,30 @@
...
   end
 
+  create_table "tests", id: :serial, force: :cascade do |t|
+    t.datetime "created_at", null: false
+    t.date "date"
+    t.text "details"
+    t.integer "investigation_id"
+    t.string "legislation"
+    t.integer "product_id"
+    t.string "result"
```
broder: Should we store this as an int rather than a string? Will we ever want values other than those in the enum?

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  # GET /tests/1
+  def show
+    render_wizard
+  end
+
+  # GET /tests/new_request
+  def new_request
+    clear_session
+    session[:test] = { type: Test::Request.name }
```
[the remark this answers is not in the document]
tom-bell-softwire: The session object is just a copy of the test attributes, which includes the appropriate `type` when copied from a model instance. Adding the `is_result` property to the session makes things a bit more fiddly, so I'd prefer to leave it as is.

web/app/controllers/investigations/tests_controller.rb
```diff
@@ -0,0 +1,126 @@
...
+                :result,
+                :details,
+                :day,
+                :month,
+                :year)
+        .merge(type: model_type)
+  end
+
+  def model_type
+    return nil if params[:test][:is_result].blank?
```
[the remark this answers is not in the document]
tom-bell-softwire: Good call.

web/app/models/audit_activity/test/base.rb
```diff
@@ -0,0 +1,30 @@
+class AuditActivity::Test::Base < AuditActivity::Base
+  include ActivityAttachable
+  with_attachments attachment: "attachment"
+
+  belongs_to :test
```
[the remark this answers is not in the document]
tom-bell-softwire: Good point. I'll remove the association and relevant field.

web/app/models/test/result.rb
tom-bell-softwire: Seems to work as expected in this case!

web/app/views/application/_form_validation_display.html.slim
tom-bell-softwire: You mean to make "item" the fallback text? Sure, that seems more generally applicable.

web/app/models/test/result.rb
broder: Yeah, it only broke in production and in some test cases...

web/test/models/test_test.rb
tom-bell-softwire: Oh yes, good catch!

web/app/controllers/investigations/tests_controller.rb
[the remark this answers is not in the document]
broder: Ah, yes, sorry

PR #274, opened 2018-11-16
MSPSDS-729: Fix changing status on case

raamSoftwire:


raamSoftwire:
Is this a case where we would ignore coveralls?

PR #275, opened 2018-11-16
MSPSDS-472 Disable pgHero stats collection

jasiekmiko:
Currently, there's no way to enable the correct config in PaaS and
keeping it just on dev is both problematic and limited in use.

Further progress to be done as part of MSPSDS-728


broder:
Should we remove all the pghero related stuff? Then we can just say to revert this commit on MSPSDS-728.

jasiekmiko:
Don't think removing everything is the right choice - we still get some info without the historical data.
Potentially we could remove the historical data tables, but it feels like more faff than it's worth it with migrations back and forth if we do end up enabling it.

I'll make a note on the ticket that if it turns out impossible, they should be removed?

PR #282, opened 2018-11-19
MSPSDS-537: Filter by status

kicferk1:
Need to write tests for filtering and for xml


web/app/models/concerns/searchable.rb
```diff
@@ -63,6 +63,7 @@ module Searchable
     end
 
     def self.full_search(query)
+      __elasticsearch__.refresh_index! if Rails.env.development? || Rails.env.test?
```
broder: What's happening, is when you make a change in something like `investigation.rb`, the file is reloaded and https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/app/models/investigation.rb#L173 is triggered, causing the index to be deleted and recreated. The line you've added forces us to wait for the index to be created before calling `search`.

I think we'll only need this for dev. I'd argue that we don't really need this (as we can just reload the page) but if you feel strongly about keeping it, then we should add a comment explaining why we need this and why it's only needed in dev...

(Actually, do we need this in test???)

kicferk1: I don't think we need it in test. 

I think eventually we might get rid of it in dev too(or be more sophisticated about it), but it ought to be useful for this sprint and maybe next one, since we will be changing and debugging search/sort/filter mechanics extensively and having to remember about refreshing a page each time something goes wrong is pretty annoying.

Even more annoying is that update of a record seems not to trigger a recreation of index either, so when you change a status it might take a few seconds for it to show up in your search results. It was a pain this morning :(

kicferk1: Actually, now that I added tests for filters, we probably do need it in tests :( Otherwise some tests on Travis build will fail.

PR #284, opened 2018-11-19
MSPSDS-538: Search cases by keyword

kicferk1:
Still a bit of a WIP. 
TODOs: 
More tests
Figure out how to deal with file metadata
Figure out how to deal with activity search


web/test/models/investigation_test.rb
```diff
@@ -56,4 +56,57 @@ class InvestigationTest < ActiveSupport::TestCase
       @investigation.save
     end
   end
+
+  test "elasticsearch should find exactly batch_number, brand, description, gtin, model, name of a product" do
```
broder: There are some arguments that each test should only have one `assert` statement - should we split this into multiple tests and inline the `search_for_product_qq` methods?

web/app/models/investigation.rb
```diff
@@ -51,7 +51,42 @@ class Investigation < ApplicationRecord
...
+          only: :description
+        },
+        products: {
+          only: %i[batch_number brand description gtin model name]
+        },
+        reporter: {
+          only: %i[name phone_number email_address other_details]
+        }
+      }
+    ).merge(status: status.downcase)
```
broder: We should probably move this into the above block.

I also don't think that we need to send `status` - we're already sending `is_closed` which is true or false

PR #285, opened 2018-11-20
Release to staging

broder:


broder:
I've already rolled it back

PR #292, opened 2018-11-21
MSPSDS-553: Fix handling of attachment metadata when recording test activity

tom-bell-softwire:
This fixes an issue where the file metadata wasn't being updated when returning to edit the details from the confirmation page.

Unused routes for test resources have also been removed.


web/app/controllers/investigations/tests_controller.rb
```diff
@@ -71,6 +71,7 @@ def store_test
 
   def test_valid?
     @test.validate(step)
+    update_file_details(@file)
```
kicferk1: I'm not sure update_file_details belongs in validation.

I wonder if it would be better to call it in load_file_attachment in FileConcern, so that whenever a file is loaded by id it also has its details updated. I can imagine the issue with updating metadata isn't unique to tests, and it seems like FileConcern should handle it in general.

broder: @MatthewBurstein and I also have been thinking about where the update should go this afternoon. We should probably do something similar to how rails-y models work (we could even wrap the attachment in an `ActiveModel`...). Maybe we update in a `before_action` for update / create and then validate in the update / create actions themselves?

tom-bell-softwire: Thanks. I think it's best to wait until #264 is merged in before updating this, since that includes a number of wider changes that should handle metadata updates more consistently in controllers.

tom-bell-softwire: Most of the changes in this PR have already been covered by #264, leaving only the smaller updates here.

PR #298, opened 2018-11-22
MSPSDS-644: View search result

kicferk1:


web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -6,4 +6,30 @@ def case_question_text(investigation)
...
+      }
+    end
+  end
+
+  def get_highlight_title(highlight)
+    field_name = highlight[0]
+    replace_unsightly_field_names(field_name).gsub('.', ', ').humanize
+  end
+
+  def replace_unsightly_field_names(field_name)
```
broder: Should we do this with a hash? Maybe `pretty_field_names = { pretty_id: "Case ID", search_index: "Comment" }`. If the field name doesn't exist in the hash, just keep it as is

PR #300, opened 2018-11-22
MSPSDS-536: Sort investigations

aidayounesi:


web/app/helpers/investigations_helper.rb
```diff
@@ -30,10 +41,15 @@ def get_status_filter
 
   def query_params
     set_default_status_filter
-    params.permit(:q, :sort, :direction, :status_open, :status_closed, :page)
+    set_default_sort_by_filter
+    params.permit(:q, :sort, :direction, :status_open, :status_closed, :page, :sort_by)
```
kicferk1: Do we need both :sort and :sort_by? 
Similarly do we need direction?

web/app/helpers/search_params.rb
```diff
@@ -2,7 +2,7 @@
 class SearchParams
   include ActiveModel::Model
 
-  attr_accessor :q, :sort, :direction, :status_open, :status_closed
+  attr_accessor :q, :sort, :direction, :status_open, :status_closed, :sort_by
```
kicferk1: Like above, sort and direction are probably obsolete

web/app/views/investigations/index.html.slim
```diff
@@ -54,14 +54,14 @@ main#main-content.govuk-main-wrapper role="main"
                 | Sort by
               .govuk-radios class="govuk-!-margin-bottom-5"
                 .govuk-radios__item
-                  = form.radio_button :sort_by, "newest", class: "govuk-radios__input"
-                  = form.label :sort_by, "Newest", class: "govuk-radios__label"
-                .govuk-radios__item
-                  = form.radio_button :sort_by, "latest", class: "govuk-radios__input"
-                  = form.label :sort_by, "Updated – recent", class: "govuk-radios__label"
+                  = form.radio_button :sort_by, "recent", class: "govuk-radios__input"
```
kicferk1: As above, it seems :sort and :direction was used to handle sorting before, and the hidden fields are there to maintain their logic in the model. But now that :sort_by is used we probably don't need them anymore.

web/test/controllers/investigations_controller_test.rb
```diff
@@ -10,6 +10,13 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
     @investigation_one.source = sources(:investigation_one)
     @investigation_one.hazard = hazards(:one)
     Investigation.import force: true
```
broder: I think we should call the import at the very end of the `setup` to ensure that elasticsearch has indexed the `created_at` / `updated_at` fields.

PR #303, opened 2018-11-23
MSPSDS-551: Fixes to record corrective action flow

tom-bell-softwire:
This addresses a number of issues that came up during testing of the functionality for recording corrective actions and test requests/results on a case.

### DIFF
--- a/web/app/controllers/investigations/tests_controller.rb
+++ b/web/app/controllers/investigations/tests_controller.rb
@@ -66,7 +66,7 @@ private
   end
 
   def store_test
-    session[:test] = @test.attributes
+    session[:test] = @test.attributes if @test.valid?(step)
   end
 
   def test_valid?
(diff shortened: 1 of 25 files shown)


web/app/controllers/investigations/tests_controller.rb
```diff
@@ -77,7 +77,7 @@ def update_attachment
   end
 
   def store_test
-    session[:test] = @test.attributes
+    session[:test] = @test.attributes if @test.valid?(step)
```
broder: @MatthewBurstein I'd guess there'll be an equivalent bug with https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/app/controllers/investigations/emails_controller.rb#L61?

web/app/helpers/legislation_helper.rb
```diff
@@ -1,5 +1,5 @@
 module LegislationHelper
-  def legislation_constants
-    Rails.application.config.legislation_constants
+  def relevant_legislation
+    Rails.application.config.legislation_constants["legislation"]&.sort
```
broder: Should we cache this?

tom-bell-softwire: Good idea.

PR #307, opened 2018-11-26
MSPSDS-551: Investigation business and product sort fix

broder:
The was an error on startup with the `.order("investigation_products.created_at")`:
```
2018-11-26T08:06:18.26+0000 [APP/PROC/WEB/0] ERR /home/vcap/deps/0/vendor_bundle/ruby/2.4.0/gems/activerecord-5.2.1/lib/active_record/connection_adapters/postgresql_adapter.rb:603:in `async_exec': PG::UndefinedColumn: ERROR:  column investigation_businesses.created_at does not exist (ActiveRecord::StatementInvalid)
```

which was coming from the elasticsearch import at the bottom of `web/config/environments/production.rb`.

Seems related to https://stackoverflow.com/questions/40247388/rails-5-has-many-through-order-on-through-table. I don't actually know if this still produces the required functionality but it at least fixes the deployment!

### DIFF
--- a/web/app/models/investigation_product.rb
+++ b/web/app/models/investigation_product.rb
@@ -1,4 +1,6 @@
 class InvestigationProduct < ApplicationRecord
   belongs_to :investigation
   belongs_to :product
+
+  default_scope { order('created_at') }
 end
--- a/web/app/models/investigation.rb
+++ b/web/app/models/investigation.rb
@@ -27,12 +27,12 @@ class Investigation < ApplicationRecord
   belongs_to_active_hash :assignee, class_name: "User", optional: true
 
   has_many :investigation_products, dependent: :destroy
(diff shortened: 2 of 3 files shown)


web/app/models/investigation_product.rb
```diff
@@ -1,4 +1,6 @@
 class InvestigationProduct < ApplicationRecord
   belongs_to :investigation
   belongs_to :product
+
+  default_scope { order('created_at') }
```
tom-bell-softwire: Does this work? I tried this and it didn't seem to be respected...

tom-bell-softwire:
Thanks for fixing this! I couldn't get the sorting to work by specifying a default scope on the join tables, but this is certainly preferable to a broken deploy!

PR #310, opened 2018-11-26
MSPSDS-536: Sort investigations

aidayounesi:
Bug fixed


web/app/models/concerns/searchable.rb
```diff
@@ -62,6 +62,10 @@ module Searchable
       ]
     end
 
+    after_update do |document|
+      document.__elasticsearch__.update_document_attributes updated_at: document.updated_at
```
broder: Should we add a TODO comment here referencing the open PR? We could say something like this line fixes the bug, we can remove it once that PR is merged?

aidayounesi: Sure

PR #319, opened 2018-11-27
MSPSDS-NONE: Test fixes for Elasticsearch failures

tom-bell-softwire:
To resolve some of the more common Elasticsearch-related errors we've been seeing in recent test runs, this PR makes the following changes:

* Import all test fixtures into Elasticsearch at the start of each test class
* Remove redundant Elasticsearch imports in tests
* Where imports are still required, refresh the index instead of recreating it
* Ensure all stubbed users are loaded as part of relevant test helper methods
* Minor refactoring of some test setup and assertions

### DIFF
--- a/web/test/controllers/investigations_controller_test.rb
+++ b/web/test/controllers/investigations_controller_test.rb
@@ -3,24 +3,29 @@ require "test_helper"
 class InvestigationsControllerTest < ActionDispatch::IntegrationTest
   setup do
     sign_in_as_admin
-    @investigation_one = investigations(:one)
-    @investigation_two = investigations(:two)
-    @investigation_three = investigations(:three)
-    @investigation_no_products = investigations(:no_products)
-    @investigation_one.source = sources(:investigation_one)
 
     User.all
+
+    @investigation_one = investigations(:one)
+    @investigation_one.created_at = Time.zone.parse('2014-07-11 21:00')
     @investigation_one.assignee = User.find_by(last_name: "Admin")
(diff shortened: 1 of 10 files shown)


tom-bell-softwire:
Not to be merged yet – just testing the fixes on Travis builds.

web/test/controllers/investigations_controller_test.rb
```diff
@@ -3,24 +3,29 @@
...
   setup do
     sign_in_as_admin
-    @investigation_one = investigations(:one)
-    @investigation_two = investigations(:two)
-    @investigation_three = investigations(:three)
-    @investigation_no_products = investigations(:no_products)
-    @investigation_one.source = sources(:investigation_one)
 
-    User.all
+    Investigation.import force: true
```
kicferk1: So we are importing with force:true twice? 
I think force:true makes elasticsearch recreate the index. I wonder if the second time we import it wouldn't be enough to just call it with refresh: true rather than force: true

jasiekmiko: Do we still need this one, given there's https://github.com/UKGovernmentBEIS/beis-mspsds/pull/319/files#diff-080a3807288e352135a86576572b12e4R25?

tom-bell-softwire: Good point!

tom-bell-softwire: Nope!

PR #322, opened 2018-11-27
MSPSDS-550 record email markups

MatthewBurstein:
Following comments [here](https://regulatorydelivery.atlassian.net/secure/RapidBoard.jspa?rapidView=18&modal=detail).

Also implements single table inheritance for correspondence


web/app/models/correspondence.rb
```diff
@@ -4,17 +4,19 @@ class Correspondence < ApplicationRecord
...
   end
 
-  has_one_attached :email_file
-  has_one_attached :email_attachment
   has_many_attached :documents
 
   enum email_direction: {
-      outbound: "Outbound",
-      inbound: "Inbound"
+      outbound: "To",
```
broder: Should this be on `Correspondence::Email`?

web/app/models/correspondence.rb
```diff
@@ -4,17 +4,19 @@ class Correspondence < ApplicationRecord
...
+      inbound: "From"
   }
 
   enum contact_method: {
```
broder: Do we need this now that we have single table inheritance?

web/app/models/correspondence/email.rb
```diff
@@ -0,0 +1,15 @@
...
+  has_one_attached :email_attachment
+
+  def validate_email_file_and_content file_blob
+    if file_blob.nil? && email_subject.empty? && details.empty?
+      errors.add(:email, "Please provide either an email file or a subject and body")
+    end
+  end
+
+  def pretty_email_direction
+    return "To" if email_direction == "outbound"
```
broder: Does the change to the enum above provide this functionality as well?

web/app/models/correspondence.rb
MatthewBurstein: doesn't it need to be in here because it is on the Correspondence table so all correspondences will know about it?

MatthewBurstein: I think since email_direction is a column in the database, this is best left on the top level model for records of that table. Is that right?

web/app/models/correspondence.rb
MatthewBurstein: Single table inheritance won't be fully implemented till the pull request for meetings and phone calls is complete, at which point I suspect this method can be dropped.

web/app/models/correspondence.rb
tom-bell-softwire: With single-table inheritance, I think it's best to put the properties on the child model that they relate to (the base model may have various null columns that are only relevant to specific child models).

Unless `email_direction` has any meaning for a base `Correspondence` record, we should probably move it to `Correspondence::Email` instead.

web/app/controllers/investigations/emails_controller.rb
```diff
@@ -93,24 +94,16 @@ def correspondence_params
   end
 
   def request_params
-    return {} if params[:correspondence].blank?
+    return {} if params[correspondence_params_key].blank?
 
-    params.require(:correspondence).permit(
+    params.require(correspondence_params_key).permit(
       :correspondent_name, :correspondent_type, :contact_method, :phone_number, :email_address, :day, :month, :year,
```
tom-bell-softwire: Do we need `phone_number` ?

web/app/models/correspondence.rb
MatthewBurstein: Thanks Tom have done so

web/app/controllers/investigations/emails_controller.rb
MatthewBurstein: Nope good point. Nor do we need correspondent_type and contact_method

PR #326, opened 2018-11-28
MSPSDS-758: Simplify cases

kicferk1:


web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -12,8 +12,6 @@ book.add_worksheet name: "Cases" do |sheet_investigations|
       investigation.description,
       investigation.assignee ? investigation.assignee.email : "Unassigned",
       investigation.source.show,
-      investigation.hazard&.description,
```
broder: You'll probably need to remove the headers for these as well?

web/app/views/investigations/show.html.slim
```diff
@@ -6,10 +6,7 @@
 
 = render "investigation_heading", investigation: @investigation
 
-- if @investigation.is_case
-  p
-
-- else
+- if !@investigation.is_case
```
broder: Will the linter complain that this should be `unless`?

PR #327, opened 2018-11-28
MSPSDS-NONE: Write more search tests

kicferk1:
Hopefully these have a better chance of catching any elasticsearch bugs that can crash int that can go unnoticed if the local database is not rich enough(like what happened on Monday).


broder:
@kicferk1, do you want to resolve the merge conflicts for this?

PR #329, opened 2018-11-29
Release to staging

kicferk1:


kicferk1:
Can someone with write rights update master(by the one merge commit that staging has and master doesn't)?

kicferk1:
Since fixing elasticsearch might take some time we are merging it now to unblock Rita

PR #332, opened 2018-11-29
MSPSDS-NONE: Add migration to remove empty attachments

kicferk1:


web/db/migrate/20181129123859_remove_empty_attachments.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveEmptyAttachments < ActiveRecord::Migration[5.2]
+  def change
+    # Due to imperfect handling of data removal previously we have some attachments pointing at nils
+    ActiveStorage::Attachment.where(blob: nil).delete_all
```
broder: Should we add this as a recurring SideKiq job?

kicferk1: Maybe? 
That's targeted at fixing int and staging, since removing hazards had an unfortunate effect of also removing associated risk assessment files, that didn't remove their attachments to investigations :(

PR #336, opened 2018-11-30
COS-31 Fix issue with the deploy script

DWRendell:
It looks like the shared-web files we're getting copied to vendor/shared-web/shared-web/* (on my machine at least).

### DIFF
--- a/cosmetics-web/deploy.sh
+++ b/cosmetics-web/deploy.sh
@@ -11,7 +11,7 @@ set -ex
 # CF_PASSWORD: cloudfoundry password
 # SPACE: the space to which you want to deploy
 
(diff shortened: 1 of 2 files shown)


broder:
I think the folder it'll go into will be `vendor/shared-web/shared-web/*` if the `vendor/shared-web/` folder already exists. If just `vendor/` exists (as it should on TravisCI), then it should correctly go into `vendor/shared-web/*`. I'm not 100% sure though so it's worth checking.

The format of `cp` commands that we use in `./worker/deploy.sh` is `cp -a ./worker/clamav/. ./web/clamav/` - does that format fix this problem?

PR #337, opened 2018-11-30
MSPSDS-777: Edit business locations

kicferk1:
Originally expected it to be a small styling and re-routing ticket, but it turned out it required a refactor of addresses. Hence a large quantity of changes that essentially say 'call it location, and use a slightly different structure'.


web/db/migrate/20181129170930_refactor_addresses_into_locations.rb
```diff
@@ -0,0 +1,32 @@
+class RefactorAddressesIntoLocations < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      reversible do |dir|
+        change_table :locations do |t|
+          dir.up do
+            Address.all.each do |address|
```
broder: Does this need to be `Location.all`?

kicferk1: In the middle of reworking this so that it works :P

kicferk1: In the end, yes it should. But it also needs to be called after renaming the table, or else rails magic breaks cause Location model can't find the location table.

PR #338, opened 2018-11-30
MSPSDS-718: Record incoming question

tom-bell-softwire:


web/app/controllers/create/creation_flow_controller.rb
```diff
@@ -0,0 +1,139 @@
+class Create::CreationFlowController < ApplicationController
```
broder: Would `Investigation::CreationFlowController` be a better name?

web/app/controllers/create/creation_flow_controller.rb
```diff
@@ -0,0 +1,139 @@
...
+  include FileConcern
+  include Wicked::Wizard
+
+  before_action :set_reporter
```
broder: Do we need `@reporter` for new?

web/app/controllers/create_controller.rb
```diff
@@ -0,0 +1,34 @@
+class CreateController < ApplicationController
+  before_action :set_available_options
+
+  # GET /create/new
+  def new; end
+
+  # POST /create
+  def create
+    type = request_params[:type]
```
broder: I wonder if we should just put this logic on `investigations#new`. If there's a valid `type` query parameter, then redirect to the relevant controller. If there's no `type` query parameter, render a GET form (with `#new` as the url). What do you think?

web/app/views/application/_reporter_type_selection.html.slim
```diff
@@ -1,43 +1,23 @@
...
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with model: reporter, local: true, url: wizard_path, method: :put do |form|
+      = render "form_validation_display", errors: reporter.errors, model_name: "reporter"
+      fieldset.govuk-fieldset
+        legend.govuk-fieldset__legend.govuk-fieldset__legend--l
+          =  "Who did the question come from?"
+        .govuk-form-group class=("govuk-form-group--error" if reporter.errors[:reporter_type].any?)
+          .govuk-radios
+            - reporter_types = { \
```
broder: Would this be better in a helper method or controller instance variable?

web/app/controllers/create/creation_flow_controller.rb
tom-bell-softwire: Nope!

web/app/controllers/investigations_controller.rb
```diff
@@ -28,7 +28,13 @@ def show
 
   # GET /investigations/new
   def new
-    @investigation = Investigation.new
+    type = params[:type]
```
broder: Should we just inline this?

web/app/views/layouts/_navbar.html.slim
```diff
@@ -7,4 +7,5 @@ nav.app-navigation.govuk-clearfix
     li[class=(params[:controller] == "products" && "app-navigation--current-page")]
       = link_to "Products", products_path, "data-topnav": "products"
   .creation
-    = link_to "Create new", new_create_path, class: "govuk-button govuk-button--start", role: "button", draggable: false
+    = link_to "Create new", new_investigation_path, class: "govuk-button govuk-button--start",
+            role: "button", draggable: false
```
broder: Do we need `draggable` here?

web/app/controllers/investigations_controller.rb
tom-bell-softwire: Done.

PR #340, opened 2018-12-03
COS-29: Add static analysis and testing to cosmetics

DWRendell:
Move testing and static analysis tasks to bash script that only runs them when changes have been made to that component


ci/get-changed-components.sh
```diff
@@ -23,7 +23,11 @@ if [[ "$TOP_LEVEL_CHANGES" =~ cosmetics-web ]]; then
     COMPONENTS="$COMPONENTS cosmetics-web"
 fi
 
-if [[ "$TOP_LEVEL_CHANGES" =~ web ]]; then
+if [[ "$TOP_LEVEL_CHANGES" =~ shared-web ]]; then
+    COMPONENTS="$COMPONENTS web worker cosmetics-web"
+fi
+
+if [[ "$TOP_LEVEL_CHANGES" =~ '^web$' ]]; then
```
broder: We should probably rename `web` to `mspsds-web` at some point but we can do that as part of a different bit of work.

shared-web/.rubocop.yml
```diff
@@ -0,0 +1,16 @@
+inherit_gem:
+  govuk-lint: configs/rubocop/all.yml
+
+AllCops:
+  Exclude:
+    - '/app/app/**/*.axlsx'
```
broder: Do we need the leading `/app`? I think that'll mean it won't work if you run the command on the host machine.

DWRendell: Apparently the paths in this file are relative to the `.rubocop.yml` rather than the working directory when running rubocop, so without the leading app it won't work on the container :(

Maybe a better solution would be to copy `.rubocop.yml` to the /app/ directory as part of the build?

cosmetics-web/.dockerignore
```diff
@@ -235,3 +235,5 @@ $RECYCLE.BIN/
 
 
 # End of https://www.gitignore.io/api/ruby,node,macos,rails,windows
+
+vendor/shared-web/.rubocop.yml
```
broder: Do we need these ignores?

PR #341, opened 2018-12-03
MSPSDS-754: Move adding activities to a separate page

kicferk1:


web/app/views/investigations/activities/add.html.slim
```diff
@@ -0,0 +1,37 @@
...
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      p.govuk-caption-xl
+        = @investigation.question_title_prefix
+        = case_question_text(@investigation).titleize + ': '
+        = @investigation.pretty_id
+      = "New activity"
+    = form_with local: true, url: add_investigation_activities_path, method: :post do |form|
```
broder: If we made this form a GET request then we'd only need to define `activities#add` as a GET - do you think that's neater?

web/app/controllers/investigations/activities_controller.rb
```diff
@@ -1,9 +1,34 @@
 class Investigations::ActivitiesController < ApplicationController
   include ActionView::Helpers::SanitizeHelper
 
-  before_action :set_investigation, only: %i[create]
+  before_action :set_investigation, only: %i[create new add]
   before_action :create_activity, only: %i[create]
 
+  def new; end
+
+  def add
```
broder: Should we just use `new`?

web/app/views/investigations/activities/add.html.slim
```diff
@@ -0,0 +1,37 @@
...
+      - if @activity_type_empty
+        h2.govuk-heading-l.govuk-error-message
+          = "The following error prevented this you from continuing:"
+        ul.govuk-list.govuk-error-message
+          = "Activity type must not be empty"
+      fieldset.govuk-fieldset
+        .govuk-form-group class=("govuk-form-group--error" if @activity_type_empty)
+          .govuk-radios
+            - activity_types = { \
```
broder: Should we put these in a helper method or an instance variable on the controller?

web/app/views/investigations/activities/add.html.slim
kicferk1: How do we distinguish between the first request and any subsequent request that has no type? 

I added a simple error message when the user tries to proceed without selecting a type, in same scenario designs lead to nothing.

web/app/views/investigations/activities/new.html.slim
```diff
@@ -11,8 +11,19 @@
...
-      = "Add comment"
-    = form_with(model: CommentActivity.new, url: investigation_activities_path(@investigation)) do |form|
-      = form.text_area :body, class: "govuk-textarea govuk-!-width-full", aria_describedby: "action__comment-hint",
-              rows: 5, id: "action_comment"
-      = form.submit "Add comment", class: "govuk-button", id: "action_comment_submit"
+      = "New activity"
+    = form_with local: true, url: new_investigation_activity_path, method: :get do |form|
+      - if @activity_type_empty
+        h2.govuk-heading-l.govuk-error-message
+          = "The following error prevented this you from continuing:"
```
broder: Should this be "The following error prevented you from continuing:"?

PR #344, opened 2018-12-03
MSPSDS-789: Add sentry integration

broder:


web/config/application.rb
```diff
@@ -26,8 +26,19 @@ class Application < Rails::Application
     # This changes Rails timezone, but keeps ActiveRecord in UTC
     config.time_zone = "Europe/London"
 
+    config.rails_activesupport_breadcrumbs = true
+
+    # https://github.com/getsentry/raven-ruby/issues/494
```
jasiekmiko: Maybe just being slow, but took me a while to understand how this comment is relating to the code.
Perhaps pointing directly to the solution with `https://github.com/getsentry/raven-ruby/issues/494#issuecomment-417277866`?

web/config/application.rb
```diff
@@ -26,8 +26,19 @@ class Application < Rails::Application
...
     config.exceptions_app = self.routes
 
+    # Inject Sentry logger breadcrumbs
+    require 'raven/breadcrumbs/logger'
+
+    Raven.configure do |config|
+      config.dsn = ENV['SENTRY_DSN']
+      config.sanitize_fields = Rails.application.config.filter_parameters.map(&:to_s)
```
jasiekmiko: Does this essentially say "we already have a list of parameters to be excluded from logging out, use it for raven, too"?

web/config/application.rb
broder: The comment came from https://github.com/getsentry/raven-ruby/blob/master/examples/rails-5.0/config/application.rb#L15 - I think I'll update the comment to point to that instead.

web/config/application.rb
broder: Yep

PR #345, opened 2018-12-03
MSPSDS-734: All results excel export

kicferk1:


web/app/controllers/investigations_controller.rb
```diff
@@ -8,10 +8,16 @@ class InvestigationsController < ApplicationController
...
-    @investigations = @answer.records
-    @results = @answer.results.map do |r|
-      r.merge(record: @answer.records.find_by(id: r._id))
+    respond_to do |format|
+      format.html do
+        @answer = search_for_investigations(20)
+        @results = @answer.results.map { |r| r.merge(record: @answer.records.find_by(id: r._id)) }
+        @investigations = @answer.records
+      end
+      format.xlsx do
```
broder: Should we add a test for this branch?

PR #348, opened 2018-12-04
MSPSDS-798: Remove reporter view details link

kicferk1:
As well as previous steps taken

### DIFF
--- a/web/app/views/application/_report_summary.html.slim
+++ b/web/app/views/application/_report_summary.html.slim
@@ -20,11 +20,10 @@ table.govuk-table
             / TODO: link this to correspondence details
             = link_to "View details", "/", class: "no-print"
 
-      tr.govuk-table__row
-        th.govuk-table__header scope="row" = investigation.is_case ? "Previous steps taken" : "Question type"
-        /TODO: add previous steps taken
-        td.govuk-table__cell
-          = investigation.is_case ? "Previous steps taken" : investigation.question_type
+      - if !investigation.is_case?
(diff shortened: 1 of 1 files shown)


web/app/views/application/_report_summary.html.slim
```diff
@@ -23,11 +20,10 @@ table.govuk-table
...
 
-      tr.govuk-table__row
-        th.govuk-table__header scope="row" = investigation.is_case ? "Previous steps taken" : "Question type"
-        /TODO: add previous steps taken
-        td.govuk-table__cell
-          = investigation.is_case ? "Previous steps taken" : investigation.question_type
+      - if !investigation.is_case?
+        tr.govuk-table__row
+          th.govuk-table__header scope="row" = "Question type"
+          td.govuk-table__cell = investigation.question_type
```
tom-bell-softwire: Should we still be displaying this? The latest designs for recording a question don't capture the question type...

web/app/views/application/_report_summary.html.slim
```diff
@@ -10,9 +10,6 @@ table.govuk-table
         th.govuk-table__header scope="row" = investigation.is_case ? "Reporter" : "Question asked by"
         td.govuk-table__cell
           = investigation.reporter&.reporter_type
-          br
```
tom-bell-softwire: Unrelated to this change, but since you're modifying this view, could you update the display format of the `created_at` date above? I think something like `%d/%m/%Y` is used elsewhere.

PR #351, opened 2018-12-04
MSPSDS-719: Record allegation

tom-bell-softwire:


web/config/initializers/constants.rb
```diff
@@ -1,3 +1,6 @@
 Rails.application.config.legislation_constants = YAML.load_file(
   Rails.root.join("config", "legislation_constants.yml")
 )
+Rails.application.config.taxonomy_constants = YAML.load_file(
```
tom-bell-softwire: If you can think of a better name than `taxonomy_constants` that would be great!

web/app/controllers/investigations/creation_flow_controller.rb
```diff
@@ -11,6 +11,7 @@ class Investigations::CreationFlowController < ApplicationController
 
   # GET /xxx/step
   def show
+    set_page_title
```
broder: Could this be a `before_action`?

web/config/initializers/constants.rb
broder: Would splitting them out into `hazard_constants` and `product_constants` be neater?

tom-bell-softwire: Yes it would! Thanks.

web/app/controllers/investigations/creation_flow_controller.rb
tom-bell-softwire: Good call.

PR #353, opened 2018-12-04
MSPSDS-762: Unify attachments flow

kicferk1:
Some things still need work. 

I'm considering renaming 'documents' to 'attachments' or 'files', though I'm not sure it's worth the effort, seeing how 'documents' is pretty general too.


web/app/views/documents/_document_preview.html.slim
```diff
@@ -1,11 +1,12 @@
 - if document.metadata["safe"]
   - if document.image?
-    = link_to render("documents/image_tag", image: document, dimensions: dimensions), document,
-            class: "image-with-preview", target: "_blank"
+    - class_name = local_assigns[:class_name] || "image-with-preview"
+    = link_to render("documents/image_tag", image: document, dimensions: dimensions, class_name: class_name),
```
broder: Should we be applying the `class_name` to both the link and to the image?

web/config/initializers/sidekiq.rb
```diff
@@ -1,18 +1,18 @@
-def remove_empty_attachments_job
-  remove_attachments_job = Sidekiq::Cron::Job.new(
-    name: 'remove attachments pointing at nothing, midnight every day',
-    cron: '0 0 * * 0-6',
-    class: "RemoveEmptyAttachmentsJob"
+def remove_files_without_attachments_job
+  remove_files_without_attachments_job = Sidekiq::Cron::Job.new(
+    name: 'remove files not attached to anything, midnight every sunday',
+    cron: '0 0 * * 0',
+    class: "RemoveFilesWithoutAttachmentsJob"
```
broder: Did you commit the changes to the job?

kicferk1: Yes, an oversight in previous commit caused a stack overflow(because there was no variable called `remove_files_without_attachments_job' and rails used the method instead).

web/app/jobs/remove_empty_attachments_job.rb
```diff
@@ -1,5 +1,5 @@
-class RemoveEmptyAttachmentsJob < ApplicationJob
+class RemoveFilesWithoutAttachmentsJob < ApplicationJob
   def perform
-    ActiveStorage::Attachment.where(blob: nil).delete_all
+    ActiveStorage::Blob.all.select { |b| return b.attachments.empty? }.delete_all
```
broder: I think this will load all the blobs into an array and then loop through each one. Could we do this using `.where` instead? I think that would then do it in a SQL query rather than in memory on the worker

kicferk1: A straight up 'where' with a condition on attachments doesn't seem to be supported, since that data is in a different table. 'having' might work but it doesn't support delete_all. 

Joining the tables on blob_id and selecting rows without attachments would allow a 1-line where solution, but I'm not sure how clever rails/sql would be with it, potentially creating many copies of blobs we don't actually need. 

So I ended up doing it in 2 lines instead, that should only load bad blobs.

broder: Does https://stackoverflow.com/a/6334682 help?

PR #355, opened 2018-12-05
MSPSDS-769: Adding case creation confirmation banner

aidayounesi:


web/app/controllers/investigations_controller.rb
```diff
@@ -59,7 +59,7 @@ def update
     @investigation.assignee = User.find_by(id: ps[:assignee_id]) if ps[:assignee_id]
     respond_to do |format|
       if @investigation.save
-        format.html { redirect_to @investigation, notice: "Investigation was successfully updated." }
+        format.html { redirect_to @investigation, notice: "Case was successfully updated." }
```
tom-bell-softwire: Should we wrap `@investigation` in `investigation_path()` to be safe here as well?

broder: Are they any different functionally?

tom-bell-softwire: Ah, sorry, I was conflating this with the ["unprotected redirect"](https://brakemanscanner.org/docs/warning_types/redirect/) vulnerability, but that isn't relevant here!

PR #356, opened 2018-12-05
MSPSDS 690 and 691: Small bugs fixed

aidayounesi:


web/app/views/investigations/meetings/context.html.slim
```diff
@@ -15,7 +15,7 @@ main.govuk-main-wrapper role="main"
             | Who was the meeting with?
           .govuk-form-group
             label.govuk-label class="govuk-!-width-two-thirds"
```
tom-bell-softwire: Sorry I didn't spot this before, but I think all the `label` elements in these views should be rendered via the form helper method, e.g.
```
= form.label :correspondent_name, "Name(s)", class: "govuk-label"
```

This has the benefit of correctly associating labels with the relevant input field, which has important implications for accessibility (screen readers, etc.).

Would you mind updating the labels in the content and context views covered by this review?

web/app/views/investigations/meetings/content.html.slim
```diff
@@ -35,7 +35,7 @@ main.govuk-main-wrapper role="main"
           .govuk-form-group
             label.govuk-label class="govuk-!-width-two-thirds"
               | Notes
-            = form.text_area :details, class: "govuk-textarea", rows: 5
+            = form.text_area :details, class: "govuk-textarea", rows: 5, maxlength: 1000
```
tom-bell-softwire: Is there an associated `validates_length_of` validation for this attribute on the `Correspondence` model?

web/app/views/investigations/meetings/context.html.slim
aidayounesi: Thank you. Done

web/app/views/investigations/meetings/content.html.slim
aidayounesi: No, it's probably a good idea to add it so the user can see the relevant error

aidayounesi: Yes, there is

PR #357, opened 2018-12-05
MSPSDS-770 Update case titles

samlionheart:


kicferk1:
Code looks reasonable, but it would nice to have some tests to make sure we know it works exactly as it's supposed to

PR #360, opened 2018-12-05
MSPSDS-763: Tidy up product add/edit pages

tom-bell-softwire:


web/app/helpers/form_options_helper.rb
```diff
@@ -11,13 +11,13 @@ def relevant_legislation
 
   def hazard_types
     Rails.cache.fetch(HAZARD_TYPES_CACHE_KEY, expires_in: 1.hour) do
```
broder: Do we need to cache it if we're not sorting it? I'd guess that `Rails.application.config` is stored in memory but it might be worth checking

web/app/views/products/edit.html.slim
```diff
@@ -1,8 +1,8 @@
 - content_for :page_title, "Edit product"
 - content_for :after_header do
-  = link_to "Back to product", @product, class: "govuk-back-link"
+  = link_to "Back to product", product_path(@product), class: "govuk-back-link"
```
broder: Is `product_path` required here?

web/app/views/products/edit.html.slim
```diff
@@ -1,8 +1,8 @@
...
 = render "product_heading", product: @product
-h2.govuk-heading-l
-  | Product details
-= form_with(model: @product, local: true) do |form|
-  = render "form", form: form, product: @product, countries: @countries, search: false
+.govuk-grid-row
+  .govuk-grid-column-full
+    = form_with model: @product, local: true do |form|
+      = render "form", form: form, product: @product, countries: @countries, search: false, full_width: false
```
broder: Are `search` and `full_width` always going to have the same value?

web/app/views/products/new.html.slim
```diff
@@ -1,14 +1,13 @@
-- content_for :page_title, "New product"
+- page_title = "New product"
```
broder: I think if we're going to do this, we should do it consistently everywhere. I'm not sure it makes the code much simpler though.

web/app/helpers/form_options_helper.rb
tom-bell-softwire: Good point.

web/app/views/products/edit.html.slim
tom-bell-softwire: Nope!

web/app/views/products/edit.html.slim
tom-bell-softwire: Yes, I think so. Since `full_width` is a bit misleading anyway, I'll just use the `search` property to determine the width class.

web/app/views/products/new.html.slim
tom-bell-softwire: The page titles/headers are all a bit inconsistent at the moment. I'll switch this back for now, and will raise a ticket to look at making them more consistent across the site.

tom-bell-softwire: I've raised [MSPSDS-812](https://regulatorydelivery.atlassian.net/browse/MSPSDS-812) to look at this.

PR #361, opened 2018-12-06
MSPSDS-757 Change deployment urls to london instances

jasiekmiko:
+ README and manual deployment docs improvements
+ Including db upgrade from pg 9 to 10

**Warning**
Merging this will move all deployments to London servers, so I will make sure to communicate that process with everyone appropriately


README.md
```diff
@@ -138,7 +138,7 @@ To create a database for the current space:
 
     cf marketplace -s postgres
     cf enable-service-access postgres
-    cf create-service postgres tiny-unencrypted-9.5 mspsds-database
+    cf create-service postgres tiny-unencrypted-10.5 mspsds-database
```
broder: Should we change the `postgres/Dockerfile` to use this version of postgres? I think it's currently using 11.

worker/deploy.sh
```diff
@@ -20,12 +23,14 @@ cp ./worker/apt.yml ./web/apt.yml
...
-./ci/install-cf.sh
-
-cf login -a api.cloud.service.gov.uk -u $CF_USERNAME -p $CF_PASSWORD -o 'beis-mspsds' -s $SPACE
-
-cf push -f ./worker/manifest.yml
-
-cf logout
+cp -a ./shared-web/. ./web/vendor/shared-web/
+
+if [[ -z ${SET_UP} ]] ; then
```
broder: To make this easier to follow, maybe we should move the install CF and login / logout steps to `ci/deploy-if-changed.sh` and then the `<<component>>/deploy.sh` files just handle the `cf push` stuff. What do you think?

worker/deploy.sh
```diff
@@ -10,6 +10,9 @@ set -ex
 # CF_USERNAME: cloudfoundry username
 # CF_PASSWORD: cloudfoundry password
 # SPACE: the space to which you want to deploy
+#
+# If SET_UP is set the script will omit installing cf cli and logging in and out
```
broder: Maybe we should call this `FIRST_RUN` or something? It should just be toggling whether or not there's a `--no-start` flag on the `cf push` command.

README.md
jasiekmiko: good spot, thanks!

worker/deploy.sh
broder: Would there be risks for the deployments if we changed it now?

PR #362, opened 2018-12-06
MSPSDS-617: Export a lot more things in excel

kicferk1:
This ought to give reasonable outputs, I checked locations are created reasonably and other things are just copy-paste of existing ones. 

Testing contents of the generated file properly would require a different gem because axlsx does not seem to support reading excel files. Since this will likely take some time I'll write tests when there are no higher priority stuff.


web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -1,20 +1,30 @@
...
     sheet_investigations.add_row [
       investigation.pretty_id,
       investigation.is_closed? ? "Closed" : "Open",
       investigation.title,
       investigation.description,
+      investigation.product_type,
+      investigation.hazard_type,
       investigation.assignee ? investigation.assignee.email : "Unassigned",
       investigation.source&.show,
+      investigation.reporter ? investigation.reporter.name : nil,
```
broder: Is this equivalent to `investigation.reporter&.name`?

web/app/views/investigations/index.xlsx.axlsx
```diff
@@ -1,20 +1,30 @@
 book = xlsx_package.workbook
 
 book.add_worksheet name: "Cases" do |sheet_investigations|
-  sheet_investigations.add_row %w[ID Status Title Description Assignee Source
-                                  Products Businesses Activities]
+  sheet_investigations.add_row %w[ID Status Title Description ProductType HazardType Assignee Source
```
broder: Do we want a mix of formatting here? i.e. `ProductType` vs `Reporter_Type`

web/test/controllers/investigations_xls_export_test.rb
```diff
@@ -0,0 +1,53 @@
+require "test_helper"
+
+class InvestigationsXlsExportTest < ActionDispatch::IntegrationTest
```
broder: Do you think it's worth testing the content of the spreadsheet? This test feels like it's very closely tied to the underlying implementation...

broder: Just to confirm, are you planning to update the tests to do the above (checking the result count, checking that formulas are escaped) as part of this ticket?

kicferk1: Yes

PR #365, opened 2018-12-06
MSPSDS-734: Fix bug with only 10 results showing up

kicferk1:
For some reason the previous code only returned 10 results. I don't really understand why, I don't think we set this limit anywhere on the way, but oh well, this fixes it.

### DIFF
--- a/web/app/helpers/investigations_helper.rb
+++ b/web/app/helpers/investigations_helper.rb
@@ -1,10 +1,8 @@
 module InvestigationsHelper
   include SearchHelper
 
-  def search_for_investigations(page_size = 0)
+  def search_for_investigations(page_size = Investigation.count)
     result = Investigation.full_search(search_query)
-    return result if page_size.zero?
-
     result.paginate(page: params[:page], per_page: page_size)
   end
 


web/app/helpers/investigations_helper.rb
```diff
@@ -1,10 +1,8 @@
 module InvestigationsHelper
   include SearchHelper
 
-  def search_for_investigations(page_size = 0)
+  def search_for_investigations(page_size = Investigation.count)
```
broder: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-from-size.html - size defaults to 10 :/

broder: Is this default value evaluated every time the method is called or on initialisation? i.e. does the value update if I add another investigation? Other people have suggested just setting it to something like 1000

kicferk1: It updates when you add an investigation, I checked this on the branch with bulk exports when writing test for exporting everything. I needed to add a bunch of cases, and the result I got matched the expected new count.

broder: Nice!

PR #366, opened 2018-12-06
MSPSDS-718: Ensure page title is set when rendering wizard step after validation errors

tom-bell-softwire:


web/app/controllers/investigations/creation_flow_controller.rb
```diff
@@ -2,7 +2,7 @@ class Investigations::CreationFlowController < ApplicationController
   include FileConcern
   include Wicked::Wizard
 
-  before_action :set_page_title, only: %i[show]
+  before_action :set_page_title, only: %i[show update]
```
kicferk1: Makes sense. 

Should you include `create` in here as well? In the case when investigation is valid but something goes wrong with saving it, we get to `render step` again, which presumably causes the same problem.

tom-bell-softwire: Good spot!

PR #367, opened 2018-12-06
MSPSDS-807 Update business fields

samlionheart:


kicferk1:
Should this be on top of MSPSDS-770?

PR #375, opened 2018-12-10
MSPSDS-651 Added surveymonkey feedback link

samlionheart:
Added feedback link that opens in new tab


kicferk1:
Looks good!

I assume someone has access to data coming from this survey and checked that they correspond to what was answered?

PR #379, opened 2018-12-10
Production release

jasiekmiko:
Goals of the release:
- test the release mechanism for prod
- fix issue with mangled worker and web code on the prod server due to imperfect manual set-up release
- allow manual testing the prod environment wiring and set-up


tom-bell-softwire:
Should we be releasing `staging` to Prod, rather than `master`?

PR #384, opened 2018-12-11
MSPSDS-772: Fixes to create comment flow

aidayounesi:
The button should be called "Continue" to match the designs and not "Add comment".

The button should be enabled always.


web/app/views/investigations/activities/comment.html.slim
```diff
@@ -14,5 +14,5 @@
       = "Add comment"
     = form_with(model: CommentActivity.new, url: investigation_activities_path(@investigation)) do |form|
       = form.text_area :body, class: "govuk-textarea govuk-!-width-full", aria_describedby: "action__comment-hint",
-              rows: 5, id: "action_comment"
-      = form.submit "Add comment", class: "govuk-button", id: "action_comment_submit"
+              rows: 5, id: "action_comment", maxlength: 1000
```
tom-bell-softwire: Should we add `validates_length_of` validation to the CommentActivity model as well?

PR #387, opened 2018-12-11
MSPSDS-806: Consistent case titles

samlionheart:


web/app/views/application/_investigations_summary.html.slim
```diff
@@ -1,26 +1,7 @@
...
-        tr.govuk-table__row
-          td.govuk-table__cell
-            = link_to investigation.title, investigation
-          td.govuk-table__cell
-            = investigation.status
-          td.govuk-table__cell
-            = investigation.assignee ? investigation.assignee.email : "Unassigned"
-          td.govuk-table__cell
-            = investigation.updated_at.strftime("%d %B %Y")
+  br
```
tom-bell-softwire: Rather than add a `<br>` element here, I think it would be better to just add a top margin to the container div in the `_cases.html.slim` partial. From the designs, it looks like `class="govuk-!-margin-top-6"` might work.

samlionheart: I agree, thanks

web/app/views/application/_investigations_summary.html.slim
```diff
@@ -1,26 +1,7 @@
 - if parent.investigations.any?
```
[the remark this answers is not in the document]
samlionheart: Thanks for pointing this out, I've done the above

web/app/views/application/_investigations_summary.html.slim
```diff
@@ -1,7 +1,5 @@
-- if parent.investigations.any?
-  br
-    - parent.investigations.each do |investigation|
+- if @product.investigations.any?
```
tom-bell-softwire: Sorry, I wasn't very clear before! I meant to pass in the `investigations` as a parameter to this partial. It's used in a couple of places (the cases tab for both products and businesses), so I think we still need to pass in the array of investigations.

web/app/views/products/tabs/_cases.html.slim
```diff
@@ -1,3 +1,3 @@
-.govuk-grid-row
+.govuk-grid-row class="govuk-!-margin-top-6"
```
tom-bell-softwire: I think you'll also need to do the same in `/businesses/tabs/_cases.html.slim`

samlionheart: Yep, my bad I was just looking over this and realised I messed it up. Should be all good now, thanks for the comments.

web/app/views/application/_investigations_summary.html.slim
samlionheart: Not your fault, thanks for the help. I for some reason forgot about the businesses' cases tab so did it only for products. Corrected now

tom-bell-softwire: Awesome! Thanks.

PR #393, opened 2018-12-12
MSPSDS-780 chore: upgrade keycloak

MatthewBurstein:
Pushed to Irish int and seems to be working there. I tested: 
- login/logout
- forgot password email
- 2FA

### DIFF
--- a/keycloak/Dockerfile
+++ b/keycloak/Dockerfile
@@ -5,8 +5,10 @@ WORKDIR /tmp/keycloak
 RUN mkdir -p ./artifacts
 RUN mkdir -p ./package
 
+ENV keycloakVersion 4.7.0
+
 # Download and unpack Keycloak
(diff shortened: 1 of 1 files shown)


keycloak/Dockerfile
```diff
@@ -5,8 +5,10 @@ WORKDIR /tmp/keycloak
 RUN mkdir -p ./artifacts
 RUN mkdir -p ./package
 
+ENV keycloakVersion 4.7.0
```
broder: Do we need to bump the keycloak dependency version in the various `pom.xml`s? e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/keycloak/providers/email/pom.xml#L17

tom-bell-softwire:
Do we need to update the submodule reference to point to the latest change on `keycloak-sms-authenticator-sns` ?

broder:
Dependabot might pick up the new commit but it might be easier to do it ourselves here.

tom-bell-softwire:
Great, thanks!

PR #411, opened 2018-12-17
Cos 29 fix coverage

DWRendell:


shared-web/lib/shared/web/coveralls_formatter.rb
```diff
@@ -0,0 +1,107 @@
...
+        # Gather the source files.
+        source_files = []
+        result.files.each do |file|
+          properties = {}
+
+          # Get Source
+          properties[:source] = File.open(file.filename, "rb:utf-8").read
+
+          # Return the absolute file path, minus the leading '/'. This will match the path in Github.
+          properties[:name] = file.filename[1..-1]
```
DWRendell: This is the only bit that's changed. Rather than shortening to the relative path, we use a path that will match up with what's in Git, and will have 'web'/'cosmetics-web' in the prefix so Coveralls can tell the difference.

broder: Does it work if we override only this method on `Coveralls::SimpleCov::CoverallsFormatter`? e.g. something like what we were doing here: https://github.com/UKGovernmentBEIS/beis-mspsds/pull/268/files#diff-2e42e344c34219771a35b649feeda933L1

broder: Or maybe our formatter should extend `Coveralls::SimpleCov::CoverallsFormatter`?

DWRendell: Updated. I had to override the format method as well to get it to work, presumably because that's where the get_source_files method is called. Not entirely sure how inheritance works in Ruby, maybe there's a way around it?

broder: That's a little bit weird to me:
```
class A
    def foo
        'foo'
    end

    def print
        puts foo
    end
end

class B < A
    def foo
        'bar'
    end
end

A.new.print
B.new.print
```

outputs "foo" and then "bar" which suggests that the parent class correctly uses the overridden child method...

broder: After looking at this for ages, I think the problem is that we're running an old version of coveralls which looks like https://github.com/lemurheavy/coveralls-ruby/blob/v0.7.1/lib/coveralls/simplecov.rb. If we override the `short_filename` method instead, it should work for our version and for newer versions.

PR #412, opened 2018-12-18
COS-40: Add cosmetics worker service

DWRendell:
Sorry this branch is a bit messy, it's been merged in and out a fair amount.

It sets up the cosmetics-worker service, and makes loading the Cosmetics website send a job to the worker through Redis that logs some text to the console.


cosmetics-web/Gemfile
```diff
@@ -36,6 +35,7 @@ gem 'shared-web', path: './vendor/shared-web'
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug'
+  gem "govuk-lint"
```
broder: Do we need this here if we have it in `shared-web`?

cosmetics-web/config/elasticsearch.yml
```diff
@@ -0,0 +1,17 @@
...
+  <<: *default
+
+test:
+  <<: *default
+  :transport_options:
+    :request:
+      :timeout: 3000
+
+production:
+  :url: <%= ENV["VCAP_SERVICES"] && CF::App::Credentials.find_by_service_name("cosmetics-elasticsearch")["uri"] %>
```
broder: Have we setup the cosmetics elasticsearch instance? Do we need it yet?

cosmetics-web/config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,9 @@
+require 'sidekiq'
```
broder: Do we need this `require` if we have the dependency loading code in https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/shared-web/lib/shared/web.rb#L2?

cosmetics-web/config/redis.yml
```diff
@@ -0,0 +1,12 @@
...
+  host: 'redis'
+
+development:
+  <<: *default
+
+test:
+  <<: *default
+
+production:
+  url: <%= ENV["VCAP_SERVICES"] && CF::App::Credentials.find_by_service_name("cosmetics-redis")["uri"] %>
```
broder: Have we set this up for int?

cosmetics-web/config/sidekiq.yml
```diff
@@ -0,0 +1,6 @@
+---
+:concurrency: 10
+:max_retries: 3
+:queues:
+  - cosmetics
+  - mailers
```
broder: Would this share the same mailer queue locally as MSPSDS?

cosmetics-web/test/test_helper.rb
```diff
@@ -1,7 +1,17 @@
 ENV['RAILS_ENV'] ||= 'test'
```
broder: Why have the permissions on this file changed? i.e. from 644 to 755

cosmetics-worker/README.md
```diff
@@ -0,0 +1,54 @@
+# MSPSDS Background Worker
```
broder: "Cosmetics Background Worker"?

cosmetics-worker/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+This folder contains the configuration to setup the worker processes which support the (website)[../web/README.md].
```
broder: Should we update this link to go to the cosmetics-web README?

cosmetics-worker/apt.yml
```diff
@@ -0,0 +1,5 @@
+---
```
broder: I wonder if we need a `shared-worker` or something like that...

docker-compose.yml
```diff
@@ -61,6 +61,9 @@ services:
       - ${PWD-.}/shared-web:/app/vendor/shared-web
     depends_on:
       - db
+      - elasticsearch
+      - keycloak
+      - redis
```
broder: Should this depend on cosmetics-worker too?

web/app/models/audit_activity/image/add.rb
```diff
@@ -0,0 +1,10 @@
+class AuditActivity::Image::Add < AuditActivity::Image::Base
```
broder: Did you mean to readd these files to web?

cosmetics-web/config/elasticsearch.yml
DWRendell: We don't need it yet, not planning on setting it up until we do.

cosmetics-web/config/initializers/sidekiq.rb
DWRendell: No, this predates that fix, thanks

cosmetics-web/config/redis.yml
DWRendell: Not yet, will add that now

cosmetics-web/config/sidekiq.yml
DWRendell: Currently yes, but I think this has changed as part of Raam's work implementing notify which branches off this branch.

cosmetics-worker/README.md
DWRendell: Good point, will go through this file and update it.

cosmetics-worker/apt.yml
DWRendell: Potentially, I'm not quite convinced it's worth it right now though. I think we'd only be able to commonise this file and the clamav config. Maybe talk about it offline?

DWRendell:
Sorry about the mess this was before, I've cherry-picked everything on top of master so it should be a lot clearer now hopefully.

cosmetics-web/Gemfile
DWRendell: This is an odd one, adding it to shared-web breaks with the new requiring of dependencies from the engine. I've tried changing it to a development dependency and adding
```
Gem.loaded_specs['shared-web'].dependencies.each do |d|
  require d.name unless d.type == :development
end
```
but then rubocop fails. Not sure what's going on

DWRendell:
Wasn't intentional to change file permissions, should be fixed now

cosmetics-web/config/environments/development.rb
```diff
@@ -13,6 +13,8 @@
   config.consider_all_requests_local       = true
   config.action_controller.perform_caching = false
 
+  config.active_job.queue_adapter = :sidekiq
```
broder: Will we need this for the `production.rb` config as well?

docker-compose.yml
broder: Do we need to do this?

docker-compose.ci.yml
```diff
@@ -24,6 +24,11 @@ services:
     image: ${DOCKER_USERNAME}/worker:${TRAVIS_BUILD_NUMBER}
     volumes:
       - ~/bundle-cache:/usr/local/bundle
+  cosmetics-worker:
+    command: tail -f /dev/null
+    image: ${DOCKER_USERNAME}/cosmetics-worker:${TRAVIS_BUILD_NUMBER}
```
broder: Will we need to add an initial version of this image to our docker account? It looks like the build is failing because it can't pull `${DOCKER_USERNAME}/cosmetics-worker`

cosmetics-web/app/controllers/helloworld_controller.rb
```diff
@@ -1,5 +1,6 @@
 class HelloworldController < ApplicationController
   def index
+    TestJob.perform_later
     puts 'Testing!'
```
raamSoftwire: Is it worth leaving this logging in?

broder: I think it's fine to leave it whilst we're still focussing on the infrastructure setup

cosmetics-worker/README.md
```diff
@@ -0,0 +1,54 @@
...
+The worker code is automatically deployed to the relevant environment by Travis CI as
+described in [the root README](../README.md#deployment).
+
+
+### Deployment from scratch
+
+Login to GOV.UK PaaS and set the relevant space as described in [the root README](../README.md#deployment-from-scratch).
+Running the following commands from the root directory will then setup the worker app.
+
+    cp -a ./cosmetics-worker/public/. ./cosmetics-web/public/
```
jasiekmiko: I've found these instructions to get out of sync when deploy script changes for whatever reason. To combat it I've made the deply script the sole source of truth and referenced it in the readmes (see [the change](https://github.com/UKGovernmentBEIS/beis-mspsds/commit/9d19c8b45f93e4a88062c0773f3d7ee2f1c9a5fb) and e,g, [web readme](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/web/README.md#deployment-from-scratch))
You might want to consider something similar here?

PR #416, opened 2018-12-19
Revert Keycloak upgrade

tom-bell-softwire:
An issue with the latest release of Keycloak breaks our ability to fetch users from MSPSDS (the exact cause is still TBD, but seems related to a permissions issue for calls to the Keycloak API when using client credentials authentication).

A (possibly related) issue in Keycloak 4.7.0 also prevents us from being able to restrict access to the Keycloak admin console, but still allow user management (see [KEYCLOAK-9066](https://issues.jboss.org/browse/KEYCLOAK-9066)).

### DIFF
--- a/web/keycloak.rb
+++ b/web/keycloak.rb
@@ -1,0 +1,962 @@
+require 'keycloak/version'
+require 'rest-client'
+require 'json'
+require 'jwt'
+require 'base64'
+require 'uri'
+
+module Keycloak
+  OLD_KEYCLOAK_JSON_FILE = 'keycloak.json'
+  KEYCLOAK_JSON_FILE = 'config/keycloak.json'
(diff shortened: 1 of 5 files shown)


web/keycloak.rb
```diff
@@ -0,0 +1,962 @@
+require 'keycloak/version'
+require 'rest-client'
+require 'json'
+require 'jwt'
+require 'base64'
+require 'uri'
+
+module Keycloak
```
broder: Do we need this file?

tom-bell-softwire: Whoops! This shouldn't be here...

tom-bell-softwire:
Closing this in favour of directly reverting #393 instead.

PR #417, opened 2018-12-19
Cos 43 setting up keycloak integration

raamSoftwire:


.env.keycloak.cosmetics-web
```diff
@@ -0,0 +1,6 @@
+# Note: These are DEVELOPMENT values only - not used in production
```
broder: We could maybe just call this file `.env.keycloak.cosmetics`?

shared-web/app/assets/stylesheets/shared-imports-ie8.scss
```diff
@@ -2,4 +2,5 @@ $govuk-is-ie8: true;
 $govuk-global-styles: true;
 $govuk-image-url-function: 'image-url';
 $govuk-font-url-function: 'font-url';
-@import "govuk-frontend/all";
\ No newline at end of file
+@import "govuk-frontend/all";
```
broder: I think the contents of this file should be:
```
$govuk-is-ie8: true;
@import "shared-imports";
```
That would save us having to make `import` changes twice.

shared-web/config/initializers/keycloak.rb
```diff
@@ -1,11 +1,13 @@
+require 'keycloak'
```
broder: Do we need this `require` anymore?

web/config/routes.rb
```diff
@@ -1,5 +1,7 @@
 # rubocop:disable Metrics/BlockLength
 Rails.application.routes.draw do
+  mount Shared::Web::Engine => '/engine', as: 'shared_engine'
```
broder: We probably want the URL to look like `/session` rather than `/engine/session` - does using concerns as described in https://stackoverflow.com/a/20691448 help?

web/app/controllers/application_controller.rb
```diff
@@ -1,55 +1,9 @@
-class ApplicationController < ActionController::Base
+class ApplicationController < Shared::Web::ApplicationController
```
broder: I think this change is what is causing the helper issues because of https://github.com/rails/rails/blob/92fece96da28d9f641bc8a8db1187b91c94cabfb/actionpack/lib/action_controller/railties/helpers.rb#L19. We could either put the `Shared::Web::ApplicationController` shared code into a concern or add `helper :all` to the `ApplicationController`s in web and cosmetics.

(@tom-bell-softwire)

shared-web/app/controllers/shared/web/sessions_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+      def signin
+        request_and_store_token(auth_code)
+        flash[:notice] = "Signed in successfully." if Shared::Web::KeycloakClient.instance.user_signed_in?
+        redirect_to main_app.root_path
+      rescue RestClient::ExceptionWithResponse => error
+        redirect_to keycloak_login_url, alert: signin_error_message(error)
+      end
+
+      def logout
+        flash[:notice] = "Signed out successfully." if Shared::Web::KeycloakClient.instance.logout
```
broder: Do we have to add `Shared::Web` in front of everything if we're already in the `Shared::Web` module?

shared-web/app/controllers/shared/web/application_controller.rb
```diff
@@ -1,7 +1,52 @@
...
+
+      def authenticate_user!
+        redirect_to helpers.keycloak_login_url unless user_signed_in? || try_refresh_token
+      end
+
+      private
+
+      def find_or_create_user
+        user = Shared::Web::KeycloakClient.instance.user_info
+        User.find_or_create(user)
```
broder: Does this use `Shared::Web::User` or the MSPSDS / cosmetics `User`?

keycloak/configuration/initial-setup.json
```diff
@@ -492,6 +492,143 @@
     "nodeReRegistrationTimeout" : 0,
     "defaultClientScopes" : [ "role_list", "profile", "email" ],
     "optionalClientScopes" : [ "address", "phone", "offline_access" ]
+  }, {
+    "id": "121ec424-71bc-4403-829a-44cf4fcb05c5",
```
broder: @tom-bell-softwire does this configuration look alright to you?

.env.keycloak.cosmetics-web
```diff
@@ -0,0 +1,6 @@
...
+
+# Keycloak client credentials
+KEYCLOAK_AUTH_URL=http://keycloak:8080/auth
+KEYCLOAK_CLIENT_ID=npis-app
```
broder: Are we going to call the service "NPIS"?

web/app/controllers/application_controller.rb
tom-bell-softwire: It feels like it would be cleaner to put this logic in a concern instead, which would make it clearer what its functionality is.

.env.keycloak.cosmetics-web
raamSoftwire: Done.

.env.keycloak.cosmetics-web
raamSoftwire: Done.

shared-web/app/assets/stylesheets/shared-imports-ie8.scss
raamSoftwire: That makes sense, thanks. I'll done that now.

shared-web/config/initializers/keycloak.rb
raamSoftwire: Nope, that got fixed recently.

web/app/controllers/application_controller.rb
broder: I think you'll need to keep the `Shared::Web::ApplicationController` as `Shared::Web::SessionsController` should extend from it. You can move most of the user methods to the concern though and just use it in `web` and `cosmetics-web`

raamSoftwire: I've done this in the review markups below, but it still doesn't pull in the helpers. Is there something 
I've missed? 

`helper :all` does work for the record.

broder: So this is now back to extending from `ActionController::Base`? You might need to restart web as well.

web/app/models/user.rb
```diff
@@ -1,39 +1,10 @@
...
-    else
-      @records ||= []
-    end
-  end
-
-  def full_name
-    "#{first_name} #{last_name}"
-  end
-
   def has_role?(role)
```
tom-bell-softwire: Could this method be moved to `Shared::Web::User` as well?

web/app/controllers/application_controller.rb
raamSoftwire: There were a few different things going on here, mainly around being able to access helper methods again. But Tom was able to pair with me on this and I've pushed those changes as the most recent commit.

web/app/models/user.rb
raamSoftwire: It could be, and probably will be at some point. But at the moment only web is using it, so it felt right to leave it there for now. Happy to move it now though if you disagree.

shared-web/app/controllers/shared/web/sessions_controller.rb
raamSoftwire: I've tried removing it from various places and it seems to blow up, showing `uninitialized constant Shared::Web::ApplicationController::ApplicationConcern`. I agree it is ugly to have to call `Shared::Web` everywhere.

raamSoftwire: I've only found one safe place to remove it so far. In the application_controller for shared web.

web/config/routes.rb
raamSoftwire: We weren't able to make it work, so I have pushed a solution using the first answer (ie. to update the route from `/engine` to `/`).

shared-web/app/controllers/shared/web/application_controller.rb
raamSoftwire: I believe this is the MSPSDS user, because the `self.find_or_create(user)` method is only defined in `web/user.rb`

broder: Does that mean that the shared web code depends on the web code?

raamSoftwire: Sorry, my mistake, it lives in `shared/web/user.rb` so no it doesn't depend on web code. But I am again unsure as to which User is being called.

broder: I wonder if we should only have the stuff directly related to `authenticate_user!` in shared web and then have each app (i.e. MSPSDS and cosmetics) implement the `current_user` stuff themselves (given that the method to load the user object might be different for both)

web/app/models/user.rb
raamSoftwire: For the record, I've now moved it to shared web.

shared-web/app/controllers/shared/web/application_controller.rb
raamSoftwire: Thanks DBR, I've kept the `authenticate_user!` in shared web and moved the `current_user` into a UserService which exists in both MSPSDS and cosmetics now.

raamSoftwire:
The current status of this work is that you are able to log in and out of both services as expected. I am still looking at two issues

1. Sometimes, on the first run, I get a `InvalidAuthenticityToken` on both sites.
2. On the MSPSDS, calls to `current_user` are causing the site to crash.

Is it worth me merging master before or after I fix these issues?

broder:
I'd probably try and merge master...

raamSoftwire:
I have now merged and both of the above issues have been resolved.

shared-web/app/controllers/shared/web/application_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Shared
   module Web
     class ApplicationController < ActionController::Base
-      protect_from_forgery with: :exception
```
tom-bell-softwire: Do we still want this line?

shared-web/app/controllers/shared/web/sessions_controller.rb
```diff
@@ -0,0 +1,40 @@
+module Shared
+  module Web
+    class SessionsController < ApplicationController
+      include Shared::Web::LoginHelper
```
tom-bell-softwire: Is this needed, if `ApplicationController` includes `AuthenticationConcern` which in turn includes `LoginHelper`?

shared-web/app/helpers/shared/web/login_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+    module LoginHelper
+      def keycloak_login_url
+        Shared::Web::KeycloakClient.instance.login_url(shared_engine.signin_session_url)
+      end
+
+      def login_page?
+        Keycloak.keycloak_controller == controller_name
+      end
+
+      def cookie_name
```
tom-bell-softwire: Is this method used in any of the views? If not, could it be moved to `AuthenticationConcern`?

shared-web/app/models/shared/web/user.rb
```diff
@@ -0,0 +1,35 @@
+module Shared
+  module Web
+    class User < ActiveHash::Base
+      include ActiveHash::Associations
+
+      field :first_name
+      field :last_name
+      field :email
+
```
tom-bell-softwire: Extra empty line.

cosmetics-web/config/routes.rb
```diff
@@ -1,58 +1,5 @@
...
-  #   resources :posts, concerns: :toggleable
-  #   resources :photos, concerns: :toggleable
-
-  # Example resource route within a namespace:
-  #   namespace :admin do
-  #     # Directs /admin/products/* to Admin::ProductsController
-  #     # (app/controllers/admin/products_controller.rb)
-  #     resources :products
-  #   end
+  mount Shared::Web::Engine => '/engine', as: 'shared_engine'
```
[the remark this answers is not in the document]
raamSoftwire: Good spot, I've just corrected that. I changed it for MSPSDS but not for Cosmetics. Thanks Tom!

cosmetics-web/Dockerfile
```diff
@@ -12,8 +12,8 @@ RUN echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources
 RUN curl -sS https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -
 RUN echo "deb http://dl.google.com/linux/chrome/deb/ stable main"  | tee /etc/apt/sources.list.d/google.list
 
-RUN apt-get update && apt-get install -y \ 
-  build-essential \ 
+RUN apt-get update && apt-get install -y --allow-unauthenticated\
```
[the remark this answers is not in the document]
raamSoftwire: It looks like my Docker image got into a weird state and I needed this option to get round it. I have since rebuilt the image using the `--no-cache` option which seems to have solved the problem. So I have now removed this.

keycloak/configuration/initial-setup.json
raamSoftwire: I guess that is the plan, to use the Keycloak console. That was how I generated the code for this file in the first place. But I'd be keen to pair on Prod when we get to that stage.

mspsds-web/app/models/user.rb
```diff
@@ -1,17 +1,12 @@
...
-  field :first_name
-  field :last_name
-  field :email
-
+class User < Shared::Web::User
   has_many :activities, dependent: :nullify
   has_many :investigations, dependent: :nullify, foreign_key: "assignee_id", inverse_of: :user
   has_many :user_sources, dependent: :delete
 
   belongs_to :organisation
```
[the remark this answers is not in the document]
raamSoftwire: Yes, I agree. I've moved it now.

mspsds-web/config/initializers/lograge.rb
```diff
@@ -5,7 +5,7 @@
 
   config.lograge.custom_payload do |controller|
     extra_payload = {}
-    extra_payload[:user_id] = controller.current_user.id if controller.respond_to?(:current_user) && controller.current_user
+    extra_payload[:user_id] = controller.current_user.id if controller.respond_to?(:user_signed_in?) && controller.current_user
```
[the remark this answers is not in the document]
raamSoftwire: Good point, I've changed this back.

shared-web/app/controllers/shared/web/application_controller.rb
raamSoftwire: This line is now in both application controllers for MSPSDS and Cosmetics. Should I move them both out and just leave this line in?

shared-web/app/controllers/shared/web/sessions_controller.rb
raamSoftwire: Cool, I thought we did need it. But I've tested it and I guess not. Thanks

shared-web/app/helpers/shared/web/login_helper.rb
raamSoftwire: I've moved it to the `AuthenticationConcern`, thanks for reminding me of the _helpers for views, concerns for models and controllers_ pattern that we discussed before Christmas.

cosmetics-web/spec/controllers/helloworld_controller_spec.rb
```diff
@@ -1,10 +1,4 @@
 require 'rails_helper'
 
 RSpec.describe HelloworldController, type: :controller do
-  describe "GET #index" do
-    it "returns a success response" do
-      get :index
-      expect(response).to be_successful
```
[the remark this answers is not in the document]
raamSoftwire: I'm not sure that would work. We had it set to ` (expect(response.status).to eq(302))` and that was passing locally but failing on Travis. But I'll give it a go.

shared-web/Gemfile.lock
```diff
@@ -0,0 +1,274 @@
+PATH
```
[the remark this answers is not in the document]
raamSoftwire: Does that mean I can safely delete it?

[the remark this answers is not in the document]
broder: Yes

cosmetics-web/spec/controllers/helloworld_controller_spec.rb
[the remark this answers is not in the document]
broder: I hope that would have been a spurious failure.

cosmetics-web/Gemfile.lock
```diff
@@ -20,6 +20,7 @@ PATH
     shared-web (1.0.0)
       aws-sdk-s3
       govuk_notify_rails
+      keycloak
```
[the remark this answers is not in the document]
raamSoftwire: I'm not sure. Are you able to explain what it's doing at the moment?

cosmetics-web/spec/controllers/helloworld_controller_spec.rb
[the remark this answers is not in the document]
raamSoftwire: I was able to recreate it locally by commenting out my environment variables. Tom and I looked at it yesterday and we think it is because Travis needs reconfiguring to handle the Keycloak changes...?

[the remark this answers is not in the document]
broder: Ah, good spot! I'll come and help you sort this out.

PR #418, opened 2018-12-19
Revert Keycloak upgrade

tom-bell-softwire:
Reverts UKGovernmentBEIS/beis-mspsds#393

An issue with the latest release of Keycloak breaks our ability to fetch users from MSPSDS (the exact cause is still TBD, but seems related to a permissions issue for calls to the Keycloak API when using client credentials authentication).

A (possibly related) issue in Keycloak 4.7.0 also prevents us from being able to restrict access to the Keycloak admin console, but still allow user management (see [KEYCLOAK-9066](https://issues.jboss.org/browse/KEYCLOAK-9066)).

### DIFF
--- a/keycloak/Dockerfile
+++ b/keycloak/Dockerfile
@@ -5,10 +5,8 @@ WORKDIR /tmp/keycloak
 RUN mkdir -p ./artifacts
 RUN mkdir -p ./package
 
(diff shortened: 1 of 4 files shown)


broder:
Potentially!

PR #420, opened 2018-12-20
MSPSDS-484: Enforce password complexity

tom-bell-softwire:
Changes apply to both the `master` and `mspsds` realms.


tom-bell-softwire:
Changes apply to both the `master` and `mspsds` realms.

tom-bell-softwire:
This won't update any existing Keycloak instance, since I haven't found a way to import the config into an existing realm without messing up any user/client details that have already been added.

I'll manually apply these updates for now, but is it worth creating a tech debt ticket to investigate a reliable way to apply config updates on deploy?

tom-bell-softwire:
I've raised [MSPSDS-857](https://regulatorydelivery.atlassian.net/browse/MSPSDS-857) to look at updating Keycloak config during deployment.

PR #421, opened 2018-12-20
MSPSDS-530: Restrict case legal reasons

kicferk1:
Blocked by designs. 
Need to know what to display when someone who is not supposed to see tries to access website.
Need to know what activity for changing visibility looks like.

Currently access criteria is based on users, needs to be updated to organisations(hopefully as part of a separate ticket). 

We use a filter on elasticsearch query to not return sensitive cases unless users are whitelisted. For parts of project that use rails relations to get cases instead, we compute visibility of cases on the fly.

### DIFF
--- a/web/app/helpers/investigations_helper.rb
+++ b/web/app/helpers/investigations_helper.rb
@@ -25,8 +25,8 @@ module InvestigationsHelper
 
   def get_must_filters
     must = []
-    must.push(get_status_filter) if get_status_filter.present?
-    must.push(get_visibility_filter)
+    must << get_status_filter if get_status_filter.present?
+    must << get_visibility_filter
     { must: must }
   end
 
@@ -42,6 +42,11 @@ module InvestigationsHelper
   end
 
   def get_visibility_filter
+    # TODO: this is bad when multiple users use service
+    # Need to store some sort of visibility object that will be matched to the one we send here
+    # String of organisations allowed to see it would work. Relation of such organizations would work if we include
+    # them in index
+    # Then we just need to send our organization name in place of can_be_displayed_string.
     { term: { can_be_displayed_string: 'true' } }
   end
 
--- a/web/app/helpers/investigations_helper.rb
+++ b/web/app/helpers/investigations_helper.rb
@@ -42,11 +42,6 @@ module InvestigationsHelper
   end
 
   def get_visibility_filter
-    # TODO: this is bad when multiple users use service
-    # Need to store some sort of visibility object that will be matched to the one we send here
-    # String of organisations allowed to see it would work. Relation of such organizations would work if we include
-    # them in index
-    # Then we just need to send our organization name in place of can_be_displayed_string.
     { term: { who_can_see: current_user.id } }
   end
 
--- a/web/app/helpers/investigations_helper.rb
+++ b/web/app/helpers/investigations_helper.rb
@@ -21,6 +21,7 @@ module InvestigationsHelper
     filters = {}
     filters.merge!(get_status_filter)
     filters.merge!(get_assignee_filter)
+    filters.merge!(get_visibility_filter)
   end
 
   def get_status_filter
@@ -65,6 +66,10 @@ module InvestigationsHelper
     assignee_terms = format_assignee_terms(assignees)
     excluded_assignee_terms = format_assignee_terms(excluded_assignees)
     { should: assignee_terms, must_not: excluded_assignee_terms }
+  end
+
+  def get_visibility_filter
+    { must: { term: { can_be_displayed: 'true' } } }
   end
 
   def format_assignee_terms(assignee_array)
--- a/web/app/helpers/investigations_helper.rb
+++ b/web/app/helpers/investigations_helper.rb
@@ -19,9 +19,15 @@ module InvestigationsHelper
 
   def filter_params
     filters = {}
-    filters.merge!(get_status_filter)
+    filters.merge!(get_must_filters)
     filters.merge!(get_assignee_filter)
-    filters.merge!(get_visibility_filter)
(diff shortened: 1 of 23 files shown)


web/app/helpers/investigations_helper.rb
```diff
@@ -31,7 +38,18 @@ def get_status_filter
...
              end
-    { must: { term: status } }
+    { term: status }
+  end
+
+  def get_visibility_filter
+    {
+      bool: {
+        should: [
+          { term: { who_can_see: current_user.id } },
```
jasiekmiko: `who_can_see` is a custom function we provide, right?
Once we move to group-based visibility, is it going to be updated when those group memberships change, or only when there are changes to the case itself?

web/app/models/investigation.rb
```diff
@@ -93,6 +94,25 @@ def status
     is_closed? ? "Closed" : "Open"
   end
 
+  def visibility
+    is_private ? "Private - Only my organisation" : "Public - Visible to all"
+  end
+
+  def can_be_displayed
```
jasiekmiko: Would this work better as `is_visible` to match the terminology around?

web/app/models/investigation.rb
```diff
@@ -93,6 +94,25 @@ def status
...
+    return true unless is_private
+
+    [assignee, source&.user].include?(current_user)
```
jasiekmiko: I believe this is an example of what your comment about access criteria relates to?
If so, it'd be good to tag it with a TODO and point to the separate ticket you want it to be :)

web/app/models/investigation.rb
```diff
@@ -93,6 +94,25 @@ def status
...
+  end
+
+  def who_can_see
+    return [] unless is_private
+
+    users = User.all.select do |u|
```
jasiekmiko: This mapping is not actually transforming isn't it, other than removing the duplicate between `assignee` and `source.user`, right? Is there something hidden here that I'm missing?
O/w this can be simplified to mapping `[assingee, source&.user] to `id`s and de-duped :)

web/app/helpers/investigations_helper.rb
kicferk1: Once we move to group visibility the ES index is not going to change when users change membership(if we stay with this system). 

If a user's group changes, it will affect this filter, cause when we change to group-based query we will send an id of a group here. Probably should add a todo to make sure we don't forget.

jasiekmiko: Ah, and the `who_can_see` method will return a list of group ids then?
Yup, that makes sense - yes those TODOs would be very helpful! :)

mspsds-web/app/models/investigation.rb
```diff
@@ -93,6 +94,25 @@ def status
     is_closed? ? "Closed" : "Open"
   end
 
+  def visibility
+    # TODO: Update to organisations when we get them
```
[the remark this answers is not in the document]
kicferk1: Yes, MSPSDS-859

mspsds-web/app/controllers/application_controller.rb
```diff
@@ -1,12 +1,15 @@
 class ApplicationController < ActionController::Base
   include Pundit
+  rescue_from Pundit::NotAuthorizedError, with: :forbidden
```
broder: Nice!

mspsds-web/app/policies/investigation_policy.rb
```diff
@@ -2,4 +2,11 @@ class InvestigationPolicy < ApplicationPolicy
   def destroy?
     @user.has_role? :admin
   end
+
+  def visible?
```
broder: Should we call this `show?` to be consistent with the other pundit policies?

mspsds-web/app/models/investigation.rb
```diff
@@ -93,6 +93,17 @@ def status
     is_closed? ? "Closed" : "Open"
   end
 
+  def visibility
```
broder: We could maybe call this `display_visibility` or `pretty_visibility` to make it clear that it should probably only be used in views.

mspsds-web/app/views/investigations/visibility.html.slim
```diff
@@ -0,0 +1,18 @@
+- content_for :page_title, "Update visibility"
+- content_for :after_header do
+  = link_to "Back to case", @investigation, class: "govuk-back-link"
```
broder: Would this ever need to be `"Back to question"`?

mspsds-web/app/views/investigations/visibility.html.slim
```diff
@@ -0,0 +1,18 @@
...
+= render "minimal_investigation_heading", investigation: @investigation, title: "Visibility information"
+h2.govuk-heading-l
+  = "#{case_question_text(@investigation).titleize} visibility"
+= form_with(model: @investigation, local: true) do |form|
+  .govuk-form-group
+    fieldset.govuk-fieldset
+      .govuk-radios
+        .govuk-radios__item
+          = form.radio_button :is_private, false, class: "govuk-radios__input"
+          = form.label :is_private, "Public - Visible to all", class: "govuk-label govuk-radios__label"
```
broder: Should we commonise these strings with the strings in the `visibility` method in `investigation.rb`?

mspsds-web/config/routes.rb
```diff
@@ -88,6 +89,7 @@
   end
 
   match "/404", to: "errors#not_found", via: :all
+  match "/403", to: "errors#forbidden", via: :all
```
broder: Do you need to add a `forbidden` method to `errors_controller.rb` as well (with the right status code)?

PR #422, opened 2018-12-21
MSPSDS-856: Tidy docker setup

broder:
This will be quite a breaking change for people's local setups...

It'll probably be easier to review this commit by commit.


cosmetics-web/README.md
```diff
@@ -0,0 +1,104 @@
+# Cosmetics Website
+
+This folder contains the code for the Cosmetics website.
+This folder also contains the code for the (background worker)[../cosmetics-worker/README.md].
```
jasiekmiko: I think you've got the parentheses and brackets the wrong way 'round here?

cosmetics-worker/README.md
```diff
@@ -33,13 +33,6 @@ This provisions the app in Cloud Foundry.
 
 This configures rails to use the production database amongst other things.
 
-    cf set-env cosmetics-worker AWS_ACCESS_KEY_ID XXX
```
jasiekmiko: Are these simply not used by cosmetics at this point in time? (it's only for S3 that we use it on mspsds, isn't it?)

broder: S3 is not used yet. It should be fairly straightforward to add these lines back in when it does start using it.

cosmetics-web/README.md
broder: Good spot!

PR #424, opened 2018-12-21
COS-41: Set up notify integration

broder:


DWRendell:
It looks good to me :)

PR #425, opened 2018-12-21
COS-29: Add initial spec to cosmetics

broder:
I want to try out rspec on Cosmetics and we'll maybe update MSPSDS to use it in the future.


mspsds-web/bin/rake
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env ruby
+begin
+  load File.expand_path('../spring', __FILE__)
```
tom-bell-softwire: Out of interest, what is this change doing?

broder: https://github.com/rails/spring is designed to speed up development by keeping the application running in the background.

I thought it would be worth adding to MSPSDS as it was added to cosmetics by default. It seems to be causing the tests to fail though (and has mixed opinions online) so I'm going to remove it to avoid adding any more magic.

PR #430, opened 2018-12-31
MSPSDS-858: Separate update assignee and status and validate them

kicferk1:


mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,35 +45,40 @@ def new
   end
 
   # GET /cases/1/status
-  def status; end
+  # PUT /cases/1/assign
```
broder: Should be `status` rather than `assign`

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,35 +45,40 @@ def new
...
+  def status
+    return if request.get?
 
-  # GET /cases/1/assign
-  def assign; end
-
-  # PATCH/PUT /cases/1
-  # PATCH/PUT /cases/1.json
-  def update
     ps = investigation_update_params
```
broder: I'm not sure that we need to use `investigation_update_params` anymore - we can just reference the specific parameter in each action.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -106,4 +111,17 @@ def investigation_update_params
     end
     params.require(:investigation).permit(:is_closed, :status_rationale, :assignee_id)
   end
+
+  def respond_to_update(origin)
+    respond_to do |format|
+      if @investigation.save
+        format.html { redirect_to @investigation, notice: "Case was successfully updated." }
```
broder: Could it be a "Question" rather than a "Case"?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,35 +45,40 @@ def new
...
+      return
     end
+
+    @investigation.assignee = User.find_by(id: ps[:assignee_id]) if ps[:assignee_id]
+    respond_to_update(:assign)
   end
 
+  # PATCH/PUT /cases/1
+  # PATCH/PUT /cases/1.json
+  def update; end
```
broder: Do we need to keep this action?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,35 +45,40 @@ def new
...
+
+  # GET /cases/1/assign
+  # PUT /cases/1/assign
+  def assign
+    return if request.get?
+
+    ps = investigation_update_params
+    unless ps[:assignee_id]
+      @investigation.errors.add(:assignee, :invalid, message: "Assignee should exist")
+      return
```
broder: Does this work? Don't we need to call the `render` method?

kicferk1: It works.

broder: What about if you make a JSON API request?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,33 +45,36 @@ def new
...
+  end
+
+  # GET /cases/1/assign
+  # PUT /cases/1/assign
+  def assign
+    return if request.get?
+
+    ps = assignee_update_params
+    unless ps[:assignee_id]
+      @investigation.errors.add(:assignee, :invalid, message: "Assignee should exist")
```
broder: Is this check already done here https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/investigation.rb#L184?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,33 +45,36 @@ def new
...
   # GET /cases/1/status
-  def status; end
+  # PUT /cases/1/status
+  def status
+    return if request.get?
 
-  # GET /cases/1/assign
-  def assign; end
+    ps = status_update_params
+    unless ps[:is_closed]
```
jasiekmiko: Am I missing the point here? but if `ps[:is_closed]` is `false` (ie one of the two legit values?) it will execute the conditional body and display an error?

kicferk1: You are right, I just copied the logic from assign without thinking about the data format. 
Should be fixed now.

mspsds-web/app/models/investigation.rb
```diff
@@ -179,12 +177,6 @@ def case_title
     title.presence || "Untitled case"
   end
 
-  def validate_assignment
```
broder: If I make a `PUT` request to `/cases/1/assign` with `assignee_id=not_blank_but_not_a_valid_id`, what happens? Does it unassign the case?

kicferk1: It probably does, now that the validation for it was removed from model. 

I added a check for the user actually existing.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -45,33 +45,36 @@ def new
...
+    respond_to_update(:status)
+  end
+
+  # GET /cases/1/assign
+  # PUT /cases/1/assign
+  def assign
+    return if request.get?
+
+    ps = assignee_update_params
+    if User.where(id: ps[:assignee_id]).empty?
```
broder: I think this means that we're now doing 2 queries (one to check they exist and one to assign below). It's probably fine though as the users info is held in memory.

PR #431, opened 2018-12-31
MSPSDS-826: Sanitize markdown text

kicferk1:
This uses sanitised copies of data to build the things going into markdown(sanitised as in: *, _, ~ are escaped). 

I'm not sure we want to allow users to input things that need to be sanitised later, or that we want to store them in our database, but that's probably something to be discussed later.


mspsds-web/app/models/audit_activity/corrective_action/base.rb
```diff
@@ -8,7 +8,7 @@ class AuditActivity::CorrectiveAction::Base < AuditActivity::Base
   private_class_method def self.from(corrective_action)
     activity = self.create(
       title: corrective_action.summary,
-      body: self.build_body(corrective_action),
+      body: self.build_body(self.sanitize_object(corrective_action)),
```
broder: Should we just use the `sanitize_text` method in `build_body` rather than have the more complicated `sanitize_object` method?

kicferk1:
The sanitisation in application_helper to make sure the output only includes br tags can well be done in the markdown text sanitisation, but at the cost of clarity in markdown method. Right now html_safe makes sense there because one line above we can see it is probably safe. But if we move it sanitisation elsewhere(and copy it over to 20 different files) it becomes harder to track. 

So in the current form, yes we need it, and I'd rather keep different sanitisations separate.

mspsds-web/app/models/activity.rb
```diff
@@ -27,6 +27,10 @@ def subtitle_slug; end
 
   def search_index;  end
 
+  def self.sanitize_text(text)
```
broder: It would maybe be a bit neater to do something like 
```
class String
  def markdown_sanitize
    # ...
  end
end
```
and then call it with e.g. `relationship.titleize.markdown_sanitize` but it doesn't really make much of a difference where the method is.

PR #437, opened 2019-01-02
MSPSDS-597: Redirect to original path after login

kicferk1:


mspsds-web/app/controllers/application_controller.rb
```diff
@@ -26,7 +26,16 @@ def user_signed_in?
...
   def authenticate_user!
-    redirect_to helpers.keycloak_login_url unless user_signed_in? || try_refresh_token
+    if user_signed_in? || try_refresh_token
+      if session[:requested_url].present?
+        redirect_url = session[:requested_url]
+        session[:requested_url] = nil
+        redirect_to redirect_url
+      end
+    else
+      session[:requested_url] = request.original_url
```
broder: I was expecting that we'd do this by updating the `keycloak_login_url` to send us to something like `https://keycloak/qq/auth?qq=qq&redirect_uri=localhost/session/signing?redirect_uri=qq` i.e. having a nested `redirect_uri` parameter. We'd then update https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/controllers/sessions_controller.rb#L13 to take us to the root path or the `redirect_uri`.

I don't know which way is preferable though - do you have any opinions?

kicferk1: I think url param is more in line with what we already do, so I changed it to work that way.

mspsds-web/app/helpers/login_helper.rb
```diff
@@ -1,9 +1,18 @@
...
+  def keycloak_login_url(request_path=nil)
+    KeycloakClient.instance.login_url(get_session_url_with_redirect(request_path))
   end
 
   def login_page?
     Keycloak.keycloak_controller == controller_name
   end
+
+  def get_session_url_with_redirect(request_path=nil)
+    if signin_session_url.exclude? "?"
```
broder: There are helpers to do this (see https://stackoverflow.com/a/26867426) which might make this a little bit neater?

mspsds-web/app/controllers/sessions_controller.rb
```diff
@@ -4,15 +4,15 @@ class SessionsController < ApplicationController
...
-    redirect_to keycloak_login_url
+    redirect_to keycloak_login_url(request.original_fullpath)
   end
 
   def signin
-    request_and_store_token(auth_code)
+    request_and_store_token(auth_code, params[:request_path])
     flash[:notice] = "Signed in successfully." if KeycloakClient.instance.user_signed_in?
-    redirect_to root_path
+    redirect_to params[:request_path] || root_path
```
broder: Would this be vulnerable to https://stevetabernacle.github.io/blog/open-redirects-the-vulnerability-class-no-one-but-attackers-cares-about/ or is it strict about only working for paths?

kicferk1: It would, I added a check to only redirect to a request_path if it's a path

PR #452, opened 2019-01-07
COS-28 Set up active storage with AWS in production and local storage…

DWRendell:
… in development


cosmetics-web/README.md
```diff
@@ -79,6 +79,12 @@ To create a redis instance for the current space.
 
 Larger options should be considered if required. The current worker (sidekiq) only works with the unclustered version.
 
+#### S3
+
+When setting up a new environment, you'll also need to create an AWS user called `cosmetics-SPACE-NAME` and keep a note of the Access key ID and secret access key.
+Give this user the AmazonS3FullAccess policy.
```
broder: `AmazonS3FullAccess` isn't what we want. I think we'll want to do something like the "Policy for Programmatic Access" section from https://aws.amazon.com/blogs/security/writing-iam-policies-how-to-grant-access-to-an-amazon-s3-bucket/ - do you want to give that a go for cosmetics and check that we still can access what we need and can't access what we don't?

cosmetics-web/db/schema.rb
```diff
@@ -10,9 +10,148 @@
...
+  create_table "cmr_materials", force: :cascade do |t|
+    t.string "name"
+    t.string "cas_registry_number"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.bigint "component_id"
+    t.index ["component_id"], name: "index_cmr_materials_on_component_id"
+  end
+
+  create_table "components", force: :cascade do |t|
```
broder: Have these come from the other branch?

cosmetics-web/README.md
broder: I've added the policy (`cosmetics-int-s3`) - do you want to update the README?

cosmetics-web/db/schema.rb
DWRendell: Yes, oops. I've rebuilt the schema without them

cosmetics-web/README.md
DWRendell: Done

DWRendell:
Fixed the filemode changes

cosmetics-web/README.md
```diff
@@ -79,6 +79,12 @@ To create a redis instance for the current space.
 
 Larger options should be considered if required. The current worker (sidekiq) only works with the unclustered version.
 
+#### S3
+
+When setting up a new environment, you'll also need to create an AWS user called `cosmetics-SPACE-NAME` and keep a note of the Access key ID and secret access key.
+Give this user the `cosmetics-int-s3` policy.
```
broder: We won't want the `cosmetics-int-s3` policy for all environments. Should we say something like "Create a policy for this user similar to the [Policy for Programmatic Access from the AWS docs](https://aws.amazon.com/blogs/security/writing-iam-policies-how-to-grant-access-to-an-amazon-s3-bucket/) but specifically for the new bucket".

(There's also still a filemode change on this file)

DWRendell: Oh of course, d'oh. 

That's annoying, it be VS Code making the filemode changes...

aidayounesi:
Same as COSBETA-5 
I think, it can be merged now because the coverage decreasement will be sorted out when COSBETA-22 and its tests are done.

PR #454, opened 2019-01-08
COS-28 Add data model for notifications

DWRendell:
Sets up the data model for notifications, including responsible people, trigger rules answers etc etc. Will likely need some tweaking down the line but should hopefully be a good enough base to start developing features.


cosmetics-web/db/migrate/20190104094930_create_formula_files.rb
```diff
@@ -0,0 +1,9 @@
+class CreateFormulaFiles < ActiveRecord::Migration[5.2]
+  def change
+    create_table :formula_files do |t|
+      t.string :filepath
```
raamSoftwire: Why does this not require a filename in the same way as create_image_files below?

broder:
Would it be useful to add some controller routes for the notifications / responsible persons as well? They won't stay forever but the `GET` routes might be a good foundation for the poison centre stuff?

DWRendell:
I'm closing this pull request. I think the models in here will be useful for later tasks but I don't think it makes sense to merge them now as they're already being overwritten

PR #455, opened 2019-01-08
MSPSDS-859: Add organisation logic to visibility

kicferk1:


mspsds-web/app/models/investigation.rb
```diff
@@ -95,11 +95,12 @@ def pretty_visibility
...
-  def who_can_see
-    return [] unless is_private
-
-    # TODO MSPSDS-859: Replace hard-coded list with computation of users from organisations
-    [assignee, source&.user].map { |u| u&.id }.uniq
+  def can_be_seen_by_current_user
+    return true unless is_private
+    return true if (assignee&.organisation&.users&.include? current_user) || assignee == current_user
+    return true if (source&.user&.organisation&.users&.include? current_user) || source&.user == current_user
+    return true if current_user&.organisation&.name == "The Office of Product Safety"
```
broder: This feels a little fragile. Maybe we should allow "admin"s to see everything and make everyone in the office an admin?

mspsds-web/app/models/investigation.rb
```diff
@@ -95,11 +95,12 @@ def pretty_visibility
...
   end
 
-  def who_can_see
-    return [] unless is_private
-
-    # TODO MSPSDS-859: Replace hard-coded list with computation of users from organisations
-    [assignee, source&.user].map { |u| u&.id }.uniq
+  def can_be_seen_by_current_user
+    return true unless is_private
+    return true if (assignee&.organisation&.users&.include? current_user) || assignee == current_user
```
broder: `include?` is a method on Array (https://apidock.com/ruby/Array/include%3F) which makes me think that it'll loop through the whole array everytime. Maybe we should use `exists?` instead? https://github.com/zilkey/active_hash/blob/master/lib/active_hash/base.rb#L121

mspsds-web/app/models/investigation.rb
```diff
@@ -95,11 +95,12 @@ def pretty_visibility
     is_private ? ApplicationController.helpers.visibility_options[:private] : ApplicationController.helpers.visibility_options[:public]
   end
 
-  def who_can_see
-    return [] unless is_private
-
-    # TODO MSPSDS-859: Replace hard-coded list with computation of users from organisations
-    [assignee, source&.user].map { |u| u&.id }.uniq
+  def can_be_seen_by_current_user
```
broder: I wonder if we should define the method as `def visible_to(user)` and then pass in `current_user` from the policy. Might make it easier to unit test as well.

mspsds-web/app/models/investigation.rb
kicferk1: I can't get it to work. I'm pretty sure that definition of has_many in ActiveRecord is different to has_many in ActiveHash. 

The one from ActiveRecord adds the collection.exists? method to the class it's in, but the one from ActiveHash does not seem to do the same, so any has_many relation defined in an ActiveHash doesn't know what collection.exists? means.

mspsds-web/app/policies/investigation_policy.rb
```diff
@@ -1,12 +1,10 @@
...
   def destroy?
     @user.has_role? :admin
   end
 
   def show?
-    return true unless @record.is_private
-
-    # TODO MSPSDS-859: Replace users with organizations when we get organizations
-    @record.who_can_see.include?(@user.id)
+    @record.visible_to(current_user)
```
broder: Could we use `@user` here?

PR #457, opened 2019-01-08
MSPSDS-871: Non-office homepage

MatthewBurstein:
A couple of questions about this PR:
- Because the redirect for non-office users happens in the investigations controller, the root route is still /cases. Is there a better way to do this?
- Similarly the location and name of non_office_index.slim.html probably aren't great. Is there convention for where this should live and what it should be called?


mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -8,6 +8,11 @@ class InvestigationsController < ApplicationController
   # GET /cases.json
   # GET /cases.xlsx
   def index
+    # TODO delete unless block when user role exists for The Office
+    if !current_user.is_office?
+      render "non_office_index"
```
kicferk1: Does this render prevent the following lines from executing? In particular, do we query elasticsearch for investigations?

mspsds-web/app/models/user.rb
```diff
@@ -62,6 +62,10 @@ def has_role?(role)
     KeycloakClient.instance.has_role? role
   end
 
+  def is_office?
+    KeycloakClient.instance.has_role? :the_office
```
kicferk1: Do any of test users actually have this role? Do we need to add it in keycloak console?

broder: We'll need to add it to the `initial-setup.json` as well. Also, is something like `office_user`, `opss_user` or `opss_employee` a better name for the role?

broder: @kicferk1, should the role in https://github.com/UKGovernmentBEIS/beis-mspsds/pull/455/files#diff-4d9a28c3cbbb7360d78650adc7a5cc78R102 be the same as whatever we call the `office` role here?

kicferk1: I think so.

mspsds-web/app/controllers/investigations_controller.rb
MatthewBurstein: Good spot, I've added a return statement here.

kicferk1:
Sorry for all the conflicts, they are probably my fault :P 
Happy to pair up on fixing them, it ought to be pretty straightforward.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -8,6 +8,11 @@ class InvestigationsController < ApplicationController
   # GET /cases.json
   # GET /cases.xlsx
   def index
+    # TODO delete unless block when user role exists for The Office
+    if !current_user.is_office?
```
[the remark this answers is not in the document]
MatthewBurstein: Yes makes perfect sense :)

mspsds-web/app/controllers/homepage_controller.rb
```diff
@@ -0,0 +1,9 @@
+class HomepageController < ApplicationController
+  def show
+    if current_user.is_opss?
+      redirect_to investigations_path
+    elsif !current_user.is_opss?
```
broder: Is there a particular reason that this isn't just `else`?

MatthewBurstein: Nope 👍

mspsds-web/app/assets/stylesheets/navbar.scss
```diff
@@ -0,0 +1,65 @@
+// Taken from https://github.com/alphagov/govuk-design-system/blob/master/src/stylesheets/components/_navigation.scss
```
broder: Do we need this and the one in `shared-web`?

MatthewBurstein: 👍 I've added my changes to the file in shared-web

PR #461, opened 2019-01-10
MSPSDS-NONE: Restructure user in tests and add visibility testing

kicferk1:
Change hashes to instances of User and Organisation, add more users and fix everything that's broken by the previous two changes. 

It's quite likely this will cause merge conflicts with MSPSDS-871, and I'd rather fix them on this branch since it will probably be less complicated.


mspsds-web/test/test_helper.rb
```diff
@@ -105,21 +107,28 @@ def group_data
...
   end
 
+  def organisations_json
+    [
+      {id: "def4eef8-1a33-4322-8b8c-fc7fa95a2e3b", name: "Organisation 1", path: "/Organisations/Organisation 1"},
+      {id: "1a612aea-1d3d-47ee-8c3a-76b4448bb97b", name: "Organisation 2", path: "/Organisations/Organisation 2"},
+    ]
+  end
+
   def organisations
```
jasiekmiko: We could remove the duplication of data between `organisations` and `organisations_json` by letting `organisations` `map` the `Organisation` constructor over the `json` result

mspsds-web/test/models/investigation_test.rb
```diff
@@ -194,4 +194,43 @@ class InvestigationTest < ActiveSupport::TestCase
...
     assert_not_includes(Investigation.full_search(query).records.map(&:id), @investigation_with_business.id)
   end
+
+  test "visible to creator organisation" do
+    create_new_private_case
+    creator = User.find_by(last_name: "User_one")
+    creator.organisation = organisations[0]
+    user = User.find_by(last_name: "User_two")
+    user.organisation = organisations[0]
+    assert_equal(@new_investigation.visible_to(user), true)
```
jasiekmiko: This can just be `assert(@new_investigation.visible_to(user))` - ditto tests below!

mspsds-web/test/test_helper.rb
```diff
@@ -37,28 +37,20 @@ def setup
     self.class.import_into_elasticsearch
   end
 
-  def sign_in_as_user_with_organisation
+  def sign_in_as_user_with_organisation(is_admin: false, other_users: [admin_user, test_user_two, test_user_three], user: test_user)
```
jasiekmiko: It feels a bit counter-intuitive to pass in a list of all the users into a method called `sign_in_as_user...`. Would it help to split out mocking keycloak's users from mocking out the sing-in methods? Alternatively, could we fix the list of all users in the method, the way it was before?
Similarly not sure we need `...with_organisation` bit of the name if we're not passing in the organisation and it's set by default

mspsds-web/test/test_helper.rb
```diff
@@ -37,30 +37,29 @@ def setup
...
-  def sign_in_as_user_with_organisation(is_admin: false,
-                                        other_users: [admin_user, test_user(name: "User_two"), test_user(name: "User_three")],
-                                        user: test_user, is_opss: true)
+  def sign_in_as_user(is_admin: false, user: test_user, is_opss: true)
     group = organisations[0].id
     user = user
     user.organisation = organisations[0]
     user_groups = [{ id: user[:id], groups: [group] }].to_json
-
+    users = all_users.map { |u| u.last_name == user.last_name ? user : u }
```
jasiekmiko: This might warrant a comment - it's not immediately obvious why we're doing it.
Alternatively rather than passing a user object, passing an indicator of which user should be selected from `all_users` to act as the logged-in one? Then this line would perform a slightly more intuitive select.

mspsds-web/test/test_helper.rb
```diff
@@ -79,9 +79,10 @@ def assert_same_elements(expected, actual, msg = nil)
 
 private
 
-  def user_groups
-    [organisations[0][:id]]
-  end
+  #
+  # def user_groups
```
jasiekmiko: Is this meant to go?

mspsds-web/test/test_helper.rb
```diff
@@ -37,31 +37,29 @@ def setup
     self.class.import_into_elasticsearch
   end
 
-  def sign_in_as_user_with_organisation
-    user = test_user.merge(groups: user_groups)
-    user_groups_json = [{ id: user[:id], groups: user_groups }].to_json
-    stub_user_credentials(user: user, is_admin: false, is_opss: true)
-    stub_user_group_data(user_groups: user_groups_json)
-    stub_user_data(users: [admin_user, user])
+  def sign_in_as_user(is_admin: false, user: test_user, is_opss: true)
```
jasiekmiko: Is "being opss" the same thing as being part of the "opss" organisation, or are they unrelated concepts?

kicferk1: I think in our app being part of keycloak organisation "Office for Product Safety and Standards" gives keycloak user a role "opss_user", which is then surfaced in user model by asking keycloak client about this particular role. 

So in a way yes, it is the same. But in tests we intercept this flow earlier by stubbing the part of keycloak that answers to has_role question, so in tests they are not really linked together.

jasiekmiko: In that case, we should probably link it - otherwise someone will come along to use this later on and be surprised that `user.organisation` is not `OPSS` despite them passing `is_opss: true`.
Can we replace the `is_opss` flag parameter with the organisation parameter which would
a) set the organisation
b) stub the `:has_role?` accordingly
?

PR #462, opened 2019-01-10
MSPSDS-594 welcome email

jasiekmiko:
This is pending final email content, but it's not a code change (email templates are held in Verify)


keycloak/Dockerfile
```diff
@@ -36,7 +36,7 @@ WORKDIR /tmp/keycloak
 COPY ./providers/email ./email
 RUN mkdir -p ./providers
 
-RUN mvn -q --settings ./email/settings.xml --file ./email/pom.xml package
+RUN mvn -q --file ./email/pom.xml package
```
jasiekmiko: A similar simplification can be done for the sms-authentication package, but wanted to get feedback on it first :)

keycloak/configuration/standalone-ha.xml
```diff
@@ -629,7 +629,8 @@
                 <provider name="default" enabled="false"/>
                 <provider name="notify-email" enabled="true">
                     <properties>
-                        <property name="verifyEmailTemplateId" value="edab6550-eb34-4cb1-910f-41606e583076"/>
+                        <property name="verifyEmailTemplateId" value="0be2ef83-bc2d-4b71-8fc2-93dc1eea0ae8"/>
```
jasiekmiko: I'm not very clear on the difference between standalone-ha and standalone. @tom-bell-softwire are you able to add any context to what their roles are?

tom-bell-softwire: The `standalone-ha.xml` configuration is used when running in [standalone clustered mode](https://www.keycloak.org/docs/latest/server_installation/index.html#_standalone-ha-mode). We aren't running in this configuration at the moment, but may want to in Prod at some point. It's a little bit of extra effort to maintain both, so we could remove it, but perhaps worth keeping both configs up to date?

keycloak/Dockerfile
tom-bell-softwire: I'm happy with moving the repo settings to the POM itself (as recommended [here](https://stackoverflow.com/questions/2225535/where-is-the-best-place-to-specify-maven-repositories-pom-xml-or-settings-xml)).

jasiekmiko: Great! I've opened a PR (https://github.com/UKGovernmentBEIS/keycloak-sms-authenticator-sns/pull/6) for it, will merge it with mspsds as part of this pull once approved

PR #463, opened 2019-01-10
MSPSDS-661: Restrict personal details gdpr

kicferk1:
We store personal details in 2 places: reporter details(when they are a consumer) and correspondence(when it is with a consumer). 

We surface them in 3 places: activity timeline, search highlights and excel exports. 

Design details for what exactly the view part of that should look like are still being developed, but Nick is working on them atm so they shouldn't take long. 

The only non-straightforward part is dealing with highlights for correspondence. Elasticsearch highlights don't seem to come with something like source id(only source name and key name), so finding the source record to check its sensitivity is hard. For reporter it's doable since we do get investigation to which it is coupled and reporter is unique for investigation. 

But correspondence is not unique :( I opted for restricting a given correspondence highlight if any of sensitive correspondences contains that highlight in relevant field, which is slightly too trigger happy. Maybe we should have 2 lists of correspondence on investigation, one sensitive and the other non-sensitive, so we could fine tune it more. It still would not be perfect, but it would mean no non-sensitive cases get restricted. 

Or maybe there is a better solution, some hidden elasticsearch option that allows for more defining highlights, but cursory google search and a read through elasticsearch docs doesn't make me optimistic :(


mspsds-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -33,9 +33,41 @@ def replace_unsightly_field_names(field_name)
...
+    false
+  end
+
+  def correspondence_should_be_hidden(highlight, investigation)
+    return false unless highlight[0].include? "correspondences"
+
+    key = highlight[0].partition('.').last
+    highlighted_texts = highlight[1]
+    sanitized_content = sanitize(highlighted_texts.first)
+    investigation.correspondences.each do |c|
```
broder: Would https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#specify-highlight-query help here? Would we be able to do something like "only return the highlight if it's not sensitive"? That does mean that even allowed users couldn't see the highlight though...

kicferk1: I think making elasticsearch highlight query approach work without some horrible hacks would require us to split correspondences into two lists anyway, so that we have a single field we can cut out from highlights that corresponds to sensitive correspondences. 

Since having two correspondence lists helps regardless of the approach, should I implement that first?

mspsds-web/app/views/investigations/activities/_activity_card.html.slim
```diff
@@ -0,0 +1,28 @@
...
+  - if activity.attached_image?
+    .govuk-grid-row
+      .govuk-grid-column-one-third
+        = render "documents/document_preview", document: activity.attachment, dimensions: "100x100",
+                class_name: "product-image"
+      .govuk-grid-column-two-thirds
+        = markdown activity.body if activity.body
+  - else
+    p
+      - if activity.sensitive_body?
```
broder: This isn't quite the same as the design prototype that Nick demoed earlier - is that alright?

kicferk1: I'll double check, but I talked about it with Nick and Ed before I started working on it and they told me if it's troublesome to separate sensitive stuff inside a markdown then hiding all of it is ok.

mspsds-web/app/helpers/investigations/display_text_helper.rb
kicferk1: After a chat with Ed and Nick we decided that if a sensitive field gives us a highlight and that highlight is exactly contained in the same field of a different object(so, say, the person who is being talked to is Michal, private in one case and public in another case), then it's ok to display the highlight. 

Nick said current implementation works ok from design standpoint.

mspsds-web/app/views/investigations/_highlight_card.html.slim
```diff
@@ -4,8 +4,8 @@
       = "#{case_question_text(investigation).capitalize} #{investigation.pretty_id}"
     span = link_to investigation.title, investigation, class: "govuk-!-font-weight-bold"
   .govuk-grid-column-one-half
-    - get_displayable_highlights(highlights).each do |highlight|
+    - get_displayable_highlights(highlights, investigation).each do |highlight|
```
broder: Could we store the result of `get_displayable_highlights` in a variable? It's probably not great that we're calling it twice in quick succession.

mspsds-web/app/models/audit_activity/investigation/add.rb
```diff
@@ -16,4 +16,17 @@ def self.build_reporter_details(reporter)
     details += "<br>#{self.sanitize_text reporter.other_details}" if reporter.other_details.present?
     details
   end
+
+  def sensitive_body?
+    !investigation.reporter&.can_be_displayed?
+  end
+
+  def safe_body
+    safe_part = body.split("<br><br>**Reporter**<br><br>")[0]
```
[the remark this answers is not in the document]
kicferk1: Yeah, it's a bit annoying. If there were more things after reporter it would require a lot of messy string work. 
Then again, it lives close enough to the thing actually creating the usual markdown that it can be easily coupled.

Alternatively, we discussed changing activity body format to JSON if we need granularity, and I guess this might well be the point at which we do it.

PR #468, opened 2019-01-11
COS-41: Remove ES6 from cosmetics

broder:
And other various fixes.

Going commit-by-commit might be easiest.

The dependabot config has been added recently and is defined [here](https://dependabot.com/blog/introducing-config-files/).

### DIFF
--- a/mspsds-web/app/views/application/_documents_list_summary.html.slim
+++ b/mspsds-web/app/views/application/_documents_list_summary.html.slim
@@ -9,8 +9,9 @@
         br
         p = document.metadata[:description]
         p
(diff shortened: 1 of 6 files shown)


mspsds-web/app/views/application/_documents_list_summary.html.slim
```diff
@@ -9,8 +9,9 @@
         br
         p = document.metadata[:description]
         p
-          = link_to "View #{pretty_type_description(document)}",
-                  document, class: "govuk-link mspsds-block-link", target: '_blank'
+          - if document.metadata["safe"]
```
tom-bell-softwire: Good catch!

PR #472, opened 2019-01-14
COSBETA-8 Set up manual journey to create cosmetics notification

DWRendell:
Currently creates notifications with two fields, `product_name` and `external_reference`. The second one I just added to test the wizard with multiple steps, it can be removed once more fields are added.

Not included: UX for when the user tries to progress to the next step without filling in a field. I'll talk to Ralph but I'm guessing we'll want to disable the button when the input is empty, and display an error message if a non JS user presses the button.


DWRendell:
Updated now with:
- Updated routes of the form `/notification/:id/manual_entry/:step`
- Error messages displayed to user if they try to proceed without filling in each step
- Confirmation page moved out of the wizard into a separate route
- Storing the state as an enum in the hopes of improving the validation logic

cosmetics-web/app/controllers/manual_entry_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    if step == steps.last
+      @notification.submit_notification!
+    end
+
+    render_wizard @notification
+  end
+
+  def create
+    @notification = Notification.create
+    session[:notification_id] = @notification.id
```
broder: We probably don't need this anymore if we're getting the ID from the URL?

cosmetics-web/app/controllers/manual_entry_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def finish_wizard_path
+    @notification = Notification.find(params[:notification_id])
+    notification_path(@notification) + '/confirmation'
+  end
+
+private
+
+  def notification_params
+    params.permit(notification: %i[
```
broder: I think this is generally formatted as `params.require(:notification).permit(:product_name, :external_reference)` (see the MSPSDS controllers)

cosmetics-web/app/controllers/manual_entry_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  def show
+    @notification = Notification.find(params[:notification_id])
+    render_wizard
+  end
+
+  def update
+    @notification = Notification.find(params[:notification_id])
+
+    if !notification_params.nil?
+      @notification.update(notification_params)
```
broder: I think this means that if we're being malicious, we can update all the parameters, on any step.

Also, do we need to validate the params before we try to save them? On MSPSDS, we often do things like
```
@foo.bar = 'foo'
if @foo.save?
  redirect_to next_wizard_path
else
  render step
end
```

Is that what `render_wizard` is doing?

cosmetics-web/app/assets/javascripts/manual_entry.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
broder: We probably don't need these empty files for now.

cosmetics-web/app/models/validators/manual_notification_validator.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+      elsif !new_value.nil?
+        notification.errors.add changed_attribute, "cannot be set at this stage"
+      end
+    end
+  end
+
+private
+
+  MANDATORY_ATTRITBUTES_BY_STATE = [
```
broder: Does this need to be in the same order as the `state` enum? If so, it's still not totally obvious.

cosmetics-web/app/views/manual_entry/add_external_reference.html.slim
```diff
@@ -0,0 +1,17 @@
...
+- external_reference_errors = @notification.errors.messages[:external_reference].any?
+.govuk-grid-row
+    .govuk-grid-column-two-thirds
+        h1.govuk-heading-xl
+            | External reference number
+        - if external_reference_errors
+            span.govuk-error-message External reference number is a required field
+        = form_for(@notification, url: wizard_path, method: :put) do |form|
+            .govuk-form-group
+                = form.label "Dummy field for testing wizard", class: "govuk-label"
```
broder: We should probably use `form.label :external_reference, "Dummy...` so that we get the correct `<label for="">` tag generated. https://apidock.com/rails/ActionView/Helpers/FormHelper/label

cosmetics-web/config/routes.rb
```diff
@@ -5,4 +5,12 @@
 
   get '/send' => 'helloworld#send_email'
   post '/' => 'helloworld#upload_file'
+
+  resources :manual_entry, only: %i[show create update]
```
broder: Do we need the controller mapped to two routes?

cosmetics-web/spec/controllers/manual_entry_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it "assigns the correct notification" do
+      notification = Notification.create
+      get(:confirmation, params: { 'notification_id' => notification.id })
+      expect(assigns(:notification)).to eq(notification)
+    end
+  end
+
+  private
+
+  def get_manual_journey_url(notification_id, step)
```
broder: Would the rails URL helpers work here? It'll look something like `notification_manual_entry_path(notification_id, step)` https://guides.rubyonrails.org/routing.html#path-and-url-helpers

cosmetics-web/app/controllers/manual_entry_controller.rb
DWRendell: True, removed.

cosmetics-web/app/controllers/manual_entry_controller.rb
DWRendell: This looks better, it means a slight rework of the update action or we get errors on the check your answers page where the notification param isn't set.

cosmetics-web/app/controllers/manual_entry_controller.rb
DWRendell: https://github.com/schneems/wicked#How
> We're passing render_wizard our @user object here. If you pass an object into render_wizard it will show the next step if the object saves or re-render the previous view if it does not save.

Sounds like it's just a shortcut that does the same thing. 

I'm fairly sure that the validator will prevent you from updating any attributes you wouldn't be able to by just navigating back to the page where that attribute is set, but maybe we should talk it over.

cosmetics-web/app/assets/javascripts/manual_entry.js
DWRendell: Agreed, deleting `manual_entry.scss`, `manual_entry.js`, `manual_entry_helper.rb`

cosmetics-web/app/models/validators/manual_notification_validator.rb
DWRendell: Yes, I agree it's a bit ambiguous. Ideally I wanted to use a map from the state names to attributes, but I could only make it work by accessing the attributes using the int form of the enum as the key.

cosmetics-web/app/views/manual_entry/add_external_reference.html.slim
DWRendell: 👍 Added for this and product_name

cosmetics-web/config/routes.rb
DWRendell: Oops, no, deleted

cosmetics-web/spec/controllers/manual_entry_controller_spec.rb
DWRendell: Ah thanks, that's really helpful

cosmetics-web/app/views/manual_entry/confirmation.html.slim
```diff
@@ -0,0 +1,15 @@
...
+            .govuk-panel__body
+                | Your cosmetic notification reference number
+                br
+                strong -
+        p We have sent confirmation of this to you
+        h3.govuk-heading-m What happens next
+        p
+            | Your notification has been added to the cosmetic records.
+            | Ingredient information can only be accessed by poison centres.
+        a href="/" Back to notification overview.
```
tom-bell-softwire: We should probably be using the Rails `link_to` helper method for links whenever possible.

cosmetics-web/app/controllers/manual_entry_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    redirect_to wizard_path(steps.first, notification_id: @notification.id)
+  end
+
+  def confirmation
+    @notification = Notification.find(params[:notification_id])
+  end
+
+  def finish_wizard_path
+    @notification = Notification.find(params[:notification_id])
+    notification_path(@notification) + '/confirmation'
```
tom-bell-softwire: Would be good to use the Rails path helper method here.

cosmetics-web/app/controllers/manual_entry_controller.rb
```diff
@@ -0,0 +1,48 @@
+class ManualEntryController < ApplicationController
+  include Wicked::Wizard
+
+  steps :add_product_name, :add_external_reference, :check_your_answers
+
+  def show
+    @notification = Notification.find(params[:notification_id])
```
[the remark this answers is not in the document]
DWRendell: Sounds good 👍

cosmetics-web/app/views/manual_entry/add_external_reference.html.slim
```diff
@@ -0,0 +1,17 @@
+- content_for :page_title, "External reference number"
```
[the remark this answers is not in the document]
DWRendell: Hmm, not sure, I'll double check this

cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+RSpec.describe Notification, type: :model do
+  describe "updating product_name" do
+    it "transitions state from empty to product_name_added" do
+      notification = Notification.create
+
+      notification.product_name = 'Super Shampoo'
+      notification.save
+
+      expect(notification.state).to eq('product_name_added')
```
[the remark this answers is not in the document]
DWRendell: That causes the test to fail, I think because we're just getting the value the state is saved in the database as here, which is just a string.

cosmetics-web/app/views/manual_entry/add_product_name.html.slim
```diff
@@ -0,0 +1,19 @@
+- content_for :page_title, "Cosmetic name"
+- content_for :after_header do
+  = link_to "Back", "/", class: "govuk-back-link"
+- product_name_errors = @notification.errors.messages[:product_name].any?
+.govuk-grid-row
+    .govuk-grid-column-two-thirds
+        h1.govuk-heading-xl
+            | Cosmetic name
+        - if product_name_errors
+            span.govuk-error-message Cosmetic name is a required field
```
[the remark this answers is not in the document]
DWRendell: I think the problem with moving the errors messages there is that we'd need a mapping from the attribute names e.g. `product_name` to 'Cosmetic name'. It would be much nicer though, I'll have a look at how to do it.

I'll talk to Ralph about the styling

cosmetics-web/spec/controllers/manual_entry_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.describe ManualEntryController, type: :controller do
+  before do
+    authenticate_user
```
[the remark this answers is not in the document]
DWRendell: Good point, done

cosmetics-web/app/views/manual_entry/confirmation.html.slim
DWRendell: Yep, done

cosmetics-web/spec/models/notification_spec.rb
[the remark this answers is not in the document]
tom-bell-softwire: Ah ok, thanks.

We might want to make use the [helper methods](https://api.rubyonrails.org/v5.2.2/classes/ActiveRecord/Enum.html) that ActiveRecord provides for enum properties, e.g. `notification.product_name_added?`, though it's probably worth being more explicit in a test assertion, so not worth changing here.

cosmetics-web/app/views/notifications/edit.html.slim
```diff
@@ -48,6 +48,5 @@
         p
             | By submitting this notification you are confirming that, to the best of your knowledge,
             | the details you are providing are correct.
-        = form_for(@notification, url: wizard_path, method: :put) do |form|
-            .govuk-form-group
-                = form.submit "Accept and notify product", class: "govuk-button"
+        a.govuk-button[href=confirmation_notification_path(@notification)]
```
tom-bell-softwire: From an accessibility point of view, I wonder if this should be a button rather than a link. This would mean either using a form with a submit element (as you had before) or adding `role="button"` to the link so that screen readers treat it appropriately.

I think the form option is probably slightly neater, but either would be fine.

If you opt to keep it as a link instead, I think it's slightly more idiomatic to have something like:

```ruby
= link_to "Accept and notify product", confirmation_notification_path(@notification), class: "govuk-button", role: "button"
```

DWRendell: Sounds like a good idea, done

PR #476, opened 2019-01-15
MSPSDS-863 Update govuk-theme submodule with MSPSDS-863

jasiekmiko:
PR: https://github.com/UKGovernmentBEIS/keycloak-theme-govuk/pull/5


jasiekmiko:
> In case you weren't aware, [Dependabot would also have picked this up overnight](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/.dependabot/config.yml#L4).

Duh! Good reminder :)

PR #483, opened 2019-01-16
(No card) Fix error building Docker image for Cosmetics Web

DWRendell:
I have no idea why, but I started getting the following error when building cosmetics:
```
Fetched 698 kB in 2s (340 kB/s)
Reading package lists...
W: GPG error: https://dl.yarnpkg.com/debian stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 4F77679369475BAA
W: The repository 'https://dl.yarnpkg.com/debian stable InRelease' is not signed.
Reading package lists...
Building dependency tree...
Reading state information...
The following additional packages will be installed:
  dpkg-dev fakeroot libalgorithm-diff-perl libalgorithm-diff-xs-perl
  libalgorithm-merge-perl libfakeroot
Suggested packages:
  debian-keyring
The following NEW packages will be installed:
  build-essential dpkg-dev fakeroot libalgorithm-diff-perl
  libalgorithm-diff-xs-perl libalgorithm-merge-perl libfakeroot nodejs yarn
0 upgraded, 9 newly installed, 0 to remove and 17 not upgraded.
Need to get 16.2 MB of archives.
After this operation, 71.8 MB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  yarn
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
ERROR: Service 'cosmetics-web' failed to build: The command '/bin/sh -c apt-get update && apt-get install -y  build-essential   nodejs   yarn' returned a non-zero code: 100
```
Adding this space before the backslash fixed it, and it does match this step in MSPSDS now. 

I think it might be something to so with the key expiring over the new year: https://github.com/yarnpkg/yarn/issues/4453 If anyone has any ideas why this fixed it please let me know.


DWRendell:
> I'm guessing this fixed it because it stopped that line from using the cached image on your machine for that line. I think doing `docker-compose build --no-cache` would have the same effect?
> 
> Still happy to merge this if you want to though.

That makes sense. Might as well, it does make the formatting match MSPSDS at least

jasiekmiko:
Can confirm `docker-compose build --no-cache` was how we fixed this with Raam last week.

PR #486, opened 2019-01-17
MSPSDS-897: Reduce number of queries

kicferk1:
This solves the n+1 problems on case list page, case page and excel export sheet. 

Loading attachments on activities proved tricky, since not all of them have attachments, and not all attachments are on the same key, and so on. The current solution uses Preloader to only attempt to load attachments for activities that have them.


mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -155,4 +173,18 @@ def respond_to_invalid_data(origin)
       format.json { render json: @investigation.errors, status: :unprocessable_entity }
     end
   end
+
+  def preload_activities
+    @activities = @investigation.activities.eager_load(:source)
+    preload_manually(@activities.select { |a| a.respond_to?("attachment") },
```
broder: Should we do this for correspondence too?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -111,6 +120,15 @@ def create
 
 private
 
+  def eager_load_investigation_data
```
broder: Should we call this `set_investigation_with_associations` or something like that?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -155,4 +173,20 @@ def respond_to_invalid_data(origin)
...
+                     [{ attachment_attachment: :blob }])
+    preload_manually(@activities.select { |a| a.respond_to?("email_file") },
+                     [{ email_file_attachment: :blob }, { email_attachment_attachment: :blob }])
+    preload_manually(@activities.select { |a| a.respond_to?("transcript") },
+                     [{ transcript_attachment: :blob }, { related_attachment_attachment: :blob }])
+    preload_manually(@activities.select { |a| a.respond_to?("correspondence") },
+                     [:correspondence])
+  end
+
+  def preload_manually(records, associations)
```
broder: Is this the best place for this method? Would we want to use it for any other models?

(Sorry, I think my internet died when adding the comment the first time so it wasn't included in the review...)

kicferk1: Not sure if we ever will want to use it elsewhere, but it does seem like it doesn't fit into investigation_controller.

PR #489, opened 2019-01-17
COSBETA-22: Upload a single file

aidayounesi:


cosmetics-web/Gemfile
```diff
@@ -26,3 +26,8 @@ group :development, :test do
 
   gem "shared-web-dev", path: "./vendor/shared-web/dev"
 end
+
+group :test do
+  gem "capybara"
```
broder: Should we add these to the `shared-web-dev.gemspec` given that they're needed for both?

cosmetics-web/app/views/products/edit.html.slim
```diff
@@ -0,0 +1,8 @@
+h1 Editing product
+
+== render 'form'
```
broder: What do `==`, `=>` and `=<` do in slim?

cosmetics-web/app/helpers/products_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ProductsHelper
```
tom-bell-softwire: Probably worth deleting this, too.

cosmetics-web/app/views/products/show.html.slim
```diff
@@ -0,0 +1,17 @@
+p#notice = notice
+
+p
+  strong Name:
+  = @product.name
+
+  - if @product.name.present?
```
tom-bell-softwire: A better test here would probably be `@product.uploaded_file.attached?`

cosmetics-web/spec/controllers/products_controller_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'rails_helper'
+
+# This spec was generated by rspec-rails when you ran the scaffold generator.
```
tom-bell-softwire: It would be good to remove a lot of these boilerplate comments, since they distract from the test cases.

cosmetics-web/spec/rails_helper.rb
```diff
@@ -58,4 +91,26 @@
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  # config.before(:suite) do
```
tom-bell-softwire: Probably worth removing these commented lines about DatabaseCleaner. If we decide to use it, we can add them then.

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  before do
+    allow(Keycloak::Client).to receive(:user_signed_in?).and_return(true)
+  end
+
+  after do
+    allow(Keycloak::Client).to receive(:user_signed_in?).and_call_original
+  end
+
+  it "enables to upload a file" do
+    visit "/products/new"
```
tom-bell-softwire: It's generally better to use the path helper functions, rather than hard-code these paths. That way, we won't need to find and replace the paths if we ever restructure or rename the routes.

In this case, I think `visit new_product_path` would work.

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    page.attach_file('uploaded_file', Rails.root + 'spec/fixtures/testImage.png')
```
tom-bell-softwire: To match convention, I think the test files should probably live under `spec/fixtures/files/`

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    click_button "Upload"
+    
+    expect(page).to have_text("testImage.png")
```
tom-bell-softwire: Do we want to test that a link is present with that name?

cosmetics-web/app/views/products/show.html.slim
```diff
@@ -0,0 +1,17 @@
...
+    p
+      ' Currently selected file:
+      = link_to @product.uploaded_file.filename, url_for(@product.uploaded_file), target: "_blank"
+  -else
+    P
```
tom-bell-softwire: P -> p

cosmetics-web/app/views/products/show.html.slim
```diff
@@ -0,0 +1,17 @@
...
+      ' No attachment
```
tom-bell-softwire: I think you can probably use `|` instead of `'` here (the latter [preserves trailing whitespace](https://www.rubydoc.info/gems/slim/frames#Verbatim_text_with_trailing_white_space__)).

cosmetics-web/app/views/products/edit.html.slim
aidayounesi: == Same as the single equals sign (=), but does not go through the escape_html method. 
==> Same as the double equals sign (==), except that it adds a trailing white space.
==< Same as the double equals sign (==), except that it adds a leading white space.

https://www.rubydoc.info/gems/slim/frames#Verbatim_text_with_trailing_white_space__

I am not quite sure why == instead of = when rendering a form! It can be a potential for XSS, am I right?

cosmetics-web/Gemfile
aidayounesi: Sure

cosmetics-web/app/assets/javascripts/products.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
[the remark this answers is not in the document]
aidayounesi: Sure

cosmetics-web/app/assets/stylesheets/products.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Products controller here.
```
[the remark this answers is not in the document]
aidayounesi: Sure

cosmetics-web/app/helpers/products_helper.rb
aidayounesi: Sure, thanks

cosmetics-web/app/views/products/show.html.slim
aidayounesi: Right, thanks

cosmetics-web/spec/controllers/products_controller_spec.rb
aidayounesi: Sure

cosmetics-web/spec/rails_helper.rb
aidayounesi: Sure

cosmetics-web/spec/system/upload_single_file_spec.rb
aidayounesi: Sure

cosmetics-web/spec/system/upload_single_file_spec.rb
aidayounesi: Sure

cosmetics-web/app/views/products/show.html.slim
aidayounesi: Thanks

cosmetics-web/app/views/products/show.html.slim
aidayounesi: Sure

cosmetics-web/spec/system/upload_single_file_spec.rb
aidayounesi: For now probably yes

cosmetics-web/app/views/products/edit.html.slim
aidayounesi: Used = instead of ==

cosmetics-web/app/views/products/_form.html.slim
```diff
@@ -0,0 +1,12 @@
+= form_for @product do |f|
+  - if @product.errors.any?
+    #error_explanation
+      h2 = '#{pluralize(@product.errors.count, "error")} prohibited this product from being saved:'
```
tom-bell-softwire: I think we've followed the opposite convention for strings elsewhere – using `"` to wrap the main string and `'` for any string arguments for functions that appear within it.

cosmetics-web/spec/controllers/products_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      get :new, params: {}, session: valid_session
+      expect(response).to be_successful
+    end
+  end
+
+  describe "POST #create" do
+    context "with valid params" do
+      it "creates a new Product" do
+        expect {
+          post :create, params: { product: valid_attributes }, session: valid_session
```
tom-bell-softwire: I think we can probably just get rid of the `session: valid_session` argument in these tests, since it isn't really adding anything.

cosmetics-web/app/views/products/show.html.slim
```diff
@@ -4,7 +4,7 @@ p
   strong Name:
   = @product.name
 
-  - if @product.uploaded_file.attached?
+  - if @product.uploaded_file.attached? && @product.uploaded_file.metadata["safe"]
```
tom-bell-softwire: Will this show links at all until the antivirus functionality has been merged?

If not, I'd suggest making this change as part of that story instead.

aidayounesi: Sure, I will reverse this change

cosmetics-web/spec/controllers/products_controller_spec.rb
aidayounesi: Sure, thanks

cosmetics-web/app/views/products/_form.html.slim
aidayounesi: Sure, thanks

tom-bell-softwire:
I think the Travis build might be failing because some of the gem versions have been inadvertently updated in the `Gemfile.lock` files. Could you try reverting the version changes in the two `Gemfile.lock` files to see if that fixes the problem?

PR #495, opened 2019-01-18
MSPSDS-873: Create project

kicferk1:
question_type was not actually entered anywhere anymore, so I removed it. 
I changed is_case(bool) into case_type(string)
I renamed question_title to user_title, because projects can have user-given titles too.


broder:
Did you think about whether we could do this with single table inheritance? (i.e. a model for Cases, Questions and Projects, all extending from a base class)

kicferk1:
I haven't, but it does sound reasonable. I'll get to it.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -107,10 +109,10 @@ def visibility
   # POST /cases
   # POST /cases.json
   def create
```
broder: Is this used anymore?

mspsds-web/app/helpers/investigations/display_text_helper.rb
```diff
@@ -1,10 +1,6 @@
 module Investigations::DisplayTextHelper
-  def case_question_text(investigation)
-    investigation.is_case ? 'case' : 'question'
-  end
-
   def report_question_text(investigation)
```
broder: Should this move to the new models?

mspsds-web/app/models/investigation.rb
```diff
@@ -110,15 +105,11 @@ def pretty_id
...
-
-  def question_title_prefix
-    question_type && !is_case ? question_type + ' ' : ''
+    "#{case_type.titleize} #{pretty_id}"
   end
 
-  def title
-    self.is_case ? case_title : question_title
+  def pretty_presentation
+    "#{case_type.titleize}: #{pretty_id}"
```
broder: Is it intentional that this method and the one above are so similar? Might be worth asking a designer

mspsds-web/app/models/investigation/allegation.rb
```diff
@@ -0,0 +1,48 @@
+class Investigation::Allegation < Investigation
+  validates :description, presence: true, on: %i[allegation_details]
+  validates :hazard_type, presence: true, on: :allegation_details
+  validates :product_type, presence: true, on: :allegation_details
+
+  index_name [Rails.env, "investigations"].join("_")
+
+  def self.model_name
+    Investigation.model_name
```
broder: Was this the cause of one of the elasticsearch issues?

Also, would `super.model_name` work instead? Maybe it's not as obvious though...

mspsds-web/app/views/investigations/_case_card.html.slim
```diff
@@ -1,7 +1,7 @@
 .govuk-grid-row class=(investigation.is_closed ? "mspsds-case-card mspsds-case-card--closed" : "mspsds-case-card")
   .govuk-grid-column-one-half
     span.govuk-caption-m[class="govuk-!-font-size-16"]
-      = "#{case_question_text(investigation).capitalize} #{investigation.pretty_id}"
+      = "#{investigation.case_type.capitalize} #{investigation.pretty_id}"
```
broder: Is this equivalent to `= investigation.pretty_description.capitalize`?

mspsds-web/app/views/investigations/project/new.html.slim
```diff
@@ -0,0 +1,28 @@
+- content_for :page_title, "New project"
+- content_for :after_header do
+  = link_to "Back to Cases", investigations_path, class: "govuk-back-link"
+h1.govuk-heading-xl
+  = "New Project"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with(model: @investigation, url: project_index_path, method: :post,) do |form|
```
broder: Is the trailing comma on the `form_with` args intentional?

mspsds-web/db/schema.rb
```diff
@@ -124,13 +124,13 @@
     t.datetime "created_at", null: false
     t.text "description"
     t.string "hazard_type"
-    t.boolean "is_case", default: true, null: false
+    t.string "investigations", default: "Investigation::Allegation"
```
broder: Do we need this column?

mspsds-web/test/test_helper.rb
```diff
@@ -26,7 +26,7 @@ def self.import_into_elasticsearch
     unless @models_imported
       ActiveRecord::Base.descendants.each do |model|
         if model.respond_to?(:__elasticsearch__)
```
broder: Does `model.respond_to?(:__elasticsearch__) && !model.super.respond_to?(:__elasticsearch__)` work? Is that what we want here?

mspsds-web/app/helpers/investigations/display_text_helper.rb
kicferk1: Turns out it is not actually used since the restructuring that question/allegation flows got a while ago.

mspsds-web/app/models/investigation.rb
kicferk1: I will, but it probably is intentional, one shows in navigation(back to case xxx-xxx) and the other in headers when adding activity to a case(case: xxx-xxx). 

Also it's quite likely I won't get an answer today :P

kicferk1: Ed(design Ed) says 'it should probably be the same, let's go with colon, preference over no colon is 55/45'.

mspsds-web/test/test_helper.rb
kicferk1: Yes, I think that's what we want. Also I think we need to include that in production environment.

PR #496, opened 2019-01-18
MSPSDS-821 speed up time to staging

jasiekmiko:
@broder when we've discussed this in a couple weeks back, I recall our conclusion was to disable tests on PR builds coming from master.
I'm no longer sure about that logic and have opted for disabling tests on non-pr builds, with the following logic.

1. We want to avoid an untested version of the code to be released. Such a commit on say staging could come from
  A) a PR from master, 
  B) a PR from feature/hotfix branch, 
  C) an admin updating the master branch with staging/prod merge to enable PR from master to staging/prod (when doing staging/prod release).
2. In cases A + B, the PR will have run the same version of the code and tested it. We've got checks that ensure this.
3. In case C, 99% of the time the commit will be empty, since master is practically ahead of the staging/prod branch except for the merge commit form the last branch. The only time when that's not true is when we've deployed a hotfix to staging/prod. In this case
  i) The hotifx code will have been tested with it's own PR, so we only need to worry about the combination of hotfix + changes on master
ii) When resorting to a hotfix, the dev should also be merging it into master as well as staging/prod
iii) Even if that's not followed, the most we break is INT - the code won't be deployed further to staging, as it will get flagged up in the staging PR.

Happy to be proven wrong here if there's a hole in my thinking!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 stages:
   - Build
   - name: Test
-    if: type = pull_request # This is designed to cut out unnecessary runs of the test suit. See MSPSDS-821 for details.
+    if: type = pull_request # This is designed to cut out unnecessary runs of the test suite. See MSPSDS-821 for details.
   - name: Deploy to int
     if: branch = master AND type != pull_request
   - name: Deploy to staging
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ cache:
 stages:
   - Build
   - name: Test
-    if: type = pull_request
+    if: type = pull_request # This is designed to cut out unnecessary runs of the test suit. See MSPSDS-821 for details.
   - name: Deploy to int
     if: branch = master AND type != pull_request
   - name: Deploy to staging
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ cache:
 
 stages:
   - Build
-  - Test
+  - name: Test
+    if: type = pull_request
   - name: Deploy to int
     if: branch = master AND type != pull_request
   - name: Deploy to staging
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,16 @@ cache:
   - "~/bundle-cache"
   - "~/npm-cache"
 
+stages:
+  - Build
+  - Test
+  - name: Deploy to int
+    if: branch = master AND type != pull_request
+  - name: Deploy to staging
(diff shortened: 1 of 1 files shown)


broder:
Should we account for people pushing directly to master without going through a PR?

.travis.yml
```diff
@@ -3,96 +3,92 @@ sudo: required
...
-    if: branch = master AND type != pull_request
-    script: BRANCH=${TRAVIS_BRANCH} BUILD_ID=${TRAVIS_BUILD_NUMBER} SPACE=int COMPONENT=mspsds-worker ./ci/deploy-if-changed.sh
-  - name: Deploy keycloak
-    if: branch = master AND type != pull_request
-    script: BRANCH=${TRAVIS_BRANCH} BUILD_ID=${TRAVIS_BUILD_NUMBER} SPACE=int COMPONENT=keycloak ./ci/deploy-if-changed.sh
-  - name: Deploy Cosmetics web
+stages:
+  - Build
+  - name: Test
+    if: type = pull_request
```
jasiekmiko:
```suggestion
    if: type = pull_request # This is designed to cut out unnecessary runs of the test suit. See MSPSDS-821 for details.
```

broder: It should probably be "test suite" but otherwise, sounds good!

jasiekmiko: Ooops! haha 😂

.travis.yml
```diff
@@ -3,96 +3,92 @@ sudo: required
...
-    if: branch = master AND type != pull_request
-    script: BRANCH=${TRAVIS_BRANCH} BUILD_ID=${TRAVIS_BUILD_NUMBER} SPACE=int COMPONENT=mspsds-worker ./ci/deploy-if-changed.sh
-  - name: Deploy keycloak
-    if: branch = master AND type != pull_request
-    script: BRANCH=${TRAVIS_BRANCH} BUILD_ID=${TRAVIS_BUILD_NUMBER} SPACE=int COMPONENT=keycloak ./ci/deploy-if-changed.sh
-  - name: Deploy Cosmetics web
+stages:
+  - Build
+  - name: Test
+    if: type = pull_request # This is designed to cut out unnecessary runs of the test suit. See MSPSDS-821 for details.
```
jasiekmiko:
```suggestion
    if: type = pull_request # This is designed to cut out unnecessary runs of the test suite. See MSPSDS-821 for details.
```

PR #498, opened 2019-01-21
COSBETA-35: Business account registration

tom-bell-softwire:


keycloak/providers/registration-form/src/main/java/uk/gov/beis/mspsds/keycloak/providers/RegistrationMobileNumber.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public void validate(ValidationContext context) {
+        MultivaluedMap<String, String> formData = context.getHttpRequest().getDecodedFormParameters();
+        List<FormMessage> errors = new ArrayList<>();
+
+        context.getEvent().detail(Details.REGISTER_METHOD, "form");
+        String eventError = Errors.INVALID_REGISTRATION;
+
+        String mobilePhoneNumber = formData.getFirst(MOBILE_NUMBER_FIELD);
+        if (Validation.isBlank(mobilePhoneNumber)) {
```
broder: Do we need to do any validation other than this?

tom-bell-softwire: I looked at using PhoneNumberLib's [`parse()`](https://github.com/googlei18n/libphonenumber#quick-examples) method, however it requires a country to be specified, e.g.
```java
PhoneNumberUtil.getInstance().parse(phoneNumber, "GB")
```

I'm not sure we can assume "GB" for everyone (could there be users from, e.g., Jersey?).

We could use some form of basic RegEx check, but that feels fiddly and may prevent some legitimate number formats.

What do you think?

tom-bell-softwire: Sorry, I meant `libphonenumber` in my comment above (not PhoneNumberLib)!

It does have methods to check if a number is valid and if it is a mobile number, but does require an assumption about the country. I'm not sure we want to restrict users to UK-only numbers.

PR #499, opened 2019-01-21
MSPSDS-869 update existing business pages

MatthewBurstein:


mspsds-web/app/helpers/businesses_helper.rb
```diff
@@ -27,47 +27,35 @@ def sort_direction
...
-      :company_type_code,
-      :company_status_code,
-      :nature_of_business_id,
-      :additional_information,
-      location_attributes: %i[id address phone_number locality country postal_code _destroy]
+      :id,
+      :legal_name,
+      :trading_name,
+      :company_number,
+      locations_attributes: %i[id address_line_1 address_line_2 phone_number county country postal_code _destroy],
```
broder: Is `_destroy` used anymore?

mspsds-web/app/helpers/businesses_helper.rb
```diff
@@ -27,47 +27,35 @@ def sort_direction
...
       defaults_on_primary_location(@business)
       @business.source = UserSource.new(user: current_user)
     else
       @business = Business.new
       @business.locations.build
+      @business.contact = Contact.new
     end
   end
 
   def advanced_search(excluded_ids = [])
```
broder: I think this method was used to add the suggestions on the right hand side so we won't need it anymore (and some of the others below as well).

mspsds-web/app/models/business.rb
```diff
@@ -75,11 +47,6 @@ def pretty_description
   def add_sic_code(c_h_info)
```
broder: Do we need this method anymore?

mspsds-web/app/controllers/businesses_controller.rb
```diff
@@ -31,18 +27,6 @@ def edit
     @business.locations.build unless @business.locations.any?
   end
 
-  # GET /businesses/suggested
-  def suggested
```
broder: I think there's quite a lot of `suggested` logic in the JavaScript which we should remove as well.

mspsds-web/db/migrate/20190115111318_update_business_and_location_fields.rb
```diff
@@ -0,0 +1,60 @@
+class UpdateBusinessAndLocationFields < ActiveRecord::Migration[5.2]
+  class Location < ApplicationRecord; end
```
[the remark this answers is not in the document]
MatthewBurstein: I did this on Jan's advice following the Stubbing Out Models section [here](https://gist.github.com/ccschmitz/7f4df754cc8933a2033c). Happy to remove it if you feel it isn't necessary though.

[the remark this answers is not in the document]
broder: Fair enough - seems sensible!

mspsds-web/app/helpers/businesses_helper.rb
```diff
@@ -17,7 +17,7 @@ def search_for_businesses(page_size)
   end
 
   def sort_column
```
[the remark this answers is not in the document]
MatthewBurstein: Does that mean that I don't need to take action here as part of this ticket?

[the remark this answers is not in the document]
broder: You don't need to sort this here, no.

mspsds-web/app/helpers/businesses_helper.rb
```diff
@@ -27,130 +27,29 @@ def sort_direction
...
-
-  def companies_house_new_business_form_url
-    @investigation.present? ? companies_house_investigation_businesses_path(@investigation) : companies_house_businesses_path
-  end
-
   def create_business
     if params[:business]
       @business = Business.new(business_params)
       @business.locations.build unless @business.locations.any?
+      @business.contact = Contact.new(business_params[:contact_attributes]) unless @business.contact
```
broder: Just to check, Rails doesn't automatically map `contact_attributes` in `business_params` to a contact model? (i.e. does it break if we remove this line?)

MatthewBurstein: Yes you're right this does seem to happen automatically. I've removed the line.

PR #500, opened 2019-01-21
MSPSDS-870 update corrective action

samlionheart:
- Remove product dropdown if there is one item
Currently just pre-selects, removing drop down would mean either a disabled text field or just plain text but designs for this need to be made (product and business creation pages may be created before corrective action details)

- Remove add product link if there is at least one item

- Make business optional

- Remove add business link entirely

- Update the add file section to match new designs on MSA flow


mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -53,24 +52,34 @@ h1.govuk-heading-xl
...
+            | Are there any files related to the action?
+        fieldset.govuk-fieldset
+          .govuk-radios.govuk-radios--conditional data-module="radios"  
+            .govuk-radios__item
+              = form.radio_button :related_file, "Yes", class: "govuk-radios__input",
+                      "data-aria-controls": "conditional-related-file"
+              = form.label :related_file, "Yes", class: "govuk-label govuk-radios__label"
+            #conditional-related-file.govuk-radios__conditional.govuk-radios__conditional--hidden
+              = render "related_attachment_fields", form: form,
+                file_blob: @file_blob, attachment_name: :file, title: ""
```
broder: Should we be giving the attachment a title?

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -34,16 +34,15 @@ h1.govuk-heading-xl
           - else
             span.govuk-hint
               ' Only products already added to the case are listed.
-              = link_to "Add a product", new_investigation_product_path(@investigation)
-            = form.collection_select :product_id, @investigation.products, :id, :name, { include_blank: true },
+            - @one_product = @investigation.products.size == 1 ? @investigation.products.first.id : ''
+            = form.collection_select :product_id, @investigation.products, :id, :name, { include_blank: true, selected: @one_product  },
```
broder: We could also do this with `selected: @investigation.products.size == 1 && @investigation.products.first.id` which might be a little bit neater? Same for the business picker below

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
samlionheart: I just added a check to see if title was initialised in related_attachment_fields before being used so it doesn't need to be assigned to null

broder:
I think we should also update the testing activities to use the new logic for selecting products, businesses and files i.e. make https://mspsds-design-prototype.herokuapp.com/pages/flows/record-test-result/01-record-test-result and https://mspsds-design-prototype.herokuapp.com/pages/flows/record-testing-request/01-record-testing-request more like https://mspsds-design-prototype.herokuapp.com/pages/flows/record-corrective-action/01-record-corrective-action

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -53,24 +53,37 @@ h1.govuk-heading-xl
           - if @investigation.businesses.empty?
             span.govuk-hint
               | There are no businesses associated with this case.
-            span.govuk-hint
-              = link_to "Add a business", new_investigation_business_path(@investigation)
-              |  to record a corrective action.
           - else
             span.govuk-hint
               ' Only businesses already added to the case are listed.
               = link_to "Add a business", new_investigation_business_path(@investigation)
```
broder: Was this intentionally re-added during the merge?

samlionheart: Yes, I thought as interim until this is decided (whether there should be an "add business" page before hand https://regulatorydelivery.atlassian.net/browse/MSPSDS-921

broder: Fair enough

PR #501, opened 2019-01-21
COSBETA-23: Scan uploaded file with antivirus

aidayounesi:
Commonising antivirus between mspsds and cosmetic.
Set up the storage initializer for cosmetics.

***Tested manually with EICAR test file


shared-web/app/analyzers/shared/web/anti_virus_analyzer.rb
```diff
@@ -0,0 +1,32 @@
+module Shared
+  module Web
+    class AntiVirusAnalyzer < ActiveStorage::Analyzer
+      def initialize(blob)
+        config = { daemonize: true }
+        config[:config_file] = "clamav/clamd.conf" if Rails.env.production?
+        Clamby.configure(config)
```
broder: Should we move the `Clamby` dependency into the `shared-web.gemspec`?

aidayounesi: Sure, thanks

PR #509, opened 2019-01-23
MSPSDS-688: Improve navigation

kicferk1:
Currently we decide whether we use back link or breadcrumbs based on request.referer. It is probably easiest for now, since the logic isn't very complicated(back link when coming from case page, breadcrumbs otherwise), but if we want to implement more of it(say, adding a product in corrective action letting you go back to corrective action) then it might be better to just include some url params.


mspsds-web/app/helpers/url_helper.rb
```diff
@@ -25,12 +25,13 @@ def remove_associated_document_path(parent, document)
...
   def build_back_link_to_case
-    result = { is_simple_link: request.referer&.match?(/cases\/\d/) }
-    if result[:is_simple_link]
-      case_id = request.referer.split(/cases\//)[1].split(/[?\/#]/)[0]
-      kase = Investigation.find(case_id)
-      result[:simple_link_text] = "Back to #{kase.pretty_description}"
-      result[:link_to] = kase
+    case_id = request.referer&.match(/cases\/(\d+)/)&.captures&.first
+    result = nil
+    if case_id.present?
```
broder: Could invert the `if`? i.e. `return nil unless case_id.present?`

Then the below could be
```
investigation = Investigation.find(case_id)
{
is_simple_link: true,
link_text: "Back to #{investigation.pretty_description}",
link_to: investigation
}
```

PR #515, opened 2019-01-24
COSBETA-20: Record basic product details (single component)

DWRendell:
This is still in progress, just putting up what I've done so far.

**Done:**
- Refactored the ManualEntryController into NotificationBuildController, reworked routes a little (start of manual journey is now `/notifications/new`
- Added Component model, added page that adds a single component to a model
- Added ComponentBuildController to handle steps for filling in component fields
- Added fork where users can add two shades

**Still todo:**
- Allow arbitrary numbers of shades to be added rather than hardcoding inputs
- Add fork for setting import country (incl using the govuk country autocomplete)
- Add tests for component/component build controller


cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+            @component.shades = [ '', '' ]
+        end
+        render_wizard
+    end
+
+    def update
+        case step
+        when :number_of_shades
+            if params[:number_of_shades] == 'single'
+                redirect_to edit_notification_path(@component.notification)
```
broder: `redirect_to finish_wizard_path`?

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -0,0 +1,71 @@
+class ComponentBuildController < ApplicationController
+    include Wicked::Wizard
+
+    steps :number_of_shades, :add_shades
+
+    before_action :set_component
+
+    def show
+        if step == :add_shades && @component.shades.nil?
```
broder: We could format this as `@component.shades = [ '', '' ] if step == :add_shades && @component.shades.nil?` if it's any neater

cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  steps :add_product_name, :add_external_reference, :is_imported, :add_import_country, :single_or_multi_component
+
+  before_action :set_notification
+  before_action :set_countries, only: %i[show]
+
+  def show
+    render_wizard
+  end
+
+  def update
```
broder: Is this going to get really complicated?

cosmetics-web/app/models/component.rb
```diff
@@ -0,0 +1,37 @@
...
+    state :empty, initial: true
+    state :component_complete
+
+    event :add_shades do
+      transitions from: :empty, to: :component_complete
+    end
+  end
+
+  def prune_blank_shades
+    unless self[:shades].nil?
```
broder:
```
return if self[:shades].nil?

self[:shades] = self[:shades].select { |shade| shade.present? }
```
https://api.rubyonrails.org/classes/Object.html#method-i-present-3F is the opposite of `blank?`

cosmetics-web/app/models/component.rb
```diff
@@ -0,0 +1,37 @@
...
+      self[:shades] = self[:shades].select do |shade|
+        !shade.blank?
+      end
+    end
+  end
+
+private
+
+  def update_notification_state
+    notification.set_single_or_multi_component!
```
broder: I think when we get to multi-components, the component could maybe pass itself as an argument and then the notification can check if all components are `component_complete`

cosmetics-web/app/views/component_build/add_shades.html.slim
```diff
@@ -0,0 +1,20 @@
+- content_for :page_title, "List the shades"
+- content_for :after_header do
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+- shades_errors = @component.errors.messages[:shades].any?
```
broder: Do you think we should start to pull out some of these UI components into partials or are the differences too large between the different pages?

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
```diff
@@ -0,0 +1,33 @@
...
+            li hair dye sets
+            li skin peeling kits
+            li permanent styling products
+            li anti-age/revitalising kits
+        = form_with url: wizard_path, method: :put do |form|
+            .govuk-form-group
+                .govuk-radios
+                    .govuk-radios__item
+                        = form.radio_button :single_or_multi_component, "single", class: "govuk-radios__input"
+                        = form.label :single_or_multi_component_single, "Yes", class: "govuk-label govuk-radios__label"
```
broder: I think you need to add `value: "single"` to get the label to point to the right input element https://apidock.com/rails/ActionView/Helpers/FormHelper/label

cosmetics-web/app/views/notifications/edit.html.slim
```diff
@@ -43,11 +44,43 @@
                         | External reference
                     td.govuk-table__cell
                         = @notification.external_reference
+                tr.govuk-table__row
+                    th.govuk-table__header
+                        | Imported
+                    td.govuk-table__cell
+                        - if @notification.import_country.nil?
```
broder: Could do something like `= @notification.import_country.present? ? "Yes" : "No"`.

cosmetics-web/app/views/notifications/edit.html.slim
```diff
@@ -43,11 +44,43 @@
...
+                    th.govuk-table__header
+                        | Number of components
+                    td.govuk-table__cell
+                        = @notification.components.length
+                - if @notification.components.first.shades.nil?
+                    tr.govuk-table__row
+                        th.govuk-table__header
+                            | Shades
+                        td.govuk-table__cell
+                            | N/A
```
broder: Does `= @notification.components.first&.shades.join(', ') || "N/A"` work here?

cosmetics-web/spec/controllers/component_build_controller_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        end
+
+        it "initialises shades array with two empty strings in add_shades step" do
+            component = create_component
+            get(:show, params: { component_id: component.id, id: 'add_shades' })
+            expect(assigns(:component).shades).to eq([ '', '' ])
+        end
+    end
+
+    describe "POST #update" do
```
broder: Good tests!

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
DWRendell: I think adding the `_single` at the end of the label ID is meant to get it to point to the right element. Adding the value doesn't seem to change the HTML.

cosmetics-web/app/views/notifications/edit.html.slim
DWRendell: No luck, the exception still gets thrown if shades is nil. I've tidied it up a bit though to use a ternary instead.

broder: You can stick in extra `&`s on whichever bits are `nil`. (I had thought that `components.first` was the `nil` bit)

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
broder: I'd probably err on changing the label ID to be just `:single_or_multi_component` and then add `value`.

I think `= form.label :single_or_multi_component, "Created: newest", value: "single", class: "govuk-label govuk-radios__label"` should change the HTML. Did you definitely have the correct order? Maybe `value` has to be before `class`.

DWRendell: Ah OK, will give it a try

cosmetics-web/app/views/notifications/edit.html.slim
DWRendell: Oh cool, that's a neat trick. That works, thanks

cosmetics-web/app/controllers/notification_build_controller.rb
DWRendell: _Hopefully_ not. I've just extracted the two radio case statements out to separate functions to try and clean it up a bit. I don't think there's any more unusual cases to add to this controller at least (there's only really the image uploads to add to this one which shouldn't involve anything special).

The component_build_controller could get a little messy with all the routes for formulation uploads, but hopefully we can reduce that by splitting it out into a separate controller.

cosmetics-web/app/models/component.rb
DWRendell: Definitely sounds like a good idea. I think AASM guards might be an option as well, so we could call this event and only transition the notification state if all components are complete or something along those lines.

cosmetics-web/app/views/component_build/add_shades.html.slim
DWRendell: I've pulled the error message out, but will probably replace them with the ones Tom has made once COSBETA-36 is merged

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
DWRendell: Turns out the reason this wasn't working was a missing config value that's present in MSPSDS: `config.action_view.form_with_generates_ids = true`. Do you think this is worth adding to the initialiser in shared-web where we turn off the field_with_errors wrapper?

cosmetics-web/app/views/notifications/edit.html.slim
DWRendell: Done

cosmetics-web/spec/controllers/component_build_controller_spec.rb
DWRendell: Thanks!

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
broder: https://guides.rubyonrails.org/configuring.html#configuring-action-view says 

> config.action_view.form_with_generates_ids determines whether form_with generates ids on inputs. **This defaults to true**.

Is that wrong?

cosmetics-web/Gemfile
```diff
@@ -12,6 +12,7 @@ gem "pg", "~> 1.1"
 gem "puma", "~> 3.0"
 gem "rails", "~> 5.2"
 gem 'rails-controller-testing'
+gem 'rspec-collection_matchers'
```
tom-bell-softwire: Should this be a `:test` dependency instead (assuming we're only using it in tests)?

cosmetics-web/app/views/notification_build/single_or_multi_component.html.slim
DWRendell: I guess so, or something else is happening. Removing that config value and restarting causes the IDs to disappear off the radio inputs, and the labels lose the for tags.

cosmetics-web/Gemfile
DWRendell: Good point, I've moved them. Thanks.

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    case step
+    when :number_of_shades
+      case params[:number_of_shades]
+      when "single"
+        @component.shades = nil
+        redirect_to finish_wizard_path
+      when "multiple"
+        render_wizard @component
+      when ""
+        @component.errors.add :shades, "Must not be nil"
```
broder: Is this error message obvious for users?

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def finish_wizard_path
+    edit_notification_path(@component.notification)
+  end
+
+private
+
+  def component_params
+    params.require(:component).permit(shades: [])
```
broder: Is the `.permit(shades: [])` doing anything more than `.permit(:shades)`?

cosmetics-web/db/migrate/20190131114656_remove_external_reference_from_notification.rb
```diff
@@ -0,0 +1,17 @@
...
+  def change
+    safety_assured do
+      reversible do |dir|
+        change_table :notifications do |t|
+          dir.up do
+            t.remove :external_reference
+          end
+
+          dir.down do
+            t.integer :external_reference
```
broder: Was `external_reference` a string?

cosmetics-web/config/application.rb
```diff
@@ -32,5 +32,7 @@ class Application < Rails::Application
     config.eager_load_paths << Rails.root.join("presenters")
     config.active_job.queue_adapter = :sidekiq
     config.action_mailer.deliver_later_queue_name = 'cosmetics-mailers'
+
+    config.action_view.form_with_generates_ids = true
```
broder: Sorry for not answering your question earlier, but it would be good if this was in shared web.

cosmetics-web/app/controllers/component_build_controller.rb
DWRendell: It's not displayed to users right now, I'll update it in case that changes later

cosmetics-web/app/controllers/component_build_controller.rb
DWRendell: Yep, it doesn't allow you to submit an array of values unless you assign it to an empty array:
https://edgeguides.rubyonrails.org/action_controller_overview.html#permitted-scalar-values
>To declare that the value in `params` must be an array of permitted scalar values, map the key to an empty array: `params.permit(id: [])`

cosmetics-web/db/migrate/20190131114656_remove_external_reference_from_notification.rb
DWRendell: Yes, thanks

cosmetics-web/config/application.rb
DWRendell: Will do, I'll put it in the same initialiser as that other config change and change it's name to something more generic like shared_rails_config

PR #523, opened 2019-01-28
COSBETA-23: Update how Clamd is backgrounded to fix "Could not connec…

broder:
…t to clamd on LocalSocket /home/vcap/app/clamav/clamd.ctl: Connection refused" error

Pushing locally suggests that it fixes the `mspsds-worker`... 🤞

### DIFF
--- a/shared-worker/clamav/clamd.conf
+++ b/shared-worker/clamav/clamd.conf
@@ -2,12 +2,12 @@
 #To reconfigure clamd run #dpkg-reconfigure clamav-daemon
 #Please read /usr/share/doc/clamav-daemon/README.Debian.gz for details
 LocalSocket /home/vcap/app/clamav/clamd.ctl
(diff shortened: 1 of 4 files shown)


shared-worker/clamav/clamd.conf
```diff
@@ -27,7 +27,7 @@ PreludeAnalyzerName ClamAV
 DatabaseDirectory /home/vcap/app/clamav
 OfficialDatabaseOnly false
 SelfCheck 3600
-Foreground false
+Foreground true
```
broder: I think this is the key change (along with the extra `&` in the startup command)

DWRendell:
Thanks. Really strange this started happening with our changes.

broder:
It definitely started happening before - no one spotted it until your changes though! 🙈

PR #507, opened 2019-01-23
COSBETA-6: Build landing page

raamSoftwire:

PR #524, opened 2019-01-28
COSBETA-6: Add landing page

tom-bell-softwire:
This PR reproduces the changes from #507 to work around a Travis build issue.

Note that I've removed the line length limit for the header and footer view partials in our slim-lint config to avoid having to break up some of the long lines for SVG paths, etc.


raamSoftwire:
Thanks for taking this on and fixing it up Tom. It looks really good!

PR #527, opened 2019-01-29
MSPSDS-780: Upgrade Keycloak to 4.8.3

tom-bell-softwire:


keycloak/configuration/initial-setup.json
```diff
@@ -2691,27 +2691,47 @@
       }
     } ]
   }, {
-    "id" : "6f622199-71b2-4988-9f52-7d37836ab04d",
+    "id" : "144b8112-a771-4c4a-a02a-450659d84ea8",
```
tom-bell-softwire: This missing scope is the cause of the issues we were seeing previously with the client service accounts being unable to fetch users from the realm-management client REST API.

I'm now looking at how this can be updated in the admin console, so that we can apply the equivalent change to our Int/Staging/Prod environments.

keycloak/configuration/standalone-ha.xml.patch
```diff
@@ -0,0 +1,67 @@
+--- standalone-ha-original.xml	2018-08-15 20:00:06.000000000 +0100
```
tom-bell-softwire: I've created these patch files to make it easier to apply our config changes to the `standalone-ha.xml` and `standalone.xml` configuration files that come with Keycloak and are sometimes modified as part of an update.

In future, we should download the relevant versions of these files for the version of Keycloak we want to use and then apply these patches to them.

keycloak/configuration/initial-setup.json
broder: Nice, good spot!

keycloak/configuration/standalone-ha.xml.patch
broder: Good idea.

tom-bell-softwire:
> I'm guessing that we don't want to merge this until we figure out how to update existing Keycloak instances?

Right!

keycloak/configuration/initial-setup.json
tom-bell-softwire: I've added instructions on how to update the configuration in the admin console after upgrading:
https://regulatorydelivery.atlassian.net/wiki/spaces/MSPSDS/pages/627408919/Keycloak+Upgrades

tom-bell-softwire:
I've now added instructions on how to update the configuration in the admin console after upgrading:
https://regulatorydelivery.atlassian.net/wiki/spaces/MSPSDS/pages/627408919/Keycloak+Upgrades

PR #529, opened 2019-01-29
MSPSDS-566 create report as msa

MatthewBurstein:
Incomplete PR, looking for feedback on the first part of the ticket which does not implement logic we have not already thought through in other parts of the app.


mspsds-web/app/views/businesses/_form.html.slim
```diff
@@ -80,5 +81,5 @@ fieldset.govuk-fieldset class=["govuk-!-margin-bottom-5"]
       = ff.label :job_title, "Job title / description",
               class: "govuk-label"
       = ff.text_field :job_title, class: "govuk-input"
-
-= form.submit "Save business", class: "govuk-button"
+- submit_text ||= "Save business"
```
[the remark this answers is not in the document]
MatthewBurstein: This throws an error because submit_text is not defined

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -0,0 +1,478 @@
...
+    else
+      @selected_businesses = session[:selected_businesses]
+      @other_business_type = session[:other_business_type]
+    end
+  end
+
+  def set_business
+    if params.include?(:business)
+      @business = Business.new business_step_params
+      @business.locations.build business_step_params[:locations_attributes]["0"]
```
broder: Why do we need the `["0"]` for locations but not contacts?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -0,0 +1,478 @@
...
+  def product_step_params
+    product_session_params.merge(product_request_params).symbolize_keys
+  end
+
+  def business_step_params
+    business_session_params.merge(business_request_params).symbolize_keys
+  end
+
+  def business_session_params
+    # TODO use this to retrieve a business for editing eg for browser back button
```
broder: What's happening with these?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -0,0 +1,478 @@
...
+
+  def store_other_information
+    other_information_types.each do |key|
+      session[further(key)] = other_information_params[key] == "1" ? "Yes" : "No"
+    end
+  end
+
+  # We use 'further' to refer to the boolean flags indicating
+  # whether the user wants to provide another entry of a given type
+  def further(key)
```
broder: I don't think `further` describes what this is doing very well - would something like `further_key` be better?

mspsds-web/app/helpers/corrective_actions_helper.rb
```diff
@@ -0,0 +1,53 @@
+module CorrectiveActionsHelper
```
broder: [Concerns](https://api.rubyonrails.org/classes/ActiveSupport/Concern.html) might be the better pattern here than helpers, I think

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -0,0 +1,478 @@
+class Investigations::TsInvestigationsController < ApplicationController
```
broder: Could we split this up using concerns or are the helpers already pulling out as much as they can?

mspsds-web/app/views/businesses/_form.html.slim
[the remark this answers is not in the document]
broder: What about `form.submit (submit_text || "Save business"), class: ...`?

mspsds-web/app/views/businesses/_form.html.slim
```diff
@@ -27,6 +25,7 @@ fieldset.govuk-fieldset class=["govuk-!-margin-bottom-5"]
 
 = form.fields_for :locations, business.primary_location do |ff|
   fieldset.govuk-fieldset class=["govuk-!-margin-bottom-5"]
+    = ff.hidden_field :name, value: "Main address"
```
broder: A malicious user could potentially modify this hidden field to make it say whatever they want. Should we define this in the controller / model instead?

mspsds-web/app/views/businesses/edit.html.slim
```diff
@@ -1,9 +1,10 @@
 - content_for :page_title, "Edit business"
 - content_for :after_header do
   = link_to "Back to business", @business, class: "govuk-back-link"
-.edit-business-page.businesses-search-form
+.edit-business-page
```
broder: Is `.edit-business-page` actually doing anything? There seem to be some other classes like this as well (e.g. `.new-business-page`)

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -10,81 +10,13 @@ h1.govuk-heading-xl
   .govuk-grid-column-two-thirds
     = form_with model: @corrective_action, local: true, url: wizard_path, method: :put do |form|
       = render "form_validation_display", errors: @corrective_action.errors, model_name: "action"
+        / TODO check this still works properly
```
broder: Have you checked that this still works?

mspsds-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
```diff
@@ -0,0 +1,36 @@
+- content_for :page_title, "Report an unsafe or non-compliant product"
+- content_for :after_header do
+  = link_to "Back to list", investigations_path, class: "govuk-back-link"
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      span.govuk-caption-xl Report an unsafe or non-compliant product
+      = title
+- scope ||= nil
```
broder: Is this doing anything?

mspsds-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
```diff
@@ -0,0 +1,36 @@
...
+        .mspsds-input class="govuk-!-margin-bottom-5"
+          fieldset.govuk-fieldset
+            legend.govuk-fieldset__legend.govuk-fieldset__legend--l class="govuk-!-margin-bottom-1"
+              = other_text
+            .govuk-form-group.full-width class=("govuk-form-group--error" if model.errors[further_key].any?)
+              .govuk-radios
+                .govuk-radios__item
+                  = radio_button "", further_key, "Yes", class: "govuk-radios__input", value: "Yes",
+                          checked: @repeat_step == "Yes"
+                  = label further_key, "Yes", class: "govuk-label govuk-radios__label"
```
broder: I think we need to add `value: "XXX"` to labels to get them to correctly refer to the radio buttons. https://apidock.com/rails/ActionView/Helpers/FormHelper/label

mspsds-web/app/views/investigations/ts_investigations/other_information.html.slim
```diff
@@ -0,0 +1,62 @@
...
+          legend.govuk-fieldset__legend.govuk-fieldset__legend--m class="govuk-!-margin-bottom-1"
+            | What information do you have?
+          .govuk-form-group
+            .govuk-checkboxes
+              .govuk-checkboxes__item
+                = form.check_box :test_results, class: "govuk-checkboxes__input",
+                        "aria-describedby": "test-results-hint", name: "test_results",
+                        value: "Test results"
+                = form.label :test_results, "Test results", class: "govuk-label govuk-checkboxes__label",
+                        for: :investigation_test_results
```
broder: What's the `for` doing on these labels? It might be related to the comment about adding `value:` above.

mspsds-web/db/migrate/20190130110940_change_investigation_business_relationship_to_string.rb
```diff
@@ -0,0 +1,54 @@
...
+  def to_enum relationship_enum_title
+    case relationship_enum_title
+    when "manufacturer"
+      0
+    when "distributor"
+      1
+    when "importer"
+      2
+    else
+      raise "Unexpected relationship enum title!"
```
broder: If we're allowing custom relationships, won't this be fairly certain to fail - should we default it to `0`?

mspsds-web/test/system/create_ts_investigation_test.rb
```diff
@@ -0,0 +1,189 @@
...
+    fill_in "Title", with: risk_assessment_title
+    fill_in "Description", with: risk_assessment_description
+    add_attachment risk_assessment_title
+    choose "further_risk_assessments_no", visible: false
+    click_button "Continue"
+
+    assert_selector "h1", text: "Find this in your system"
+    fill_in_complainant_reference
+
+    # TODO add assertions for corrective actions once corrective action audit activity bug has been fixed so activities
```
broder: Is there a ticket for this to be resolved?

mspsds-web/app/assets/javascripts/investigations/hazards.js
```diff
@@ -0,0 +1,10 @@
+$(document).ready(function () {
+  var element = document.querySelector('#hazard-type-autocomplete-container');
```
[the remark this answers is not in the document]
MatthewBurstein: I've updated the js files in this commit to use jquery. The variation in ids is because some are rails generated and some we have set ourselves. I guess we should use underscores for all the ones we have set ourselves, although changing that feels like it might turn into quite a lengthy operation. Should I do it as part of this ticket?

mspsds-web/app/assets/javascripts/investigations/ts_investigations/which_businesses.js
```diff
@@ -0,0 +1,32 @@
...
+  ];
+  var elementOther = document.getElementById('businesses_other');
+  var elementNone = document.getElementById('businesses_none');
+
+  var deselectOthers = function () {
+    normalElements.forEach(function (element) {
+      element.checked = false;
+    });
+    if (elementOther.checked) {
+      elementOther.click();
```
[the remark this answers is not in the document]
MatthewBurstein: Because clicking it is responsible for showing and hiding a text box which only happens when it is clicked, not when you set the value manually.

mspsds-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
MatthewBurstein: without this,  `form_with scope: scope` throws an error because the scope variable has not been declared

mspsds-web/db/migrate/20190130110940_change_investigation_business_relationship_to_string.rb
MatthewBurstein: Yes that's a good point. Wouldn't it be better to provide a hidden `unknown` option or something rather than producing inaccurate data?

mspsds-web/test/system/create_ts_investigation_test.rb
MatthewBurstein: The bug is actually fixed. Have added some assertions for this test.

mspsds-web/app/assets/javascripts/investigations/hazards.js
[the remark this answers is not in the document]
broder: No, we can address it later

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
MatthewBurstein: Yes good point. In fact we were not properly saving Contacts and Locations. Fixed now.

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
MatthewBurstein: I have a note to create another ticket to fix the back button functionality. These will be needed then.

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
MatthewBurstein: 👍

mspsds-web/db/migrate/20190130110940_change_investigation_business_relationship_to_string.rb
MatthewBurstein: This has been set to default to 0

mspsds-web/app/views/businesses/edit.html.slim
```diff
@@ -1,7 +1,6 @@
 - content_for :page_title, "Edit business"
 - content_for :after_header do
   = link_to "Back to business", @business, class: "govuk-back-link"
-.edit-business-page
   = render "business_heading", business: @business
```
broder: I think the indentation on this line will need to change now that you've deleted the enclosing `div` - is it getting put in the `after_header` block at the minute?

There are also some others that we can get rid of e.g. `.new-product-page`.

mspsds-web/app/views/investigations/tests/_form.html.slim
```diff
@@ -47,7 +47,7 @@ fieldset.govuk-fieldset
         - Test::Result.results.each do |(result, label)|
           .govuk-radios__item
             = form.radio_button :result, result, class: "govuk-radios__input"
-            = form.label :result, label, class: "govuk-label govuk-radios__label"
+            = form.label "result_#{result}", label, class: "govuk-label govuk-radios__label"
```
broder: Can't we do this with `value: result`?

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
broder: Did you see this comment?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
broder: Should we create the ticket now and reference it from here?

mspsds-web/app/assets/javascripts/investigations/ts_investigations/which_businesses.js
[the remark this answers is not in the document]
broder: Might be worth leaving a comment here to explain that.

mspsds-web/app/views/businesses/_form.html.slim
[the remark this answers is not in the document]
broder: Did this work?

mspsds-web/app/views/businesses/_form.html.slim
broder: Did you think about this?

mspsds-web/app/views/investigations/tests/_form.html.slim
MatthewBurstein: Yes much better :)

mspsds-web/app/views/businesses/edit.html.slim
MatthewBurstein: Yes sorry you're right, have resolved.

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
MatthewBurstein: I've created the issue [here](https://regulatorydelivery.atlassian.net/browse/MSPSDS-980)

mspsds-web/app/views/businesses/_form.html.slim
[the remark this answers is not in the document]
MatthewBurstein: Ah yes this works thanks :)

mspsds-web/config/routes.rb
```diff
@@ -16,6 +16,7 @@
   resources :question, controller: "investigations/question", only: %i[show new create update]
   resources :allegation, controller: "investigations/allegation", only: %i[show new create update]
   resources :project, controller: "investigations/project", only: %i[new create]
+  resources :msa_investigation, controller: "investigations/msa_investigations", only: %i[show new create update]
```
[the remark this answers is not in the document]
MatthewBurstein: renamed to ts_investigation

mspsds-web/app/views/businesses/_form.html.slim
```diff
@@ -80,5 +79,4 @@ fieldset.govuk-fieldset class=["govuk-!-margin-bottom-5"]
       = ff.label :job_title, "Job title / description",
               class: "govuk-label"
       = ff.text_field :job_title, class: "govuk-input"
-- submit_text ||= "Save business"
-= form.submit submit_text, class: "govuk-button"
+= form.submit (submit_text ||= "Save business"), class: "govuk-button"
```
broder: Does just `||` work (rather than `||=`)?

MatthewBurstein: resolved with `local_assigns[:submit_text]` as discussed

PR #531, opened 2019-01-30
MSPSDS-870 Update test request and result pages

samlionheart:
Matching test request and test result activity pages with corrective actions.


mspsds-web/app/controllers/investigations/tests_controller.rb
```diff
@@ -100,7 +99,13 @@ def attach_files
   end
 
   def save_attachment
-    @file_blob.save if @file_blob
+    p "PARAMS"
```
broder: We probably don't want to commit this logging.

PR #536, opened 2019-01-31
MSPSDS-566 Use redis for session data storage

jasiekmiko:
The connection of this change to the msa flow lies in the amount of data
the msa flow wants to store in the session before committing everything
to the db at the end of it. We started running out of space in the 4kb
of the cookie.

Additionally, server-side session stores have
some security advantages over client-side:
https://www.happybearsoftware.com/almost-protect-yourself-from-cookie-session-store


jasiekmiko:
TODO:
[] Set up new redis instance on int
[] Set up new redis instance on staging
[] Set up new redis instance on prod

mspsds-web/config/redis_queue.yml
```diff
@@ -8,4 +9,4 @@ test:
   <<: *default
 
 production:
-  url: <%= ENV["VCAP_SERVICES"] && CF::App::Credentials.find_by_service_name("mspsds-redis")["uri"] %>
+  url: <%= ENV["VCAP_SERVICES"] && CF::App::Credentials.find_by_service_name("mspsds-redis-queue")["uri"] %>
```
jasiekmiko: @broder I imagine for this I'll need to create new service instances and, once deployed, delete the old ones. Is there something to be done to not loose pending jobs here?

broder: Would https://cli.cloudfoundry.org/en-US/cf/rename-service.html be useful here?

jasiekmiko: 🤦‍♂️ yes! :P
I thought we can't rename a bound service and wasn't sure if unbinding while it's running would be OK to do, but I must have confused the limitations with that of upgrading. Thanks!

mspsds-web/README.md
```diff
@@ -130,9 +130,11 @@ To create an Elasticsearch instance for the current space:
 To create a redis instance for the current space. 
 
     cf marketplace -s redis
-    cf create-service redis tiny-unclustered-3.2 mspsds-redis
+    cf create-service redis tiny-unclustered-3.2 mspsds-redis-queue
```
broder: I don't think `tiny-unclustered-3.2` exists anymore:

```
Hi,

We have released a new Redis plan - medium-ha-3.2. This provides 6.38GiB memory and provides a hot standby service for failover in the event that the original service fails.

We have also enabled backups for our tiny-3.2 plan for all instances created from now on. Please contact support at gov-uk-paas-support@digital.cabinet-office.gov.uk if you would like backups enabled for your existing instance(s).

In light of these changes we're deprecating our tiny-unclustered-3.2 plan. All new instances should be created using the tiny-3.2 plan instead.

Please contact gov-uk-paas-support@digital.cabinet-office.gov.uk if you have any questions.

Regards,
GOV.UK PaaS Team
```

Hopefully that doesn't have much impact for us.

mspsds-web/manifest.yml
```diff
@@ -10,5 +10,6 @@ applications:
   services:
     - mspsds-database
     - mspsds-elasticsearch
-    - mspsds-redis
+    - mspsds-redis-queue
```
broder: I wonder if calling them `mspsds-queue` and `mspsds-session` would be better. Something about decoupling the function from the implementation?

jasiekmiko: Yeah, sounds good.

mspsds-web/README.md
jasiekmiko: hmm, it's not obvious from this if `tiny-3.2` is clustered? if it is, then the comment below that sidekiq doesn't work with clustered is a bit worrying...

jasiekmiko: cf lists what we've got currently as `tiny-3.2` so it looks like those got moved over automatically, / it's just a name convention change ..

PR #537, opened 2019-01-31
MSPSDS-547: Assign to team

kicferk1:
I ended up doing it via polymorphic associations and filling the pieces that seem to be missing from ActiveHash and needed to make it work. 

The logic of assign page changed quite significantly, since we now show the radio boxes list regardless of previously assigned entities, and so some tests were no longer needed. I added team functionality to mock keycloak data for tests, and it seems to work with our system.


kicferk1:
I think it might be useful to add unit tests for methods computing stuff that goes into the lists of radios and autocomplete boxes, that's still a todo, but hopefully a small and easy one now that team functionality is added to test setup.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -80,14 +80,17 @@ def status
   def assign
     return if request.get?
 
+    authorize @investigation, :assign?
```
broder: I think if you leave out the action name (i.e. `:assign?`), pundit will automatically use current action.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -80,14 +80,17 @@ def status
...
     ps = assignee_update_params
-    if User.where(id: ps[:assignee_id]).empty?
+
+    potential_assignees = User.where(id: ps[:assignable_id]) + Team.where(id: ps[:assignable_id])
```
broder: Did you think about having a base entity class and then we could do something like `Entity.where(id: ps[:assignable_id])`? I guess that this isn't the end of the world as these will be in memory queries.

mspsds-web/app/models/investigation.rb
```diff
@@ -115,12 +130,42 @@ def pretty_description
...
+    Team.get_visible_teams(current_user).each do |team|
+      teams << team
+    end
+    teams << assignee if assignee.is_a? Team
+    teams.uniq
+  end
+
+  def past_teams
+    activities = AuditActivity::Investigation::UpdateAssignee.where(investigation_id: id)
+    user_id_list = activities.map(&:assignable_id)
```
broder: `team_id_list`?

mspsds-web/app/models/team.rb
```diff
@@ -30,5 +33,25 @@ def self.all(options = {})
       @records ||= []
     end
   end
+
+  def display_name
+    return name if current_user.organisation == organisation
+
+    name == "OPSS Enforcement" ? "The Office for Product Safety and Standards" : name
```
broder: Do we need to do this for all of the office teams?

mspsds-web/app/models/team.rb
```diff
@@ -30,5 +33,25 @@ def self.all(options = {})
...
+  def full_name
+    display_name
+  end
+
+  def assignee_short_name
+    display_name
+  end
+
+  def self.get_visible_teams(user)
+    return Team.where(name: ["OPSS Enforcement", "OPSS Processing", "OPSS Incident management"]) if user.is_opss?
```
broder: Is there a way of doing this without using the name? Feels like it could be changed quite easily in the Keycloak admin UI which could cause bugs.

mspsds-web/app/models/user.rb
```diff
@@ -74,15 +74,19 @@ def is_opss?
...
-      select_options[label] = user.id
+  def self.get_team_members(user:)
+    users = []
+    user.teams.each do |team|
+      team.users.each do |team_member|
+        users << team_member
+      end
     end
-    select_options
+    users.uniq
```
broder: Could we use sets to keep the lists de-duplicated?

mspsds-web/app/views/investigations/_assignee_form.html.slim
```diff
@@ -0,0 +1,60 @@
...
+          = form.label :assignable_id_radio, user.display_name, value: user.id,
+                  class: "govuk-label govuk-radios__label"
+      .govuk-radios__item
+        = form.radio_button :assignable_id_radio, "Someone in your team", class: "govuk-radios__input",
+                "data-aria-controls": "conditional-assignee-team-mate"
+        = form.label :assignable_id_radio, "Someone in your team", value: "Someone in your team",
+                class: "govuk-label govuk-radios__label"
+      #conditional-assignee-team-mate.govuk-radios__conditional.govuk-radios__conditional--hidden
+        = render "investigations/assignee_selection", investigation: investigation, form: form,
+                key: :assignee_1_id,
```
broder: Is there a more descriptive name than `assignee_1_id`? `current_team_assignee_id` or something like that?

mspsds-web/app/views/investigations/_assignee_form.html.slim
```diff
@@ -0,0 +1,60 @@
...
+    .govuk-radios.govuk-radios--conditional data-module="radios"
+
+      - investigation.important_assignable_people.each do |user|
+        .govuk-radios__item
+          = form.radio_button :assignable_id_radio, user.id, class: "govuk-radios__input",
+                  checked: investigation.assignee == user
+          = form.label :assignable_id_radio, user.display_name, value: user.id,
+                  class: "govuk-label govuk-radios__label"
+      .govuk-radios__item
+        = form.radio_button :assignable_id_radio, "Someone in your team", class: "govuk-radios__input",
```
broder: Do we need to hide this if the user doesn't have a team? Or should we call it "Someone in your organisation"?

mspsds-web/app/views/investigations/assign.html.slim
```diff
@@ -1,41 +1,21 @@
...
+- if policy(@investigation).assign?
+  = render "assignee_form", investigation: @investigation
+- else
+  p
+    = "You do not have permission to assign this #{@investigation.case_type}."
+    br
+    | Contact
+    strong = " #{@investigation.assignee.display_name} "
+    | to change assignment.
+  = link_to "Return to case", @investigation, class: "govuk-button"
```
broder: Do we need this as well as the link at the top?

mspsds-web/db/migrate/20190129155220_add_team_assignee_to_investigation.rb
```diff
@@ -0,0 +1,10 @@
+class AddTeamAssigneeToInvestigation < ActiveRecord::Migration[5.2]
+  safety_assured do
+    change_table :investigations, bulk: true do |t|
+      t.uuid :assignable_id
```
broder: Do we need to migrate & drop `assignee_id`?

mspsds-web/app/models/team.rb
kicferk1: No, OPSS Enforcement is a special team that gets cases from other organisations. The other teams should not be visible unless you are in opss. 

I'll think about making it more general and more obvious.

mspsds-web/app/models/team.rb
kicferk1: There is no inherent meaning behind teams at the moment. 

We could check that its parent is OPSS(name check again) and that the name includes a keyword. That gives us more leeway with changing names(say, removing OPSS shortcut).

mspsds-web/app/views/investigations/assign.html.slim
kicferk1: Design has both of them.

mspsds-web/app/views/investigations/_assignee_form.html.slim
kicferk1: I don't know, asked design Ed H about it.

I heard some plans to treat small organisations as if they were teams, which would solve this problem, but there is no sign of it on ticket description, so I will dig some more.

kicferk1: After talking to Ed, we should hide it.

mspsds-web/app/controllers/investigations_controller.rb
broder: Possibly not. Do you know if a user and group can have clashing IDs? I wonder if we should be trying to pass whether the `assignable_id` param is for a user or team... Could we send it as `user_id` or `team_id` and infer from that?

mspsds-web/db/migrate/20190129155220_add_team_assignee_to_investigation.rb
```diff
@@ -1,10 +1,31 @@
...
+            add_index :investigations, %i[assignable_type assignable_id]
+          end
 
-    add_index :investigations, %i[assignable_type assignable_id]
+          dir.down do
+            t.uuid :assignee_id
+            Investigation.all.each do |investigation|
+              investigation.update! assignee_id: investigation.assignable_id
+            end
+            t.remove :assignable_id
```
broder: Do we need this in addition to the `change_table` `assignable_id` bit above?

mspsds-web/app/models/team.rb
```diff
@@ -49,9 +49,19 @@ def assignee_short_name
   end
 
   def self.get_visible_teams(user)
-    return Team.where(name: ["OPSS Enforcement", "OPSS Processing", "OPSS Incident management"]) if user.is_opss?
+    return Team.find_teams_in_organisation(%w[Enforcement Processing Incident], "Safety and Standards") if user.is_opss?
```
broder: I'm not sure partial matching makes this that much better... For example, it would be quite flakey to e.g. adding another team which has "Enforcement" in the name

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: In general we could, but the 'previously assigned' selection is a mix of users and teams :( 

After some digging it seems they used UUIDs: https://issues.jboss.org/browse/KEYCLOAK-289 
to generate their ids at some point. Searching their issues on https://issues.jboss.org/browse/KEYCLOAK-2217?jql=project%20%3D%20KEYCLOAK%20AND%20text%20~%20%22generate%20ID%22%20ORDER%20BY%20created%20DESC didn't reveal any documented change in the way they generate ids, so hopefully it's still UUIDs. 
Which means chances of clashing ids are pretty low. 

I can use user_id, team_id and entity_id to decrease them further, but it will probably add some complexity.

PR #541, opened 2019-02-01
MSPSDS-870 Corrective actions: Fix activity logging, clean up details and confirmation page

samlionheart:
**Fix activity log for corrective actions:** Corrective actions weren't showing in the activity log because it it required it to belong to a business, which was made optional in the corrective action model in the previous PR but not in the audit activity.

**Changed details page:** When there is one product, it shows a disabled text field(there are no designs for this but I think this makes sense) with the name of the product, but submits a hidden text field with the product id as the value. This seems a bit hacky but works fine.

**Confirmation page:** Styling


mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -31,13 +31,15 @@ h1.govuk-heading-xl
             span.govuk-hint
               = link_to "Add a product", new_investigation_product_path(@investigation)
               |  to record a corrective action.
+          - elsif @investigation.products.size == 1
+            = form.text_field :product_id, { value: @investigation.products.first.name, disabled: true, readonly: true },
```
broder: Does calling this field `product_id` cause any clashing issues with the hidden field below?

mspsds-web/app/views/investigations/corrective_actions/details.html.slim
```diff
@@ -31,13 +31,15 @@ h1.govuk-heading-xl
...
+            class: "govuk-input govuk-!-width-two-thirds"
+            = form.hidden_field :product_id, value: @investigation.products.first.id
           - else
             span.govuk-hint
               ' Only products already added to the case are listed.
             = form.collection_select :product_id, @investigation.products, :id, :name,
-                    { include_blank: true,
-                    selected: @investigation.products.size == 1 && @investigation.products.first.id },
-                    class: "govuk-select", id: "product-picker"
+                    { include_blank: true }, class: "govuk-select", id: "business-picker"
```
broder: This should probably be `product-picker` rather than `business-picker`

mspsds-web/db/schema.rb
```diff
@@ -143,9 +143,11 @@
     t.uuid "assignee_id"
     t.datetime "created_at", null: false
     t.text "description"
+    t.text "hazard_description"
```
broder: Has this come from another branch?

samlionheart: I'm not really sure where this came from, could have been from me messing around with switching to an old branch and using git stash. Sorry!

mspsds-web/db/schema.rb
```diff
@@ -146,6 +146,7 @@
     t.string "hazard_type"
     t.boolean "is_closed", default: false
     t.boolean "is_private", default: false, null: false
+    t.text "non_compliant_reason"
```
broder: Has this also come from another branch?

PR #544, opened 2019-02-01
COSBETA-36: Create responsible person

tom-bell-softwire:
Putting this up now to get feedback, but still need to add some tests.


cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -0,0 +1,98 @@
+class ResponsiblePersons::AccountWizardController < ApplicationController
+  include Wicked::Wizard
+
+  steps :create_or_join_existing, :join_existing, :select_type, :enter_details
```
broder: I wonder if we should have separate routes for `create_or_join` and `join` and then use a wizard for the `create` flow

cosmetics-web/app/controllers/landing_page_controller.rb
```diff
@@ -1,5 +1,12 @@
 class LandingPageController < ApplicationController
   skip_before_action :authenticate_user!
+  before_action :show_dashboard
 
   def index; end
+
+  def show_dashboard
+    return unless user_signed_in? && current_user.responsible_persons.present?
+
+    redirect_to responsible_person_path(current_user.responsible_persons.first)
```
broder: Should we allow them to read the content on the landing page rather than just redirecting them straight away? Maybe we should change the "Create an account" button to be "Continue to dashboard" if they're logged in... Ralph might have ideas for how / if the landing page is used on other services once you've logged in.

cosmetics-web/app/models/team_member.rb
```diff
@@ -1,4 +1,12 @@
 class TeamMember < ApplicationRecord
```
broder: Is `ResponsiblePersonUser` a better name for this many-to-many join?

cosmetics-web/app/controllers/landing_page_controller.rb
raamSoftwire: Will the user always want to pick the first RP as default? Will we need to provide an option for them to set a default RP? Definitely not something to worry about for now though...

cosmetics-web/app/models/team_member.rb
```diff
@@ -1,4 +1,12 @@
...
   belongs_to :responsible_person
   belongs_to :user
+
+  def full_name
+    user&.full_name
```
broder: Do we need to define this on `TeamMember` or should we just user `teamuser.user.whatever`?

cosmetics-web/app/models/user.rb
```diff
@@ -1,4 +1,12 @@
 class User < Shared::Web::User
   has_many :team_members, dependent: :destroy
   has_many :responsible_persons, through: :team_members
+
+  def self.find_or_create(user)
+    User.find_by(id: user[:id]) || User.create(user.except(:groups))
```
broder: Should we want to exclude `groups` in the `shared-web` user?

cosmetics-web/app/models/user.rb
```diff
@@ -1,4 +1,12 @@
...
+  end
+
+  def responsible_persons
```
broder: Don't we get this from the `has_many` above?

cosmetics-web/app/views/application/_error_summary.html.slim
```diff
@@ -0,0 +1,8 @@
+- if errors&.any?
```
broder: @DWRendell I think you were planning to do something like this?

cosmetics-web/app/views/application/_form_input.html.slim
```diff
@@ -0,0 +1,14 @@
+- errors = form.object.errors[field]
```
broder: Should / could these components be pushed into `shared-web`?

@jasiekmiko is this a component that we could use on MSPSDS?

cosmetics-web/config/locales/en.yml
```diff
@@ -20,4 +20,10 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
-  hello: "Hello world"
+  activerecord:
```
broder: Awesome!

cosmetics-web/app/models/team_member.rb
raamSoftwire: This is cool, didn't know you could do that. Nice!

cosmetics-web/app/controllers/landing_page_controller.rb
tom-bell-softwire: Great point! I think selecting an RP needs to be more deliberate in general, to avoid accidentally registering a product under the wrong one. I'll raise this with Ralph as it'll need some design thought and user testing.

cosmetics-web/app/views/responsible_persons/show.html.slim
```diff
@@ -14,27 +14,21 @@
           th.govuk-table__header[scope="row"]
             | Responsible person
           td.govuk-table__cell
-            | Example name
-          td.govuk-table__cell
-            = link_to "Edit", "/", class: "govuk-link--no-visited-state"
```
raamSoftwire: What's happening to the edit links? Are they going to be handled in a different way?

cosmetics-web/app/models/team_member.rb
tom-bell-softwire: It's not necessary, but I think this syntactic sugar makes the use of `TeamMember` a bit more readable.

cosmetics-web/app/models/team_member.rb
tom-bell-softwire: It's more idiomatic, but quite a confusing name! Part of the reason we need this at all is that ActiveHash doesn't support `has_and_belongs_to_many` so we need to make the join table a model in its own right.

Perhaps something like `ResponsiblePersonAccount` might make the relationship clearer?

cosmetics-web/app/models/user.rb
tom-bell-softwire: Good idea.

cosmetics-web/app/models/user.rb
tom-bell-softwire: This is where ActiveHash starts to break down as an abstraction, since ActiveRecord makes some assumptions about the models that don't apply to ActiveHash classes.

This workaround has been suggested to (partially) deal with this limitation:
https://github.com/zilkey/active_hash/issues/25

cosmetics-web/app/views/responsible_persons/show.html.slim
tom-bell-softwire: We've got COSBETA-44 in the backlog to cover editing responsible person details.

I think the implications of editing a responsible person still need thinking about, and there aren't designs for the edit page, so I've removed the link to avoid confusion.

cosmetics-web/app/models/user.rb
broder: Should we add a comment to explain this? Either on the `has_many` above or on this method.

cosmetics-web/app/views/application/_error_summary.html.slim
DWRendell: Yes, I've added something but its a lot more barebones than this. I'll definitely look at using this once it's been merged.

cosmetics-web/app/models/team_member.rb
broder: I think it's a confusing name because `ResponsiblePerson` is itself a confusing name... Should we rename that in the code (and continue calling it RP in the UI) or just live with it?

Confusing names aside, I'd expect in a codebase with `ResponsiblePerson` and `User` as models, `ResponsiblePersonUser` would be a join https://stackoverflow.com/questions/1764483/sql-join-table-naming-convention

tom-bell-softwire:
Yes, that feels like the right place to do it.

cosmetics-web/app/models/user.rb
tom-bell-softwire: Yes, good idea!

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
tom-bell-softwire: Done.

cosmetics-web/app/models/team_member.rb
tom-bell-softwire: I think we should follow domain terminology where possible, even if the naming isn't ideal, so let's live with it.

I'll rename this to `ResponsiblePersonUser`, though, to make it clear that it's essentially a join table.

cosmetics-web/app/views/application/_form_input.html.slim
tom-bell-softwire: Great! Yes, happy to move this to a shared component as part of that process.

cosmetics-web/app/controllers/landing_page_controller.rb
tom-bell-softwire: Good call! I discussed with Ralph and have changed the behaviour so you still hit the landing page, but with a button to jump to "My products" instead of sign in.

cosmetics-web/spec/models/responsible_person_spec.rb
```diff
@@ -0,0 +1,97 @@
+require 'rails_helper'
+
+RSpec.describe ResponsiblePerson, type: :model do
+  describe "create record" do
+    it "succeeds when all required attributes are specified" do
+      responsible_person = build(:responsible_person)
```
broder: Should / could we pull this out into a `let` at the top of the spec?

cosmetics-web/spec/factories/responsible_person.rb
```diff
@@ -0,0 +1,16 @@
+FactoryBot.define do
```
broder: This feels more powerful than fixtures to me - what do you think?

tom-bell-softwire: Definitely agree. You get nice things like inheritance and conditional/sequential properties as well.

cosmetics-web/spec/models/responsible_person_spec.rb
tom-bell-softwire: Done.

PR #547, opened 2019-02-05
MSPSDS-525: Notify team when assigned

kicferk1:
This solution sends a copy of confirmation to each team member separately. Ideally we would send all these emails in bulk, which seems to be in the feature list on [https://www.notifications.service.gov.uk/features](url). The problem is that there is no documentation for it, and more obvious ways of making it work failed. 

I asked them about it and hopefully they respond soon.


kicferk1:
Ok, I got the update. 

"The bulk sending is only currently offered through the web interface, by uploading a spreadhseet/csv of recipient data. For the API you'll need to make individual API requests for each message you want to send.

So how you've got that set up is the best way for the moment."

Sad but clear.

PR #549, opened 2019-02-05
MSPSDS-874 Understand case context at a glance

samlionheart:
May need minor content changes that are still in discussion


mspsds-web/app/views/investigations/_current_status.html.slim
```diff
@@ -19,5 +20,11 @@
             div[class="govuk-!-font-size-16"]
               = investigation.assignee.organisation.name
         td.govuk-table__cell [class="govuk-!-font-size-16"]
-          - if investigation.can_be_assigned_by(current_user)
-            = link_to "Change", assign_investigation_path(investigation)
+          = link_to "Change", assign_investigation_path(investigation)
```
kicferk1: I think we probably want to keep that 'if' there, because we restricted who can assign cases and people who can't will just see a page telling them to go back cause they can't assign it. Unless we changed our minds about it very recently :D

mspsds-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -0,0 +1,46 @@
+h2.govuk-heading-l[class="govuk-!-margin-bottom-6"]
+    | Summary
```
kicferk1: This file looks like it's indented with 4 spaces instead of 2 :(

mspsds-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -0,0 +1,46 @@
...
+          th.govuk-table__header Product category
+          td.govuk-table__cell = @investigation.product_category
+        tr.govuk-table__row
+          th.govuk-table__header  Hazard type
+          td.govuk-table__cell = @investigation.hazard_type
+- if @investigation.complainant
+    hr.govuk-section-break.govuk-section-break--m
+    h3.govuk-heading-m
+        | Reporter
+    table.govuk-table
```
kicferk1: Probably needs a gdpr check and generic 'hidden for gdpr reasons' if `investigation.complainant.can_be_displayed?` is false

mspsds-web/test/system/create_enquiry_test.rb
```diff
@@ -151,11 +151,19 @@ class CreateEnquiryTest < ApplicationSystemTestCase
...
-
+    click_on "Activity"
     assert_text "Attachment: #{attachment_filename}"
     assert_text "View attachment"
   end
 
+  test "enquiry details should be shown in overview"
+  select_complainant_type_and_continue
+  fill_complainant_details_and_continue
+  fill_allegation_details_and_continue
```
kicferk1: fill_enquiry_details_and_continue?

mspsds-web/app/views/application/_tabs.html.slim
```diff
@@ -8,5 +8,5 @@
           = tab.title
   - tabs.each do |tab|
     section.govuk-tabs__panel id=tab.id
-      h2.govuk-heading-l[class="govuk-!-margin-bottom-0"] = tab.title
+      h2.govuk-heading-l[class="govuk-!-margin-bottom-0"] = tab.title if tab.title != "Overview"
```
broder: Should we put something on the `Tab` presenter called `hide_title` which is false by default, but we set to true for the overview tab?

mspsds-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -0,0 +1,49 @@
+h2.govuk-heading-l[class="govuk-!-margin-bottom-6"]
+  | Summary
+.govuk-grid-row[class="govuk-!-padding-bottom-6"]
+  p.govuk-body[class="govuk-grid-column-two-thirds"]
```
broder: If the class doesn't have an `!` in it, you can just do `p.govuk-body.govuk-grid-column-two-thirds`

mspsds-web/app/views/investigations/tabs/_overview.html.slim
```diff
@@ -0,0 +1,49 @@
...
+h2.govuk-heading-l[class="govuk-!-margin-bottom-6"]
+  | Report details
+h3.govuk-heading-m
+  = @investigation.case_type.titleize
+table.govuk-table
+  tbody.govuk-table__body
+    tr.govuk-table__row
+      th.govuk-table__header [scope="row" class="govuk-!-width-one-quarter"]  Date recorded
+      td.govuk-table__cell = @investigation.created_at.strftime("%d/%m/%Y")
+      - if @investigation.case_type == "allegation"
```
broder: Maybe `if @investigation.is_a?(Investigation::Allegation)` would be neater?

mspsds-web/app/presenters/tab.rb
```diff
@@ -6,4 +6,8 @@ def initialize(id, title, render)
     @title = title
     @render = render
   end
+
+  def hide_title
+    @title == "Overview"
```
broder: We could pass this in as an argument to `initialize` (something like `def initialize(id, title, render, hide_title = false)`) and then we could specify it in the `investigations/show.html.slim` view where we create all the tabs.

PR #550, opened 2019-02-05
COSBETA-25: loading spinner for file upload

raamSoftwire:


cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -1,6 +1,12 @@
 class ResponsiblePersons::NotificationsController < ApplicationController
   before_action :set_responsible_person
 
+  def index
+    @pending_notification_files = NotificationFile.where(
+      ["responsible_person_id = ? and user_id = ?", @responsible_person.id, current_user.id]
```
raamSoftwire: I took this syntax from the rails docs, but it still looks odd to me. I've checked that it works, but want to know if there is a more elegant way to do this.

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
```diff
@@ -1,5 +1,19 @@
 - content_for :page_title, "Your cosmetic products"
 = render "layouts/rp_navbar"
+- if @pending_notification_files > 0
```
raamSoftwire: Is there a more Railsy way of doing this logic?

cosmetics-web/config/routes.rb
```diff
@@ -16,6 +14,7 @@
   end
 
   resources :responsible_persons do
+    resources :notification_files, controller: "responsible_persons/notification_files"
```
raamSoftwire: I wasn't sure here, should it be `only %i [edit, index, new, show]`?

cosmetics-web/db/migrate/20190204112230_add_responsible_person_to_notification_files.rb
```diff
@@ -0,0 +1,10 @@
+class AddResponsiblePersonToNotificationFiles < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :notification_files, :responsible_person, foreign_key: true, index: false
+    add_index :notification_files, :responsible_person_id, algorithm: :concurrently
+
+    add_column :notification_files, :user_id, :uuid
```
raamSoftwire: Given that I've also added a column, is it worth changing the filename to `xxxxxxx_ updating_notification_file`?

cosmetics-web/config/routes.rb
aidayounesi: Probably we only want new

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -26,12 +27,14 @@ def create
 
     if notification_file_params && notification_file_params[:uploaded_file]
       @notification_file.name = notification_file_params[:uploaded_file].original_filename
+      @notification_file.responsible_person_id = @responsible_person.id
```
broder: I think you can set it via `@notification_file.responsible_person = @responsible_person` but it's functionally the same

cosmetics-web/config/routes.rb
broder: Would we want `create` as well? It might be worth deleting the unused actions in the `NotificationFilesController` whilst we're at it.

cosmetics-web/db/migrate/20190204112230_add_responsible_person_to_notification_files.rb
broder: No, I think it's fine

cosmetics-web/db/schema.rb
```diff
@@ -49,6 +49,9 @@
     t.string "name"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.bigint "responsible_person_id"
+    t.uuid "user_id"
```
broder: I think we're using `string` for `user_id` in other places. I'm not sure if one is necessarily better than the other but we should be consistent.

cosmetics-web/spec/rails_helper.rb
```diff
@@ -51,6 +51,7 @@
   options.add_argument('--no-sandbox')
   options.add_argument('--disable-dev-shm-usage')
   options.add_argument('--window-size=1400,1400')
+  client.read_timeout = 180
```
broder: What is this doing?

cosmetics-web/spec/rails_helper.rb
```diff
@@ -106,3 +107,20 @@ def sign_out_user
 def fill_autocomplete(locator, with:)
   fill_in locator, with: "#{with}\n"
 end
+
+def sign_in_test_user
```
broder: Should we get rid of `authenticate_user`?

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -10,11 +10,11 @@
...
   it "enables to upload a file" do
-    visit new_notification_file_path
-
+    responsible_person = ResponsiblePerson.create
+    visit new_responsible_person_notification_file_path(responsible_person.id)
     page.attach_file('uploaded_file', Rails.root + 'spec/fixtures/testImage.png')
     click_button "Upload"
 
-    expect(page).to have_text("testImage.png")
+    expect(page).to have_text("Your cosmetic products")
```
broder: Is this still testing the right thing? (should it be testing that the file is uploaded and displayed?)

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
broder: No, I think what you've done looks good.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
broder: I think it's probably fine as is. Does `NotificationFile.where(responsible_person: @responsible_person, user: current_user).count` work?

cosmetics-web/app/assets/stylesheets/responsible_persons.scss
```diff
@@ -28,3 +28,33 @@
...
+
+  .loading_spinner__dot {
+    width: 10px;
+    height: 10px;
+    border-radius: 100%;
+    background-color: govuk-colour("black");
+    display: inline-block;
+    margin: 0 10px 3px 0;
+    opacity: 1;
+    animation-name: fadeInOpacity;
```
broder: Do we need to think about IE8? https://caniuse.com/#feat=css-animation

(The answer probably is no, or not yet)

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
broder: Also, do we actually want to filter it by the user as well as RP? Should I be able to see all draft notifications for my RP, regardless of who added them in the first place?

raamSoftwire: `responsible_person_id: @responsible_person.id, user_id: current_user.id` works, and feels more natural. 

We do want to filter by both RP and user for this method as it is used to show the number of files currently uploading in the loading spinner message. This should be specific to the user as you might have multiple people on a team uploading to an RP at the same time.

Further down the page, a user will be able to see all notifications for a given RP.

broder: Yeah, good point!

cosmetics-web/config/routes.rb
raamSoftwire: I've added create and new, and removed the other actions from the controller.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: That's cool to know, I've changed it now.

cosmetics-web/db/schema.rb
raamSoftwire: I saw we were using a uuid for `user_id` on MSPSDS and thought it is a Keycloak thing. Should I write a migration to update team members to use uuids for `user_id`?

cosmetics-web/spec/rails_helper.rb
raamSoftwire: My tests were failing locally and I was getting a `Net: Timeout` message. This is a known issue and the tests just need more time to run. This line tells it to wait for longer.

cosmetics-web/spec/system/upload_single_file_spec.rb
raamSoftwire: At this point we wouldn't expect the filename to be displayed. Only once it is processed and a `notification` rather than a `notification_file` would it be shown in the tabs below. This is being done as part of COSBETA-24, so for now I wasn't sure of the most sensible thing to test, other than the redirect to the correct page.

I didn't want to expect to have text from the loading spinner, because in a case where they are prcoessed very quickly, the user wouldn't see the spinner.

cosmetics-web/app/assets/stylesheets/responsible_persons.scss
raamSoftwire: No, or at least not yet 😉

cosmetics-web/db/schema.rb
broder: Yeah, sounds good

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
broder: Does that also work for `user`?

cosmetics-web/db/schema.rb
raamSoftwire: I've spoken to @tom-bell-softwire and he raised a really good point. Keycloak is providing the user id and it looks like a uuid, but because it is coming from a third party service we should be as flexible as possible and so it'd be better to use a string, rather than a uuid here. 

So I'll update notification_files to use string instead.

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
raamSoftwire: It looks likes `.positive?` is the Railsy approach according to Travis. https://api.rubyonrails.org/classes/Numeric.html#method-i-positive-3F

broder: You could also maybe move the logic into the controller and make it something like `@has_pending_notification_files = NotificationFile.where(...).any?` (or `exists?` https://stackoverflow.com/questions/48350500/differences-between-any-and-exists-in-ruby-on-rails)

raamSoftwire: I could do but then we end up with both 

```    @pending_notification_files = NotificationFile.where(
      responsible_person_id: @responsible_person.id, user_id: current_user.id
    ).count

    @has_pending_notification_files = NotificationFile.where(
        responsible_person_id: @responsible_person.id, user_id: current_user.id
    ).exists?
```
I don't know how expensive db lookups are, but it feels a bit weird to do this twice, rather than just query the value using `.positive?`.

We need both the boolean to decided whether or not to show the banner, and the number to show inside the banner.

broder: Yeah, ignore me - I forgot that we needed to display the number as well!

PR #551, opened 2019-02-05
MSPSDS-928: Filter by team

kicferk1:
The logic behind assignee filters is quite complex now, since we have 3(or more) checkboxes, each of which can be selected or not, and if 'other' is selected but no entity is chosen there is a different bit of logic for who to include. 

I refactored the logic, it passes previous tests, and is easier to follow. I'll check with Eds that this is what we want, and after that I might add more tests for all edge cases. 

There are some minor improvements to add here and there, but that's probably the meat of it.


mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,53 @@ def get_status_filter
...
 
-    if params[:assigned_to_me] == "unchecked" &&
-        params[:assigned_to_someone_else] == "checked" &&
-        params[:assigned_to_someone_else_id].blank?
-      excluded_assignees << current_user.id
-    end
+  def no_boxes_checked
+    no_people_boxes_checked = params[:assigned_to_me] == "unchecked" && params[:assigned_to_someone_else] == "unchecked"
+    no_teams_checked = true
+    teams_with_keys.each { |team| no_teams_checked = no_teams_checked && query_params[team[:key]].blank? }
```
broder: Would https://apidock.com/ruby/Enumerable/all%3F be helpful here?

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -112,4 +130,21 @@ def build_breadcrumb_structure
...
+    current_user.teams.map.with_index do |team, index|
+      {
+        key: "assigned_to_team_#{index}".to_sym,
+        team: team
+      }
+    end
+  end
+
+  def assignee_list_with_team(list, team)
+    list << team.id
```
broder: Would doing this with https://ruby-doc.org/core-2.2.0/Array.html#method-i-map and https://apidock.com/ruby/Array/concat be neater?

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,53 @@ def get_status_filter
...
+  def assignee_filter_exclusive
+    params[:assigned_to_someone_else] == "checked" && params[:assigned_to_someone_else_id].blank?
+  end
+
+  def compute_excluded_terms
+    excluded_assignees = []
+    teams_with_keys.each do |team_with_key|
+      if query_params[team_with_key[:key]].blank?
+        team = team_with_key[:team]
+        excluded_assignees = assignee_list_with_team(excluded_assignees, team)
```
broder: Should this be a `concat` or `<<`? (rather than overwriting the excluded_assignees with each team)

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,53 @@ def get_status_filter
...
+      end
     end
+    excluded_assignees << current_user.id if params[:assigned_to_me] == "unchecked"
+    excluded_assignees = excluded_assignees - [current_user.id] if params[:assigned_to_me] == "checked"
```
broder: Should we do an `if .. else` here? Might make it a bit easier to understand the logic.

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,53 @@ def get_status_filter
...
+    format_assignee_terms(excluded_assignees)
+  end
 
-    if params[:assigned_to_me] == "checked" &&
-        params[:assigned_to_someone_else] == "checked" &&
-        params[:assigned_to_someone_else_id].present?
-      assignees << current_user.id
+  def compute_included_terms
+    assignees = []
+    teams_with_keys.each do |team_with_key|
```
broder: Could we pull this into a method called `team_assignees` or something like that? Then we could reuse it above.

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,53 @@ def get_status_filter
...
+      if query_params[team_with_key[:key]].present?
+        team = team_with_key[:team]
+        assignees = assignee_list_with_team(assignees, team)
+      end
+    end
+    if params[:assigned_to_someone_else] == "checked"
       assignees << params[:assigned_to_someone_else_id]
+      team = Team.find_by(id: params[:assigned_to_someone_else_id])
+      assignees = assignee_list_with_team(assignees, team) if team.present?
```
broder: Should this be a `concat` again?

If it was a `concat` and the `assigned_to_someone_else_id` is a team, would the id be added twice?

mspsds-web/app/views/investigations/index.html.slim
```diff
@@ -30,14 +30,18 @@
               .govuk-checkboxes__item
                 = form.check_box :assigned_to_me, { class: "govuk-checkboxes__input" }, "checked", "unchecked"
                 = form.label :assigned_to_me, "Me", class: "govuk-checkboxes__label"
+              - teams_with_keys.each do |team|
```
broder: Could we make this a bit neater with destructuring? http://tony.pitluga.com/2011/08/08/destructuring-with-ruby.html Maybe such that it was something like `teams_with_keys.each do |team, key|`

mspsds-web/app/helpers/investigations_helper.rb
kicferk1: Yet it would.

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,49 @@ def get_status_filter
...
+  end
 
-    if params[:assigned_to_me] == "unchecked" &&
-        params[:assigned_to_someone_else] == "checked" &&
-        params[:assigned_to_someone_else_id].blank?
-      excluded_assignees << current_user.id
-    end
+  def no_boxes_checked
+    no_people_boxes_checked = params[:assigned_to_me] == "unchecked" && params[:assigned_to_someone_else] == "unchecked"
+    no_people_boxes_checked && teams_with_keys.all? { |key, _t, _n| query_params[key].blank? }
```
broder: I'd maybe keep the `no_team_boxes_checked` variable to make it easier to read

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,49 @@ def get_status_filter
...
-        params[:assigned_to_someone_else] == "checked" &&
-        params[:assigned_to_someone_else_id].present?
-      assignees << current_user.id
+  def compute_included_terms
+    # If 'Me' is not checked, but one of current_users teams is selected, we don't exclude current_user from it
+    assignees = checked_team_assignees
+    if params[:assigned_to_someone_else] == "checked"
       assignees << params[:assigned_to_someone_else_id]
+      team = Team.find_by(id: params[:assigned_to_someone_else_id])
+      assignees.concat(assignee_ids_from_team(team)) if team.present?
```
broder: Do we want to add the team ID twice?

mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -36,36 +36,49 @@ def get_status_filter
...
+  end
 
-    assignee_terms = format_assignee_terms(assignees)
-    excluded_assignee_terms = format_assignee_terms(excluded_assignees)
-    { should: assignee_terms, must_not: excluded_assignee_terms }
+  def checked_team_assignees
+    assignees = []
+    teams_with_keys.each do |key, team, _n|
+      if query_params[key].present?
+        team = team
```
broder: Probably don't need this

mspsds-web/test/controllers/team_test.rb
```diff
@@ -0,0 +1,198 @@
+require "test_helper"
+
+class TeamTest < ActionDispatch::IntegrationTest
+  setup do
+    sign_in_as_user
+    organisations = Organisation.all
+    @teams = Team.all
+    user_groups = [
```
broder: Should this be in a helper?

PR #557, opened 2019-02-06
MSPSDS-530: Restrict case access

kicferk1:
Updating case visibility to match designs: 
Removed full details tab, 
Added an option for "add activity" that has text depending on current visibility,
Made that option only appears if the user would be able to see the case after restricting it.


mspsds-web/app/views/investigations/visibility.html.slim
```diff
@@ -1,23 +1,32 @@
...
-      .govuk-radios
-        .govuk-radios__item
-          = form.radio_button :is_private, false, class: "govuk-radios__input"
-          = form.label :is_private, visibility_options[:public], class: "govuk-label govuk-radios__label"
-        .govuk-radios__item
-          = form.radio_button :is_private, true, class: "govuk-radios__input"
-          = form.label :is_private, visibility_options[:private], class: "govuk-label govuk-radios__label"
-  .govuk-form-group
-    = form.submit "Save", class: "govuk-button"
+  = "Visibility"
```
broder: Could just do `| Visibility` here.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -99,6 +99,7 @@ def assign
   def visibility
     return if request.get?
 
+    authorize @investigation
```
broder: Should we put this in a `before_action`? (it's also used in `assign`, and probably should be in `status`).

kicferk1: I'm not sure, putting it in before_action will prevent users from GETting the page as well(they will see generic 403 error), which doesn't seem fully in line with designs. 

I'll chat to Ed and ask him about visibility, assign and status stuff.

kicferk1: So Ed id not a fan of redirecting to 403, he'd rather have the user stay on a particular page, with more specific information than generic 'you don't have access to that'. 

In the scope of the whole app it is probably a story for another day, but since we have designs for what assign page should look like, we probably shouldn't hide it by generic 403. 

We could still do it with before_action, but only if we separated get actions from put actions. Which might be a way to do it? For now I used authorise without query and defined policies for other actions. 

Also, we Ed says anyone who can see a case can change its status for now.

broder: Ah sorry, I forgot that we were also doing an `authorize` call in `set_investigation`.

Should we do something like ``` if policy(@investigation).assign?
      @investigation.errors.add(:assignee, :invalid, message: "some message about permissions")
      respond_to_invalid_data(:assign)
      return
    end
```
if we don't want to display a 403?

PR #558, opened 2019-02-07
MSPSDS-907: Add maintenance page

broder:


maintenance/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+## Overview
+
+The site is written in [Node](https://nodejs.org/) using [Express](https://expressjs.com/) and uses the [GOV.UK Design System](https://design-system.service.gov.uk/).
+
+
+## Getting Setup
+
+Install Node and Yarn and run the site using `yarn start`.
```
broder: I've not added this to our `docker-compose.yml` because I don't think it'll change that much. It should be fairly straightforward to do though.

mspsds-web/deploy.sh
```diff
@@ -10,7 +10,29 @@ set -ex
 # SPACE: the space to which you want to deploy
 # If NO_START is set the app won't be started
 
+if cf app mspsds-web; then
```
broder: @tom-bell-softwire should I add this to cosmetics as well?

tom-bell-softwire: Yes, that would be great, thanks!

PR #560, opened 2019-02-07
MSPSDS-912 componentize radio buttons

jasiekmiko:
Early review request with focus on the structure of the components system.
The "form_components" folder is the alternative I came up with to having a custom form builder (as discussed earlier today with @broder), but am very open to other suggestions of structuring it.

Rest of the PR will include converting radio buttons throughout the app to follow suit. And also some documentation on the new direction


shared-web/app/helpers/shared/web/govuk_form_builder.rb
```diff
@@ -0,0 +1,25 @@
+class GovukFormBuilder
```
broder: As discussed, it might be better to just stick with partials.

shared-web/app/views/components/_govuk_error_message.html.slim
```diff
@@ -0,0 +1,8 @@
+/ Implementation based on
+/ https://github.com/alphagov/govuk-frontend/blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/error-message/template.njk
+span.govuk-error-message *(params[:attribute] || {}) id=params[:id] class=params[:classes]
```
broder: I think we should maybe just use local variables rather than hijacking `params`

shared-web/app/views/components_gallery/radios.html.slim
```diff
@@ -0,0 +1,389 @@
+/ Exmpales in this file are based on  https://github.com/alphagov/govuk-frontend/blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/radios/radios.yaml
```
broder: Neat!

shared-web/app/views/form_components/_govuk_radios.html.slim
```diff
@@ -0,0 +1,10 @@
+- if form && key
+  - params[:name] ||= form.object_name.present? ? "#{form.object_name}[#{key.to_s}]" : key.to_s
+  - if form.object
+    - if p form.object.send(key)
+      - item_matching_state = params[:items].find{|item| item[:value] == form.object.send(key) }
+      - item_matching_state[:checked] = true if item_matching_state
+    - if form.object.errors
+      - errorText = form.object.errors[key].first.presence
+      - params[:errorMessage] ||= {text: errorText} if errorText
+= render "components/govuk_radios", params: params
```
broder: If you want to pass all the local variables, you can use `local_assigns` (I think)

shared-web/app/views/layouts/shared/web/component_gallery.html.slim
```diff
@@ -0,0 +1,67 @@
+doctype html
```
broder: We should also componentise this in future :)

mspsds-web/app/controllers/application_controller.rb
```diff
@@ -17,7 +17,7 @@ def authorize_user
     return unless user_signed_in?
 
     unless current_user&.is_mspsds_user?
-      raise Pundit::NotAuthorizedError
+      # raise Pundit::NotAuthorizedError
```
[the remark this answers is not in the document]
jasiekmiko: oooops, rebase gone wrong. sorry!!

shared-web/app/helpers/shared/web/govuk_form_builder.rb
jasiekmiko: yup, wasn't meant to be here 😨

shared-web/app/views/components/_govuk_error_message.html.slim
jasiekmiko: Yeah, I've just tried it and it does come out quite a bit nicer.
It also allows for a nice distinction between required params (`name`) vs optional ones (`local_assigns[:name]`).
Also improved on some of the hand-wiring by using splats where appropriate

jasiekmiko:
The webpack changes have broken the gallery pages - i could use a hand in re-wiring it up correctly.

I also wanted to put in a simple test visiting the gallery page, but not sure if there's a way to make it app-agnostic without making another test entry point...

mspsds-web/app/views/application/_form_validation_display.html.slim
```diff
@@ -1,6 +1,17 @@
 - model_name = "item" if local_assigns[:model_name].blank?
```
jasiekmiko: I'm not particularly happy with this is a solution, but I think it does do some importnat exloratory work which can be improved upon in MSPSDS-888. However, some initial thoughts:
1. The use of the radios component alone forces a bit of a change to the error messages, since it incorporates them into the component. I'm expecting this to follow as other components get created.
2. Currently, our messages do not follow the guidelines very well (https://design-system.service.gov.uk/components/error-message/).
3. The alternative to providing the message "by hand" every time is to use the `full_messages_for` helper, for which we can define the format (currently at the default `{attribute_name} {message}`), but I'm not convinced it allows for enough flexibility to cover all messages and make them good.
4. If we do override the default message always (which is the approach assumed so far in this ticket), then there are two ways of achieving that, with examples of both found here: [on the validates directive](https://github.com/UKGovernmentBEIS/beis-mspsds/pull/560/files#diff-7174257140ecd4acd40bd0f75b023c8eR4) and [via the locale file](https://github.com/UKGovernmentBEIS/beis-mspsds/pull/560/files#diff-5c1fafb8523deabb20cf6849be308599R33). The latter is the only way to override messages from association validations (short of making associations optional and adding `validates` directives), but it does move the messages further away from the fields that use them. We probably want to pick one way to follow.

broder: Those thoughts sound sensible to me - I think we should probably define the full message manually to give us the flexibility as you said. It's unfortunate that we can't specify the error alongside the association. I think using the locale file would be fine or maybe doing something like `validates_associated` wouldn't be that bad? https://stackoverflow.com/a/24024223

jasiekmiko: One option would be to move all error messages to the locale file - as an extension of that, it could be all messages, not just error ones, though that would be a discussion for another day!

I don't think `validates_associated` isn't quite what we want. In fact, the whole topic is a bit tricky to google around for, as there's two things you could want to verify in an association relationship - the base model and the associated models. E.g. where we have in `TestResult` a `belongs_to :product` association and we call `test_result.validate`:
* rails be default validates if `test_results.product` is *present* - this is the message that we currently override with the locale file.
* we could use `validates_associated` if we wanted to make sure that `test_result.product` is *valid*. So technically an empty association wouldn't produce errors here

shared-web/app/views/layouts/shared/web/component_gallery.html.slim
jasiekmiko: agreed, I've documented it here: https://regulatorydelivery.atlassian.net/browse/MSPSDS-988

mspsds-web/app/views/application/_form_validation_display.html.slim
```diff
@@ -1,6 +1,19 @@
...
 - if errors&.any?
-  - error_messages = errors.details.keys.without(:day, :month, :year).map { |key| errors.full_messages_for(key).first }
+  ruby:
+    keys = errors.details.keys.without(:day, :month, :year)
```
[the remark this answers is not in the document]
jasiekmiko: I was tempted to clean this up and elected to not go down that particular rabbit hole, given that it'll be a much more natural fit when updating the date component to match :)

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
```diff
@@ -1,18 +1,45 @@
...
+      { text: user.display_name, value: user.id, checked: investigation.assignee == user }
+    end
+    if current_user.teams.any?
+      items.push text: "Someone in your team",
+                 value: "someone_in_your_team",
+                 conditional: { html: someone_in_your_team }
+    end
+    items.push text: "Previously assigned", value: "previously_assigned", conditional: { html: previously_assigned }
+    items.push divider: "Teams"
+    teams = investigation.important_assignable_teams.map do |team|
```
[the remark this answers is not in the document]
jasiekmiko: nice!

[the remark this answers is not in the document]
jasiekmiko: Oh, except we then have a an awkward set of parenthesis around the do-end block...
```
items.push(*investigation.important_assignable_teams.map do |team|
      { text: team.display_name, value: team.id, checked: investigation.assignee == team }
    end)
```
makes me think that's not worth it?

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
```diff
@@ -1,18 +1,45 @@
...
+            options: compute_select_options(entities: Team.all, ignore_visibility_restrictions: true),
+            accessible_hint: "Other team name"
 
-      .govuk-radios__divider or
-      = render "investigations/assign/someone_else_select", form: form
+  - someone_else = capture do
+    = render "investigations/assign/assignee_selection", form: form, key: :select_someone_else,
+            options: compute_select_options(entities: User.all),
+            accessible_hint: "Name"
+  ruby:
```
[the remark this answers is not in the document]
jasiekmiko: I think the captures might make it tricky. It could be something along the lines of 
```
  - if current_user.teams.any?
              options: compute_select_options(entities: User.get_team_members(user: current_user)),
              accessible_hint: "Name"
  - else
    - someone_in_your_team = nil
  - = capture do
    = render "investigations/assign/assignee_selection", form: form, key: :select_previously_assigned,
            options: compute_select_options(entities: investigation.past_assignees + investigation.past_teams,
                    ignore_visibility_restrictions: true),
            accessible_hint: "Select previously assigned person or team"
  - other_teams = capture do
    = render "investigations/assign/assignee_selection", form: form, key: :select_other_team,
            options: compute_select_options(entities: Team.all, ignore_visibility_restrictions: true),
            accessible_hint: "Other team name"

  - someone_else = capture do
    = render "investigations/assign/assignee_selection", form: form, key: :select_someone_else,
            options: compute_select_options(entities: User.all),
            accessible_hint: "Name"
  - items = assignee_options(someone_in_your_team, previously_assigned, other_teams,  someone_else)
```
where 'assignee_options` is a helper function, but I'm not sure that's actually that helpful.
It's all logic of putting the view together, as opposed to business logic, so I'm not massively concerned with it here..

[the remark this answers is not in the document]
broder: Fair enough!

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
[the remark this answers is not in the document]
broder: I think it still reads a little better even with the parentheses but it's fairly minor

PR #566, opened 2019-02-08
PSD-904: Submit latest coverage when not running all tests

broder:
At the minute, [our coverage](https://coveralls.io/github/UKGovernmentBEIS/beis-opss) looks like it fluctuates between the value for Cosmetics and the value for PSD.

I think it's because we only run the [Cosmetics|PSD] tests when we change [Cosmetics|PSD]. This PR should hopefully submit coverage for both apps, even if the tests aren't run.

I still need to test it though!


cosmetics-web/spec/spec_helper.rb
```diff
@@ -1,10 +1,8 @@
 require 'simplecov'
 require 'simplecov-console'
-require 'shared/web/coveralls_formatter'
 SimpleCov.formatters = [
     SimpleCov::Formatter::HTMLFormatter,
-    SimpleCov::Formatter::Console,
-    Shared::Web::CoverallsFormatter
```
jasiekmiko: How come this is no longer needed? It's just thanks to the module naming change below, right, we're still running the custom formatter?

ci/test-if-changed.sh
```diff
@@ -1,15 +1,17 @@
...
     docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm start_dependencies
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT \
-        bin/rake db:create db:schema:load test:all
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
+    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/rake db:create db:schema:load test:all
 else
     echo 'Testing not required.'
 fi
+
+docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/rake submit_coverage
```
jasiekmiko: I'm picturing this sending coverage twice on every Travis run (once for mspsds, once for cosmetics). Is that correct?

broder: Yeah, great spot!

broder: Wait, it will submit twice. But one of the submissions will be for cosmetics and one for MSPSDS (which is kind of what is already happening with the formatter running for each test suite). The MSPSDS container doesn't have the coverage for cosmetics and vice versa due to `- ~/coverage-cache/cosmetics:/cosmetics-web/coverage` in `docker-compose.ci.yml`

cosmetics-web/spec/spec_helper.rb
broder: SimpleCov generates an XML file which `submit_coverage` then reads and sends. Previously we were sending the coverage using the same in-memory object that's used to create the XML. Does that make sense?

PR #568, opened 2019-02-08
MSPSDS-982: Update documentation surrounding database setup

broder:
This doesn't need to go in before the production release, it's mostly just for future reference.


keycloak/README.md
```diff
@@ -42,7 +42,7 @@ To create a Keycloak database for the current space:
 
     cf marketplace -s postgres
     cf enable-service-access postgres
-    cf create-service postgres tiny-unencrypted-10.5 keycloak-database
+    cf create-service postgres small-10.5 keycloak-database
```
jasiekmiko: Just to clarify that we've moved all envs to this size?

mspsds-web/README.md
```diff
@@ -130,11 +128,8 @@ To create an Elasticsearch instance for the current space:
 To create a redis instance for the current space. 
 
     cf marketplace -s redis
-    cf create-service redis tiny-3.2 mspsds-redis-queue
-    cf create-service redis tiny-3.2 mspsds-redis-session
-
-Larger options should be considered if required.
-The current worker (sidekiq), which uses mspsds-redis-queue only works with the unclustered version.
```
jasiekmiko: do we know that the unclustered clause no longer apply?

keycloak/README.md
broder: We have now :)

mspsds-web/README.md
broder: No, I had assumed that they removed the option for unclustered instances on PaaS and therefore, it was working with a clustered version. `tiny-3.2` is a single node so shouldn't actually be doing any clustering. The other size available (`medium-ha-3.2`) does have multiple nodes so might be clustered, so we probably should keep the warning.

PR #573, opened 2019-02-08
MSPSDS-963: Notify relevant users

kicferk1:
Each activity has its update text, and can override the way it defines 'relevant users'. 

Should probably add a way to use different templates, for different users, which should hopefully be straightforward.


mspsds-web/app/policies/investigation_policy.rb
```diff
@@ -25,17 +25,17 @@ def visibility?(user: @user)
 
   def visible_to(user:, private: @record.is_private)
     return true unless private
-    return true if @record.assignee.present? && (@record.assignee&.organisation == user.organisation)
-    return true if @record.source.respond_to?(:user) && @record.source&.user&.present? && (@record.source&.user&.organisation == user.organisation)
     return true if user.is_opss?
+    return true if @record.assignee.present? && (@record.assignee&.organisation == user.organisation)
+    return true if @record.source&.user&.present? && (@record.source&.user&.organisation == user.organisation)
```
broder: If `source` was a `ReportSource`, would `source&.user&.present?` error? i.e.  do we need to keep `source.respond_to?(:user)`?

mspsds-web/app/controllers/investigations/activities_controller.rb
```diff
@@ -41,6 +41,7 @@ def comment; end
   def create
     respond_to do |format|
       if @activity.save
+        @activity.notify_relevant_users
```
broder: Should we call `notify_relevant_users` from an `after_save` callback or similar?

mspsds-web/app/models/audit_activity/investigation/update_assignee.rb
```diff
@@ -9,11 +9,32 @@ def subtitle_slug
...
+    # We store assignable_id in title field, this is computing title based on that
     "Assigned to #{(User.find_by(id: assignable_id) || Team.find_by(id: assignable_id))&.display_name}"
   end
+
+  def email_update_text
+    "#{investigation.case_type.titleize} was assigned to #{investigation.assignee.display_name} by #{source&.show&.titleize}."
+  end
+
+  def users_to_notify
+    activity = AuditActivity::Investigation::UpdateAssignee.where(investigation_id: investigation.id).order("created_at").last
```
broder: If we used a callback for `notify_relevant_users`, could we get this from something like `investigation.saved_changes`? https://stackoverflow.com/a/20657833

mspsds-web/app/mailers/notify_mailer.rb
```diff
@@ -23,4 +23,17 @@ def assigned_investigation_to_team(investigation_id, user_name, email, team_name
 
     mail(to: email)
   end
+
+  def updated_investigation(investigation_id, name, email, update_text)
+    set_template('10a5c3a6-9cc7-4edb-9536-37605e2c15ba')
+    set_reference('Case assigned')
```
broder: Is this reference always correct?

mspsds-web/app/policies/investigation_policy.rb
kicferk1: If source is a ReportSource, should we have a policy for displaying GDPR stuff, visibility stuff, assign stuff? 

Rather than adding an extra check somewhere in the middle of computation perhaps it would be best to condition the whole thing on source type. Except I'm not sure if we have any policies for ReportSource :/ 

I'll chat to Eds and try to find out.

mspsds-web/test/models/notification_test.rb
```diff
@@ -0,0 +1,147 @@
...
+    @user_two = User.find_by(last_name: "User_two")
+    @user_three = User.find_by(last_name: "User_three")
+  end
+
+  teardown do
+    logout
+  end
+
+  test "should notify current assignee when the assignee is a person and there is any change" do
+    @investigation.assignee = @user_one
```
broder: Could we do the `.assignee = ` and `.save` on one line using `.update`?

mspsds-web/test/test_helper.rb
```diff
@@ -74,30 +74,41 @@ def sign_in_as_user(is_admin: false, user_name: "User_one", organisation: organi
     stub_user_group_data(user_groups: user_groups, users: users)
     stub_user_data(users: users)
     stub_client_config
+    stub_notify_mailer
+  end
+
+  def stub_notify_mailer
+    result = ""
+    allow(result).to receive(:deliver_later)
+    allow(NotifyMailer).to receive(:assigned_investigation) { result }
```
broder: `NotifyMailer` only has `updated_investigation` now, right?

PR #574, opened 2019-02-08
COSBETA-24: read data from file

aidayounesi:
Unzipping the file

Loading the xml

Parsing the xml and extract the product name

Save the data

delete the uploaded file

Creating the drafts tab and showing an entry for the product


.idea/dataSources.xml
```diff
@@ -15,7 +15,7 @@
       <imported>true</imported>
       <remarks>$PROJECT_DIR$/config/database.yml</remarks>
       <jdbc-driver>org.postgresql.Driver</jdbc-driver>
-      <jdbc-url>jdbc:postgresql://localhost:5432/app_development</jdbc-url>
+      <jdbc-url>jdbc:postgresql://localhost:5433/cosmetics_development</jdbc-url>
```
broder: Should there be separate entries here for MSPSDS and cosmetics?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
+require 'zip'
+
+class ReadDataAnalyzer < ActiveStorage::Analyzer
+  def initialize(blob)
+    super(blob)
+  end
+
+  def self.accept?(given_blob)
+    if given_blob.present?
```
broder: Could do `return false unless given_blob.present?` to reduce a bit of nesting.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+      # this analyzer only accepts notification files which are zip
+      notification_file = ::NotificationFile.find_by(id: ActiveStorage::Attachment.find_by(blob_id: given_blob.id).record_id)
+      return true if notification_file.present?
```
broder: Then this would become `return notification_file.present?`

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    delete_notification_file
+    {}
+  end
+
+private
+
+  def create_notification_from_file
+    notification_file = ::NotificationFile.find_by(id: ActiveStorage::Attachment.find_by(blob_id: blob.id).record_id)
+    @notification = ::Notification.new(product_name: get_notification_current_name,
+                                       state: :draft_complete,
```
broder: Is `draft_complete` the right state for now, @DWRendell? Should it be something else once the images functionality goes in?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+class ReadDataAnalyzer < ActiveStorage::Analyzer
+  def initialize(blob)
+    super(blob)
+  end
+
+  def self.accept?(given_blob)
+    if given_blob.present?
+      # this analyzer only accepts notification files which are zip
+      notification_file = ::NotificationFile.find_by(id: ActiveStorage::Attachment.find_by(blob_id: given_blob.id).record_id)
```
broder: Can we get the attachment for a blob using `blob.attachments`? https://github.com/rails/rails/blob/94b5cd3a20edadd6f6b8cf0bdf1a4d4919df86cb/activestorage/app/models/active_storage/blob.rb#L32

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    notification_file = ::NotificationFile.find_by(id: ActiveStorage::Attachment.find_by(blob_id: blob.id).record_id)
+    @notification = ::Notification.new(product_name: get_notification_current_name,
+                                       state: :draft_complete,
+                                       responsible_person: notification_file.responsible_person)
+    @notification.save
+  end
+
+  def get_notification_current_name
+    get_xml_file do |xml_file|
+      xml_doc = Nokogiri::XML(xml_file.get_input_stream.read.gsub('sanco-xmlgate:', ''))
```
broder: Is this the correct was to deal with the `sanco-xmlgate` namespace?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  def get_xml_file
+    download_blob_to_tempfile do |file|
+      Zip::File.open(file.path) do |zip_file|
+        yield zip_file.glob(get_xml_file_name_regex).first
+      end
+    end
+  end
+
+  def get_xml_file_name_regex
+    blob.filename.base[0...8] + '*.xml'
```
broder: This assumes that the user won't have renamed the zip file, right? Can we get the file without assuming that?

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -5,6 +5,7 @@ def index
     @pending_notification_files = NotificationFile.where(
       responsible_person_id: @responsible_person.id, user_id: current_user.id
     ).count
+    @notifications = Notification.where(responsible_person_id: @responsible_person.id)
```
broder: Could we do this using `@reponsible_person.notifications`?

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
```diff
@@ -3,11 +3,12 @@
 RSpec.describe ResponsiblePersons::NotificationFilesController, type: :controller do
   let(:responsible_person) { create(:responsible_person) }
   let(:valid_attributes) {
-    { uploaded_file: fixture_file_upload("testImage.png", "image/png") }
+    { uploaded_file: fixture_file_upload("5D8F949A.zip") }
   }
 
   before do
     sign_in_as_member_of_responsible_person(responsible_person)
+    mock_antivirus
```
broder: We should probably unmock the antivirus in the `after` block.

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
```diff
@@ -26,7 +27,7 @@
       it "creates a new NotificationFile" do
         expect {
           post :create, params: { responsible_person_id: responsible_person.id, notification_file: valid_attributes }
-        }.to change(NotificationFile, :count).by(1)
+        }.to change(NotificationFile, :count).by(0)
```
broder: Is this really testing that `it "creates a new NotificationFile"`?

mspsds-web/Gemfile.lock
```diff
@@ -446,7 +446,6 @@ DEPENDENCIES
   redis-rails
   roo (~> 2.8.2)
   rspec-mocks
-  rubyzip (>= 1.2.1)
```
broder: Does MSPSDS still need this?

cosmetics-web/app/analyzers/read_data_analyzer.rb
DWRendell: Yes, it should be a new state, probably something along the lines of `:notification_file_imported`?

cosmetics-web/app/analyzers/read_data_analyzer.rb
DWRendell: I think we might be stuck with something like this, just because the XML files in the export are technically invalid and Nokogiri refuses to work with them. I think the problem is that the namespace isn't properly defined.

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Sure, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Right thanks

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
aidayounesi: Change to "a new NotificationFile is removed after processing"

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
aidayounesi: Right thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Yes, this is the assumption.
There are several xml files in the zip file. Based on the sample zip file that I got from @DWRendell , it is the only xml file that its name starts with a number, is it always the case? or that it has at least one digit number in it's name first 4 characters?
We can iterate over them and choose the one that contains e.g. "<sanco-xmlgate:product>".

cosmetics-web/spec/notification_file_helpers.rb
```diff
@@ -0,0 +1,5 @@
+module NotificationFileHelpers
+  def mock_antivirus
+    allow(Clamby).to receive(:safe?).and_return(true)
```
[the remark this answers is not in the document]
aidayounesi: 👍

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Right. Would it be good enough to use 'first' one of attachments?

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
aidayounesi: Sure, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
DWRendell: The numbers are hex digits, so it is possible they won't have a number in the first 4 characters.

It should be a file with a name like `54450A8F-E3F1-B8D0-5ABE800DCF71C3E4-B218491E348A8F795CE5460E94E8FE5F_20190211`, i.e. 8 hex digits, followed by 4, 4, 16 and 32, then a 8 digit date string. So maybe some regex something like
```
[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{16}-[0-9A-F]{32}_[0-9]{8}.xml
```
?

.idea/dataSources.xml
aidayounesi: Sure. Something like that is working on my pc.

```

    <data-source source="LOCAL" name="Local Dev DB Cosmetics">
      <driver-ref>postgresql</driver-ref>
      <synchronize>true</synchronize>
      <imported>true</imported>
      <remarks>$PROJECT_DIR$/config/database.yml</remarks>
      <jdbc-driver>org.postgresql.Driver</jdbc-driver>
      <jdbc-url>jdbc:postgresql://localhost:5433/cosmetics_development</jdbc-url>
    </data-source>
    <data-source source="LOCAL" name="Local Dev DB MSPSDS">
      <driver-ref>postgresql</driver-ref>
      <synchronize>true</synchronize>
      <imported>true</imported>
      <remarks>$PROJECT_DIR$/config/database.yml</remarks>
      <jdbc-driver>org.postgresql.Driver</jdbc-driver>
      <jdbc-url>jdbc:postgresql://localhost:5433/mspsds_development</jdbc-url>
    </data-source>
```

cosmetics-web/app/analyzers/read_data_analyzer.rb
tom-bell-softwire: I wonder if it's worth just matching the first UUID part of the filename, to be a bit more flexible?

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
```diff
@@ -3,11 +3,12 @@
 RSpec.describe ResponsiblePersons::NotificationFilesController, type: :controller do
   let(:responsible_person) { create(:responsible_person) }
   let(:valid_attributes) {
-    { uploaded_file: fixture_file_upload("testImage.png", "image/png") }
+    { uploaded_file: fixture_file_upload("5D8F949A.zip") }
```
tom-bell-softwire: Would be good to call this file something like `testExportFile.zip`

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -13,9 +14,25 @@
...
     visit new_responsible_person_notification_file_path(responsible_person.id)
-    page.attach_file('uploaded_file', Rails.root + 'spec/fixtures/testImage.png')
+    page.attach_file('uploaded_file', Rails.root + 'spec/fixtures/5D8F949A.zip')
     click_button "Upload"
 
     expect(page).to have_text("Your cosmetic products")
   end
+
+  it "set a notification name in dashboard based on the uploaded file" do
+    visit new_responsible_person_notification_file_path(responsible_person.id)
```
tom-bell-softwire: I think you can just pass in `responsible_person`, rather than its ID.

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -13,9 +14,25 @@
...
+                     Rails.root + 'spec/fixtures/5D8F949A.zip')
+    click_button "Upload"
+
+    5.times do
+      break if not page.text.include? "Refresh the browser"
+
+      visit responsible_person_notifications_path(responsible_person.id)
+    end
+
+    page.find_by_id("tab_unfinished").click
```
tom-bell-softwire: Does the following work?
```
click_link "tab_unfinished"
```

mspsds-web/Gemfile.lock
aidayounesi: Sure, added again, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
broder: I'd hope that there would be something within Nokogiri to deal with the namespaces.

Otherwise, I suspect Tom's solution of declaring the namespace would be faster than removing all instances of the namespace from the entire file. I think we should be considering performance in this analyser as much as possible given that (we think) it'll be used a lot.

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: ```:notification_file_imported``` state is added. For now the previous state of this state is ```:empty```

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Thanks.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+    @notification = ::Notification.new(product_name: get_notification_current_name,
+                                       state: :draft_complete,
+                                       responsible_person: notification_file.responsible_person)
+    @notification.save
+  end
+
+  def get_notification_current_name
+    get_xml_file do |xml_file|
+      xml_doc = Nokogiri::XML(xml_file.get_input_stream.read.gsub('sanco-xmlgate:', ''))
+      notification_current_name = xml_doc.
```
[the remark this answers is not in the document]
aidayounesi: Sure, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,59 @@
...
+  def metadata
+    create_notification_from_file
+    delete_notification_file
+    {}
+  end
+
+private
+
+  def create_notification_from_file
+    notification_file = ::NotificationFile.find_by(id: ActiveStorage::Attachment.find_by(blob_id: blob.id).record_id)
```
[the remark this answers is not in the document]
aidayounesi: sure, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Thank you all. Combination of Dave and Tom answer matching ``` /[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{16}(.*)xml/```

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
aidayounesi: Sure, thanks

cosmetics-web/spec/system/upload_single_file_spec.rb
aidayounesi: Sure, thanks

cosmetics-web/spec/system/upload_single_file_spec.rb
aidayounesi: Thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: https://nokogiri.org/tutorials/searching_a_xml_html_document.html
''But I'm Lazy and Don't Want to Deal With Namespaces!¶
Lazy == Efficient, so no judgements. :)

If you have an XML document with namespaces, but would prefer to ignore them entirely (and query as if Tim Bray had never invented them), then you can call remove_namespaces! on an XML::Document to remove all namespaces. Of course, if the document had nodes with the same names but different namespaces, they will now be ambiguous. But you're lazy! You don't care!''

aidayounesi: Is it a good idea  to insert ```xmlns:sanco-xmlgate="http://sawes.ws.in.xmlgatev2.sanco.cec.eu"``` in a fixed position (72 which is after ```sanco-xmlgate:product``` tag) and then use remove_namespaces!

Because otherwise the xpath query would be something like this:
``` return xml_doc.xpath('//sanco-xmlgate:currentVersion/sanco-xmlgate:generalInfo/sanco-xmlgate:productNameList/sanco-xmlgate:productName/sanco-xmlgate:name',
                           'sanco-xmlgate' => "http://sawes.ws.in.xmlgatev2.sanco.cec.eu").first.text```

Comment (aidayounesi) [reply cosmetics-web/app/analyzers/read_data_analyzer.rb]: @tom-bell-softwire @broder

Comment (aidayounesi) [reply cosmetics-web/app/analyzers/read_data_analyzer.rb]:
```
get_xml_file_content do |xml_file_content|
      xml_doc = Nokogiri::XML(xml_file_content.insert(71, ' xmlns:sanco-xmlgate="http://sawes.ws.in.xmlgatev2.sanco.cec.eu"'))
      return xml_doc.xpath('//sanco-xmlgate:currentVersion/sanco-xmlgate:generalInfo/sanco-xmlgate:productNameList/sanco-xmlgate:productName/sanco-xmlgate:name',
                           'sanco-xmlgate' => "http://sawes.ws.in.xmlgatev2.sanco.cec.eu").first.text
    end
```

File: cosmetics-web/app/models/notification.rb [hunk 18]
```diff
@@ -32,6 +33,10 @@ class Notification < ApplicationRecord
     event :submit_notification do
       transitions from: :draft_complete, to: :notification_complete
     end
+
+    event :notification_file_parsed do
+      transitions from: :empty, to: :notification_file_imported
```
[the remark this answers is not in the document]
DWRendell: I think it probably makes sense to have a separate state for the imported notifications, it sounds like there could be some changes coming to what is required for bulk uploads compared to the manual journey.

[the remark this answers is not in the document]
broder: Fair enough!

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
aidayounesi: Good point. Not always. Should I expect that either NotificationFile count is changed by 1 OR Notification count?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -1,17 +1,20 @@
 require 'zip'
 
 class ReadDataAnalyzer < ActiveStorage::Analyzer
+  include AnalyzerHelper
+  extend AnalyzerHelper
```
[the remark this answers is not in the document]
aidayounesi: I think so because it's called in both a class method and an instance method.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -23,32 +26,35 @@ def metadata
...
   end
 
   def get_notification_current_name
-    get_xml_file do |xml_file|
-      xml_doc = Nokogiri::XML(xml_file.get_input_stream.read.gsub('sanco-xmlgate:', ''))
-      notification_current_name = xml_doc.
-          xpath('//currentVersion/generalInfo/productNameList/productName/name').first.text
-      notification_current_name
+    get_xml_file_content do |xml_file_content|
+      xml_doc = Nokogiri::XML(xml_file_content.insert(71, ' xmlns:sanco-xmlgate="http://sawes.ws.in.xmlgatev2.sanco.cec.eu"'))
```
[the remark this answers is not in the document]
tom-bell-softwire: This Stack Overflow post looks like it could be relevant:
https://stackoverflow.com/questions/7961883/calling-supers-method-add-namespace-to-nokogiri-xml-document

It looks like specifying the namespace in the xpath query might be the simplest way to handle it, rather than try to modify the content of the file itself. It's makes the queries a bit more ugly (though we can probably wrap in a helper method) and I don't know if that impacts the performance of the xpath query, but might be more robust?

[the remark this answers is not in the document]
aidayounesi: The main problem is that if we ask Nokogiri to parse the text (without adding namespace manually or removing all of prefix) then Nokogiti::XML(content) would parse the whole xml as just a single node! Of course we can add the namespace to this single node, but then we need to parse it again to get all the nodes. So, I think we have only 2 options:
1. Manually, ruby and regex, adding the namespace and then Nokogiri parse. 
2. Nokogiri parse it add_namespace and parse again. I think DBR solution and Tom's link solution are both in this category. 

If we care about time-complexity the first option is faster.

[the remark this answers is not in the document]
aidayounesi: Some tests, both approach returns nil:

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
broder: Fair enough - could we use `self.get_notification_file_from_blob` on the instance method and then just have `extend`?

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
broder: Ah, I hadn't realised that it was interpreting the incorrect XML as a single node - in which case maybe using `doc = Nokogiri::XML(doc_without_ns.to_xml)` isn't actually too bad? I was assuming that it would have to recurse through a lot of nodes to re-produce the XML...

It's still not ideal though - would it be worth running some quick tests to see how fast each approach (replacing all instances of the namespace; inserting the namespace; parsing, fixing and correcting the XML using Nokogiri) actually is?

RE: using regex, we might be able to get away with it but [it's generally frowned upon](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454)

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
aidayounesi: ``` NoMethodError: undefined method `get_notification_file_from_blob' for #<ReadDataAnalyzer:0x000055fee2fc3a28>```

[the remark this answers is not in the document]
tom-bell-softwire: I wonder if it might just be simpler to move that method to a private method within this class? Is it likely to be used elsewhere, or is it specific to the logic in this analyzer?

(I realise I suggested making this a helper method in a previous comment, which may have overly complicated the issue!)

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  def get_xml_file_name_regex
+    /[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{16}.*\.xml/
+  end
+
+  def delete_notification_file
+    attachment = ActiveStorage::Attachment.find_by(blob_id: blob.id)
+    notification_file = ::NotificationFile.find_by(id: attachment.record_id)
```
tom-bell-softwire: Could you also use the `get_notification_file_from_blob` method here?

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
tom-bell-softwire: What is it we're testing here?

If it's that the controller action successfully creates a NotificationFile, then I think we should be asserting that the count is changed by 1. In which case, would it make sense to mock the analyzer in the same way that you've mocked Clamby, to prevent it from processing and removing the NotificationFile?

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
aidayounesi: Some sort of benchmark:
T = average runtime of a piece of code (run 100 times for test).

T=0.0007355590000000002
```
    xml_file_content = xml_file_content.sub('<sanco-xmlgate:product>',
                                            '<sanco-xmlgate:product xmlns:sanco-xmlgate="http://sawes.ws.in.xmlgatev2.sanco.cec.eu">')
    xml_doc = Nokogiri::XML(xml_file_content)
    xml_doc.xpath('//sanco-xmlgate:currentVersion/sanco-xmlgate:generalInfo/sanco-xmlgate:productNameList/sanco-xmlgate:productName/sanco-xmlgate:name',
                         'sanco-xmlgate' => "http://sawes.ws.in.xmlgatev2.sanco.cec.eu").first.text
```

T = 0.003012679999999998
```
    xml_doc = Nokogiri::XML(xml_file_content)
    xml_doc.root.add_namespace('sanco-xmlgate', 'http://sawes.ws.in.xmlgatev2.sanco.cec.eu')
    xml_doc = Nokogiri::XML(xml_doc.to_xml)
    xml_doc.xpath('//sanco-xmlgate:currentVersion/sanco-xmlgate:generalInfo/sanco-xmlgate:productNameList/sanco-xmlgate:productName/sanco-xmlgate:name',
                         'sanco-xmlgate' => "http://sawes.ws.in.xmlgatev2.sanco.cec.eu").first.text
```

T=0.0005453340000000003
```
    xml_file_content.gsub!('sanco-xmlgate:', '')
    xml_doc = Nokogiri::XML(xml_file_content)
    xml_doc.xpath('//currentVersion/generalInfo/productNameList/productName/name').first.text
```

T=0.0007140420000000001
```
    xml_file_content = xml_file_content.sub('<sanco-xmlgate:product>',
                                            '<sanco-xmlgate:product xmlns:sanco-xmlgate="http://sawes.ws.in.xmlgatev2.sanco.cec.eu">')
    xml_doc = Nokogiri::XML(xml_file_content)
    xml_doc.remove_namespaces!
    xml_doc.xpath('//currentVersion/generalInfo/productNameList/productName/name').first.text
```

cosmetics-web/app/analyzers/read_data_analyzer.rb
aidayounesi: Definitely, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
aidayounesi: Thanks, makes sense

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
aidayounesi: Thanks, it makes sense

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
broder: Should we go for `gsub!` for now then? I'd be tempted to revisit this again in the future once we have more example data that we can run these tests on - maybe raise a ticket for it?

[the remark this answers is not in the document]
aidayounesi: Sure, thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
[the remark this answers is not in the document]
aidayounesi: Solved by keeping it in a helper method.
In ReadDataAnalayzer: ```extend AnalyzerHelper```
Call it get_notification_file_from_blob from class methods and self.class.get_notification_file_from_blob from instance methods

cosmetics-web/app/views/responsible_persons/notifications/_unfinished_notification.html.slim
```diff
@@ -0,0 +1,26 @@
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    .product
+      h2
+        = link_to notification.product_name, "/", class: "govuk-link--no-visited-state"
+      .govuk-body-s.product-data
+        dt
+          | UK cosmetic product number:
+        dd
+          | UKCP-78424
```
broder: Could we use `notification.reference_number` here? Might be another ticket though!

aidayounesi: It's part of my next ticket

PR #579, opened 2019-02-11
COSBETA-38: view drafts and submitted products

raamSoftwire:


cosmetics-web/app/policies/application_policy.rb
```diff
@@ -0,0 +1,53 @@
+class ApplicationPolicy
```
raamSoftwire: I've copied this file from MSPSDS, not sure if we need all of these methods.

cosmetics-web/app/views/errors/forbidden.html.slim
```diff
@@ -0,0 +1,7 @@
+- content_for :page_title, "Access denied"
+
```
raamSoftwire: I have removed the back link that existed on the corresponding MSPSDS page because I don't think it is obvious where we should send them back to.

tom-bell-softwire: Good call!

cosmetics-web/spec/responsible_person_helpers.rb
```diff
@@ -4,4 +4,9 @@ def sign_in_as_member_of_responsible_person(responsible_person)
     responsible_person.add_user(user)
     sign_in as_user: user
   end
+
+  def sign_in_as_member_of_responsible_person_as_user(responsible_person, user)
```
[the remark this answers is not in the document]
tom-bell-softwire: I wonder if the correct approach here would be to use the existing `sign_in as_user:` method and to add factories to create the relevant user/RP associations for each scenario.

cosmetics-web/spec/system/responsible_person_dashboard_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'rails_helper'
+
+RSpec.describe "Responsible person dashboard", type: :system do
+  let(:user_1) { create(:user) }
+  let(:user_2) { create(:user) }
+  let(:responsible_person_1) { create(:responsible_person) }
```
tom-bell-softwire: You could either explicitly add `user_1` to the RP in this block, or create a factory method that associates a user with a responsible person ([see here](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations) for how to add associations).

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -14,4 +17,8 @@ def create_or_join_responsible_person
 
     redirect_to create_or_join_existing_account_index_path if current_user.responsible_persons.empty?
   end
+
+  def forbidden
+    redirect_to '/403'
```
broder: I wonder if we should render the error and preserve the users URL, rather than redirecting them to `/403` (this functionality is a currently outstanding bug on MSPSDS)

cosmetics-web/spec/responsible_person_helpers.rb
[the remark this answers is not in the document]
raamSoftwire: Done, thanks

cosmetics-web/spec/system/responsible_person_dashboard_spec.rb
raamSoftwire: I've now changed that, thanks

cosmetics-web/app/controllers/application_controller.rb
raamSoftwire: Done, thanks

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -14,4 +17,8 @@ def create_or_join_responsible_person
 
     redirect_to create_or_join_existing_account_index_path if current_user.responsible_persons.empty?
   end
+
+  def forbidden
+    render "errors/forbidden"
```
broder: Does this also return the correct status code? (e.g. in `errors_controller`, we're passing `status: :forbidden` to the `render` method)

Thinking about it, we may actually want to handle this by using the `config` bit from https://github.com/varvet/pundit/blob/740cbffd2ff7c529d19f57696da75aca3f3152ab/README.md#rescuing-a-denied-authorization-in-rails

cosmetics-web/app/controllers/responsible_persons/team_members_controller.rb
```diff
@@ -5,5 +5,6 @@ class ResponsiblePersons::TeamMembersController < ApplicationController
 
   def set_responsible_person
     @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
+    authorize @responsible_person, :index?
```
broder: Do we need to specify `:index?` in these other places as well?

cosmetics-web/app/policies/responsible_person_policy.rb
```diff
@@ -0,0 +1,9 @@
+class ResponsiblePersonPolicy < ApplicationPolicy
+  def show?
```
broder: Do we need to distinguish between `index?` and `show?` here?

Generally, `index` would be used for listing all responsible persons so I wouldn't expect `@record` to be set. `show` is used for showing a single responsible person so I think defining `show?` as you have done makes sense. This will then probably have a knock-on effect for how you use the `authorize` method elsewhere.

Happy to chat through it!

cosmetics-web/app/policies/responsible_person/notification_policy.rb
```diff
@@ -0,0 +1,2 @@
+class ResponsiblePerson::NotificationPolicy < ResponsiblePersonPolicy
```
broder: I'm not sure whether we need these files. AFAIK, Pundit chooses the policy based on the model that you pass it (i.e. `@repsonsible_person`) rather than the controller.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -1,15 +1,20 @@
 class ResponsiblePersons::NotificationsController < ApplicationController
+  include ResponsiblePersons::NotificationsHelper
```
tom-bell-softwire: Could this be a concern, rather than a helper? [Generally speaking](https://stackoverflow.com/questions/18751027/whats-the-difference-between-a-controller-helper-and-a-concern), helpers should be used for presentation-related methods (such as formatting model properties for display), while concerns are modules to pull out common logic from controllers (or models).

broder: It's not being used in more than one place so should we just keep it in the controller for now?

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
```diff
@@ -20,7 +17,9 @@
       | Your cosmetic products
     p
       b = @responsible_person.name
-      |  has no cosmetic products registered in the UK.
```
tom-bell-softwire: Worth checking with Ralph, but from looking at the prototype it seems like all this explanatory text (these three `<p>` blocks) is removed once the RP has registered products.

raamSoftwire: That's a really good spot! I guess that's not the simple copy change I originally thought. I've made that change now.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
raamSoftwire: I've now moved the methods into the controller.

PR #580, opened 2019-02-11
MSPSDS-566 fix js errors

MatthewBurstein:


mspsds-web/README.md
```diff
@@ -85,7 +85,7 @@ You can run the Slim linting with `docker-compose exec mspsds-web bin/slim-lint
 
 You can run the Sass linting with `docker-compose exec mspsds-web yarn sass-lint -vq -c vendor/shared-web/.sasslint.yml 'app/assets/stylesheets/**/*.scss'`.
 
-You can run the JavaScript linting with `docker-compose exec mspsds-web yarn eslint -c vendor/shared-web/.eslintrc.yml app/assets/javascripts`.
+You can run the JavaScript linting with `docker-compose exec mspsds-web yarn eslint -c vendor/shared-web/.eslintrc.yml app/assets/application/javascripts`.
```
broder: Ah, good spot! [The command that we're running on Travis has a slightly different format](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/ci/static-analysis-if-changed.sh#L11) - it might be good to keep them in sync. We've also changed the commands for the sass and slim linting if you want to update those as well.

mspsds-web/app/views/investigations/ts_investigations/which_businesses.html.slim
```diff
@@ -1,4 +1,5 @@
 - content_for :page_title, "Report an unsafe or non-compliant product"
+#which-businesses-page
```
broder: We could put this ID on the existing `.govuk-grid-row` `div` below rather than creating a new one.

mspsds-web/README.md
MatthewBurstein: I've had a go at changing these commands but wasn't sure what to do about the references to vendor. Do they need updating too?

PR #582, opened 2019-02-11
COSBETA-18 Add unique UK cosmetics product reference number to submitted notifications

DWRendell:
- add `reference_number` column to notifications
- randomly generate reference number on creation of a notification
- display the reference number with a `UKCP-` prefix on check your answers and confirmation pages
- use the reference number in the URLs for the notification

### DIFF
--- a/cosmetics-web/app/controllers/notification_build_controller.rb
+++ b/cosmetics-web/app/controllers/notification_build_controller.rb
@@ -43,7 +43,7 @@ class NotificationBuildController < ApplicationController
   end
 
   def new
-    redirect_to wizard_path(steps.first, notification_id: @notification.id)
+    redirect_to wizard_path(steps.first, notification_id: @notification.reference_number)
   end
 
   def finish_wizard_path
@@ -63,7 +63,7 @@ private
   end
 
   def set_notification
-    @notification = Notification.find(params[:notification_id])
+    @notification = Notification.where(reference_number: params[:notification_id]).first
   end
 
   def set_countries
--- a/cosmetics-web/app/controllers/notification_build_controller.rb
+++ b/cosmetics-web/app/controllers/notification_build_controller.rb
@@ -63,7 +63,7 @@ private
   end
 
   def set_notification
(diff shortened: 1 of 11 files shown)


cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -63,7 +63,7 @@ def notification_params
   end
 
   def set_notification
-    @notification = Notification.find(params[:notification_id])
+    @notification = Notification.where(reference_number: params[:notification_id]).first
```
broder: Would https://api.rubyonrails.org/v5.2.2/classes/ActiveRecord/FinderMethods.html#method-i-find_by be better here?

cosmetics-web/app/models/notification.rb
```diff
@@ -8,6 +8,15 @@ class Notification < ApplicationRecord
 
   accepts_nested_attributes_for :image_uploads
 
+  before_create do
+    new_reference_number = nil
+    loop do
+      new_reference_number = SecureRandom.rand(100000000)
```
broder: Will we ever want to left-pad the number for display or for the URL?

DWRendell: I've left padded the number in the reference_number_for_display method, but haven't for the URL routing (also haven't added the `UKCP-` prefix for the routing).

I think it will make the routing quite a bit more complicated to left pad the URL or add the prefix, but maybe I'm missing some rails magic?

cosmetics-web/app/controllers/notification_build_controller.rb
DWRendell: That's much nicer, thanks

cosmetics-web/app/models/notification.rb
broder: I think you could just trim the leading zeros on the parameter to regenerate the original ID. Having the zeroes in the URL would probably look a bit silly though... Either way, leave it as it is for now - it should be straightforward to change going forwards!

PR #587, opened 2019-02-12

kicferk1:
Some solutions to this problem on the internet suggest storing pretty_id(or its equivalent) in the database, and searching by it. This is probably reasonable, given that eventually we might also want to change the format to not be easily reversible into id. 

So a todo is to store pretty_id in our database, replace most finds with find_by, and remove the reverse pretty id method, unless it's something that falls under the 'use different format' ticket.


broder:
Storing (and indexing) it in the database and then using `find_by` is a good idea! We should do it now.

Is there a separate ticket to update the format to be `YYMM-XXXX`? Would it be sensible to merge the two?

mspsds-web/app/models/investigation.rb
```diff
@@ -57,7 +59,9 @@ class Investigation < ApplicationRecord
   has_one :source, as: :sourceable, dependent: :destroy
   has_one :complainant, dependent: :destroy
 
-  before_create :assign_current_user_to_case
+  after_initialize
```
broder: Is this doing anything?

mspsds-web/app/models/investigation.rb
```diff
@@ -189,6 +188,19 @@ def add_business(business, relationship)
     create_audit_activity_for_business(business)
   end
 
+  def to_param
+    pretty_id
+  end
+
+  def self.next_pretty_id(id: Investigation.this_month.count)
+    month_count = id.to_s.rjust(4, '0')
+    "#{Time.zone.now.strftime('%y%m')}-#{month_count}"
```
broder: Should we be using the `created_at` attribute on the investigation rather than the current time?

mspsds-web/test/models/investigation_test.rb
```diff
@@ -249,6 +249,20 @@ class InvestigationTest < ActiveSupport::TestCase
     assert policy(investigation).assign?(user: User.find_by(last_name: "Admin"))
   end
 
+  test "pretty_id should contein YYMM" do
```
broder: There's a typo in "contain"

mspsds-web/app/models/investigation.rb
```diff
@@ -189,6 +186,16 @@ def add_business(business, relationship)
     create_audit_activity_for_business(business)
   end
 
+  def to_param
+    pretty_id
+  end
+
+  def add_pretty_id(id: Investigation.this_month.where.not(pretty_id: nil).count)
+    month_count = id.to_s.rjust(4, '0')
+    pretty_id = "#{created_at.strftime('%y%m')}-#{month_count}"
```
broder: We could also do this with `self.pretty_id = "#{created_at.strftime('%y%m')}-%04d" % id` but it's functionally the same

PR #592, opened 2019-02-13
MSPSDS-566 markups

MatthewBurstein:
Depends on outcome of discussion in Jira about users without organisations.


mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -421,8 +421,12 @@ def records_saved?
   def save_businesses
     session[:businesses].each do |session_business|
       business = Business.create!(session_business[:business])
-      business.contacts << Contact.create!(session_business[:contact].merge(business_id: business.id))
-      business.locations << Location.create!(session_business[:location].merge(business_id: business.id))
+      if session_business[:contact]
+        business.contacts << Contact.create!(session_business[:contact].merge(business_id: business.id))
```
broder: Does it work if you use `.new` here rather than `.create`? Then you might not need to `merge` the `business_id` (and reduce the number of queries?)

mspsds-web/app/models/investigation.rb
```diff
@@ -175,6 +175,12 @@ def title; end
 
   def case_type; end
 
+  def reason_created
+    "Product reported because it is unsafe and non-compliant." if hazard_type && non_compliant_reason
```
broder: Do we need to `return` on these lines? Both 1 & 2 or 1 & 3 could be hit on a single execution

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
MatthewBurstein: Yes this works :)

PR #593, opened 2019-02-14
MSPSDS-6: Various tech debt

broder:


.idea/dataSources.xml
```diff
@@ -1,29 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
broder: Should we be source controlling these files?

cosmetics-web/Gemfile.lock
```diff
@@ -6,7 +6,7 @@ PATH
       capybara (= 3.13.2)
       coveralls (= 0.8.22)
       debase (= 0.2.2)
-      govuk-lint (= 3.10.0)
+      govuk-lint (= 3.11.0)
```
broder: When dependabot updates a gem in `shared-web.gemspec`, we need to update the `Gemfile.lock`s in Cosmetics and MSPSDS ourselves (by running e.g. `docker-compose restart cosmetics-web`). It's not great but I couldn't figure out how to get dependabot to do it all for us.

maintenance/package.json
```diff
@@ -9,10 +9,10 @@
     "start": "node server.js"
   },
   "dependencies": {
-    "express": "^4.16.4",
-    "govuk-frontend": "^2.5.1",
-    "html5shiv": "^3.7.3",
-    "node-sass-middleware": "^0.11.0",
-    "nunjucks": "^3.1.7"
+    "express": "4.16.4",
```
broder: I'm hoping that this will trigger dependabot to update our npm dependencies - maybe the `^` was being too loose? 

It's kind of similar to the change I made to the `shared-web.gemspec` here: https://github.com/UKGovernmentBEIS/beis-mspsds/commit/fbe9254be05a070f3b2edb423812fd40d8137821#diff-eb65ebabb15935a975b3d2705570212a

cosmetics-web/Gemfile.lock
tom-bell-softwire: Thanks, I'll watch out for this in future.

PR #596, opened 2019-02-14
MSPSDS-963: Notify relevant users

kicferk1:
Two changes: 
1. Only notify current assignee(user) about changes to case if someone else made them, 
2. Don't save comment activity twice. Turns out `investigatuion.activities << @activity` will save `@activity` in the background, which used to send duplicate email.


mspsds-web/app/controllers/investigations/activities_controller.rb
```diff
@@ -40,7 +40,7 @@ def comment; end
   # POST /activities.json
   def create
     respond_to do |format|
-      if @activity.save
+      if @investigation.activities << @activity
```
broder: Does `<<` return a bool?

kicferk1: No, it returns list of activities if successful, and nil if not

PR #598, opened 2019-02-14
MSPSDS-888 Add error summary component and move error messages closer to govuk style

jasiekmiko:


shared-web/app/views/components/_govuk_error_summary.html.slim
```diff
@@ -0,0 +1,26 @@
...
+        "class": local_assigns[:classes],
+        "aria-labelby": "error-summary-title",
+        role: "alert",
+        tabindex: "-1" }.merge(local_assigns[:attributes] || {})
+.govuk-error-summary *attributes
+  h2.govuk-error-summary__title#error-summary-title
+    == local_assigns[:titleHtml] || local_assigns[:titleText]
+  .govuk-error-summary__body
+    - if local_assigns[:descriptionHtml] || local_assigns[:descriptionText]
+      p = local_assigns[:descriptionHtml] || local_assigns[:descriptionText]
```
broder: Should this be `==`?

Also, I'm feeling a little bit nervous about adding opportunities for HTML injection. I think we've been fine so far but maybe for some of the next components where we're more likely to pass user generated content.

shared-web/app/views/form_components/_govuk_error_summary.html.slim
```diff
@@ -0,0 +1,13 @@
+/ This partial acts as glue between rails form helpers and the component govuk_error_summary
+/ With access to the form object, we can infer a lot of information from it
+
+ruby:
+  errors = local_assigns[:errors] || form.object.errors
+  error_list = errors.details.keys.flat_map do |attribute|
+    # Infer name from attribute key
+    # TODO commonize with radios
```
broder: What does this `TODO` involve?

jasiekmiko: A helper function - incoming! :)

shared-web/app/views/components/_govuk_error_summary.html.slim
jasiekmiko: ah yes, thanks!

Yeah, it's an interesting one - I'm obviously following what the nunjucks macros are doing, but I can see that we could accidentally leave a gap somewhere unintended.

Two things that we can do to lower the surface area of these (but I think it'll be a good idea to keep things consistent across the components):

A) Limit escaping to `html` attributes only. I initially implemented this as
```
p 
  - if local_assigns[:descriptionHtml]
    == local_assigns[:descriptionHtml]
  - else
    = local_assigns[:descriptionText]
```
but it causes the linter to complain that it can be shorthanded with `||`. We'd need to disable that cop I reckon.

B) Stop escaping altogether.
This wouldn't actually impact the way we're using the components so far, since html generated using `capture` can be displayed with just a `=`. It does veer away from the original implementation that allows one to put in html text in that field.
This solution would render the commented-on line as correct :)
(The only place that would need to change to accommodate this would be the example page, see https://github.com/UKGovernmentBEIS/beis-mspsds/blob/9ecd6518c4c1ff5bdbdc1ded974d9d1d4db318f4/shared-web/app/views/components_gallery/radios.html.slim#L50)

jasiekmiko: As agreed, I've removed the unescaped html now and provided a note in the readme re the reasoning

PR #607, opened 2019-02-15
MSPSDS-877: Use new way to compute pretty ids

kicferk1:
Old way is bugged and results in the same ids for many cases. This should re-generate pretty_ids and since the computation is ordered by created_at they should be different.


mspsds-web/db/migrate/20190215121322_fix_incorrect_pretty_ids.rb
```diff
@@ -0,0 +1,5 @@
+class FixIncorrectPrettyIds < ActiveRecord::Migration[5.2]
+  def change
+    Investigation.all.each(&:add_pretty_id)
```
broder: Should this be `.in_batches` rather than `.all`?

PR #609, opened 2019-02-15
COSBETA-105: Validate Responsible Person email address

DWRendell:
On creating a responsible person:

- Generate a random key for this responsible person
    - Set an expiry date on this key, and add a scheduled task to delete expired keys
- Send the responsible person email a link containing this key
- On clicking this link, Set a flag `is_email_verified` on the responsible person to `true`

Does not include locking access to the site until the address is verified.


cosmetics-web/spec/jobs/remove_expired_email_verification_keys_job_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe RemoveExpiredEmailVerificationKeysJob, type: :job do
+  let(:email_verification_key) { create(:email_verification_key) }
+  let(:expired_email_verification_key) { create(:expired_email_verification_key) }
+
+  before do
+    email_verification_key.save
```
broder: Does the `create` FactoryBot method not also save the records?

DWRendell: Apparently not :/ I thought it did, but without this the test fails, and says there aren't any keys in the database. Unless there's something else going on?

DWRendell: Apparently not :/ I thought it did, but without this the test fails, and says there aren't any keys in the database. Unless there's something else going on?

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -1,7 +1,7 @@
 class ResponsiblePersons::AccountWizardController < ApplicationController
   include Wicked::Wizard
 
-  steps :select_type, :enter_details
+  steps :select_type, :enter_details, :verify_email
```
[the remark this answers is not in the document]
DWRendell: That does sound like a good idea, done

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+    email_verification_key = EmailVerificationKey.find_by(
+      responsible_person_id: params[:responsible_person_id],
+        key: params[:key]
+    )
+
+    if email_verification_key.nil? || email_verification_key.is_expired?
+      redirect_to "/404"
+    else
+      email_verification_key.responsible_person.is_email_verified = true
+      email_verification_key.responsible_person.save
```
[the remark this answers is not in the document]
DWRendell: Nice, done

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,16 @@
+class ResponsiblePersons::VerificationController < ApplicationController
+  def show
+    email_verification_key = EmailVerificationKey.find_by(
```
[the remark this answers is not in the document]
DWRendell: That's way nicer, done

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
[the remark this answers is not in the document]
broder: Do we need to remove the step from here?

cosmetics-web/config/routes.rb
```diff
@@ -16,13 +16,18 @@
...
 
     collection do
       resources :account, controller: "responsible_persons/account_wizard", only: %i[show update] do
         collection do
           get :create_or_join_existing
           get :join_existing
         end
+
+        member do
+          get 'verify_email'
```
broder: Should we make this a symbol rather than a string?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,11 @@
+class ResponsiblePersons::VerificationController < ApplicationController
+  def show
+    email_verification_key = EmailVerificationKey.find_by!(
+      "responsible_person_id = ? AND key = ? AND expires_at >= ?", 
+      params[:responsible_person_id], params[:key], DateTime.current
+    )
+
+    email_verification_key.responsible_person.update('is_email_verified' => true)
```
broder: Does the new hash syntax work here? i.e. `is_email_verified: true`

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,11 @@
+class ResponsiblePersons::VerificationController < ApplicationController
+  def show
+    email_verification_key = EmailVerificationKey.find_by!(
```
broder: We could maybe make this more readable by pushing this into a model method like `validate_key_for_responsible_person(rp_id, key)` which could raise an error or set `is_email_verified` to `true`. It might then be easier to unit test as well. It's functionally equivalent though!

cosmetics-web/spec/factories/email_verification_key.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :email_verification_key do
+    key { "aaaaaaaaaaa" }
+    expires_at { 2.days.from_now }
+    after(:create) do |email_verification_key|
```
tom-bell-softwire: Is this `after` block needed?

shared-web/config/routes.rb
```diff
@@ -9,5 +12,6 @@
 
   if Rails.env.development?
     get "components/:component" => "components_gallery#show"
+    mount Sidekiq::Web => '/sidekiq'
```
tom-bell-softwire: Oooh, what does this do?!

DWRendell: It mounts the sidekiq web UI, which lets you see queued jobs, failed jobs, scheduled jobs and other cool stats which are useful for debugging async tasks.

Just found out it's broken the sign out link for some reason though, will have to figure out why...

cosmetics-web/spec/jobs/remove_expired_email_verification_keys_job_spec.rb
DWRendell: I've had a try and this doesn't work, nor does using `let!` which is meant to call before each test. The problem is that the `after(:create)` blocks in the factory aren't being called so the keys are created with random strings as keys and the default expiration date :|

cosmetics-web/spec/factories/email_verification_key.rb
DWRendell: Yes, because the EmailVerificationKey uses ActiveRecord on_create callbacks to set a random key and set the expiration date, we need to use the callback to set these values or they get overwritten. I've removed the blocks above that try and set the value before though, they aren't needed

cosmetics-web/spec/jobs/remove_expired_email_verification_keys_job_spec.rb
tom-bell-softwire: Thanks for looking into this! (And thanks for mentioning the `let!` variant, which I wasn't aware of.)

Well... that's annoying! Fine to leave this as is, in that case.

cosmetics-web/spec/factories/email_verification_key.rb
tom-bell-softwire: Thanks!

DWRendell:
I was trying to fix something in a review markup and I realised I'd already fixed it in my COSBETA-106 branch and couldn't be bothered to figure out how to merge things, so I'm just going to push all my changes for COSBETA-106 to this review as well

cosmetics-web/app/mailers/notify_mailer.rb
```diff
@@ -1,14 +1,14 @@
 class NotifyMailer < GovukNotifyRails::Mailer
-  def send_test_email(name, email)
-    set_template('c6124932-1be2-483e-8be8-66dda071375e')
-    set_reference('Test email')
+  # https://www.notifications.service.gov.uk/services/b1dab0f5-4651-4af8-9f15-fa8143ff338d/templates/50072d05-d058-4a02-a239-0d73ef7291b2
```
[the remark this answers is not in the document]
DWRendell: OK, removed

cosmetics-web/config/routes.rb
DWRendell: Done/redundant

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: Done

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: Sounds good, done

shared-web/config/routes.rb
DWRendell: This has stopped happening... maybe it was just a glitch in the matrix...

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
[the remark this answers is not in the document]
DWRendell: Done

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -28,12 +28,23 @@ def show
     render_wizard
   end
 
+  # GET /responsible_persons/account/verify_email
+  def verify_email; end
```
tom-bell-softwire: Is this used any more?

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -4,7 +4,7 @@ class ResponsiblePersons::AccountWizardController < ApplicationController
   steps :select_type, :enter_details
 
   skip_before_action :create_or_join_responsible_person
-  before_action :set_responsible_person, only: %i[show update]
+  before_action :set_responsible_person, only: %i[show update verify_email]
```
tom-bell-softwire: Is `verify_email` used any more?

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -46,6 +57,11 @@ def update
     end
   end
 
+  def finish_wizard_path
+    @responsible_person = User.current.responsible_persons.first
```
tom-bell-softwire: Is this correct? What happens if they already had a verified RP account and have just finished creating an additional RP?

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -28,12 +28,23 @@ def show
...
-        redirect_to responsible_person_path(@responsible_person)
+        key = @responsible_person.email_verification_keys.create
+
+        NotifyMailer.send_responsible_person_verification_email(
+          @responsible_person.email_address,
+          User.current.full_name,
+          responsible_person_email_verification_key_url(@responsible_person, key.key)
+        ).deliver_later
+
+        redirect_to finish_wizard_path
```
tom-bell-softwire: Is there any benefit to this over just redirecting directly to `responsible_person_email_verification_keys_path(@responsible_person)` ?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,34 @@
+class ResponsiblePersons::VerificationController < ApplicationController
+  before_action :set_responsible_person, only: %i[index resend_email]
+  skip_before_action :create_or_join_responsible_person
+
+  def show
+    email_verification_key = EmailVerificationKey.verify_key_for_responsible_person(
+      params[:responsible_person_id],
+      params[:key]
+)
```
tom-bell-softwire: Indentation looks a bit off here.

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    email_verification_key.responsible_person.update(is_email_verified: true)
+    redirect_to responsible_person_path(email_verification_key.responsible_person)
+  end
+
+  def index; end
+
+  def resend_email
+    key = @responsible_person.email_verification_keys.create
```
tom-bell-softwire: Should we invalidate/remove the previous key when creating a new one?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  before_action :set_responsible_person, only: %i[index resend_email]
+  skip_before_action :create_or_join_responsible_person
+
+  def show
+    email_verification_key = EmailVerificationKey.verify_key_for_responsible_person(
+      params[:responsible_person_id],
+      params[:key]
+)
+
+    email_verification_key.responsible_person.update(is_email_verified: true)
```
tom-bell-softwire: Will `email_verification_key` be nil if someone uses an expired key?

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -28,12 +28,23 @@ def show
...
+  # GET /responsible_persons/account/verify_email
+  def verify_email; end
+
   # PATCH/PUT /responsible_persons/account/:step
   def update
     case step
     when :enter_details
       if responsible_person_saved?
-        redirect_to responsible_person_path(@responsible_person)
+        key = @responsible_person.email_verification_keys.create
```
tom-bell-softwire: This logic is pretty much identical to that in `VerificationController#resend_email`. Is it worth extracting it into a method in a concern?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -0,0 +1,34 @@
+class ResponsiblePersons::VerificationController < ApplicationController
+  before_action :set_responsible_person, only: %i[index resend_email]
+  skip_before_action :create_or_join_responsible_person
+
+  def show
+    email_verification_key = EmailVerificationKey.verify_key_for_responsible_person(
+      params[:responsible_person_id],
```
tom-bell-softwire: Could you just pass in `@responsible_person` here?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
tom-bell-softwire: Do we need to display an error message if the key has expired?

cosmetics-web/app/models/email_verification_key.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.key_validity_duration
+    1.day
+  end
+
+  def is_expired?
+    expires_at < DateTime.current
+  end
+
+  def self.verify_key_for_responsible_person(responsible_person_id, key)
```
tom-bell-softwire: Might be neater if this accepted a `ResponsiblePerson` rather than its ID.

shared-web/config/routes.rb
tom-bell-softwire: Awesome! This sounds like it'll be really useful.

cosmetics-web/spec/responsible_person_helpers.rb
```diff
@@ -1,6 +1,7 @@
 module ResponsiblePersonHelpers
   def sign_in_as_member_of_responsible_person(responsible_person, user: nil)
     user ||= build(:user)
+    responsible_person.update is_email_verified: true
```
tom-bell-softwire: Not sure this is needed, since the default factory sets this property to `true`.

Might be useful to create a nested factory variant called something like `unverified_responsible_person` where that property is false, for use in scenarios that specifically test that logic.

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -4,7 +4,7 @@ class ResponsiblePersons::AccountWizardController < ApplicationController
   steps :select_type, :enter_details
 
   skip_before_action :create_or_join_responsible_person
```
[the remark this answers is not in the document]
DWRendell: I spoke to Ralph about this, we think users should be able to create more Responsible People before they've verified their first one

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
DWRendell: No, thanks, removed

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
DWRendell: Removed

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
DWRendell: Good point, removed as part of removing this method

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
DWRendell: Nope, in fact just redirecting makes it a lot easier to get the RP so will do that, thanks

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: I've tried fixing it but Rubocop keeps breaking it again 🤷‍♂️

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: Good idea, will do

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: If the key is expired then the method should throw a NotFoundError and we should get redirected to a 404. Ralph thinks it's a good idea to show an error message for an expired key so I'll get onto implementing that (it will still 404 for keys that have been deleted by the cleanup job however)

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
DWRendell: I've moved the common logic inside the mailer method, does that sound OK?

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
DWRendell: Good idea, done

cosmetics-web/app/models/email_verification_key.rb
DWRendell: Done

cosmetics-web/spec/responsible_person_helpers.rb
DWRendell: Yep, all works without it. Will look for places that extra factory could be useful

PR #611, opened 2019-02-15
COSBETA-86: Populate basic info

aidayounesi:
Import the following fields into a new Notification from an imported file:

- Product Name
- Whether the product is imported/where it's imported from (note, this is a different field than the existing country field for the manual journey)
- Shades
- CPNP reference number


cosmetics-web/app/models/xml_info.rb
```diff
@@ -0,0 +1,27 @@
...
+  def cpnp_reference
+    @xml_doc.xpath('//cpnpReference').first&.text
+  end
+
+  def is_imported
+    @xml_doc.xpath('//currentVersion/generalInfo/imported').first&.text&.casecmp?('Y')
+  end
+
+  def imported_country
+    @xml_doc.xpath('//currentVersion/generalInfo/importedCty').first&.text
```
broder: Will this be the same form that we're storing countries from the manual journey? @DWRendell, are we storing them as e.g. "country:NZ" in the manual journey?

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
```diff
@@ -0,0 +1,10 @@
+class AddCpnpBasicInfoToNotification < ActiveRecord::Migration[5.2]
+  safety_assured do
+    change_table :notifications, bulk: true do |t|
+      t.string :cpnp_reference
+      t.boolean :cpnp_is_imported
```
broder: Are we going to have a `cpnp_*` equivalent for lots of fields?

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.string :cpnp_imported_country
+      t.string :shades
```
broder: Should this be stored on a component rather than on a notification?

cosmetics-web/spec/analyzers/read_data_analyzer_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    sign_in_as_member_of_responsible_person(responsible_person)
+    create_notification_file(blob)
+  end
+
+  after do
+    remove_uploaded_files
+  end
+
+  describe "accept" do
+    it "null blob fails" do
```
broder: These should ideally be readable i.e. "it rejects a null blob"

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
aidayounesi: I think these are the only fields with cpnp prefix.

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
aidayounesi: Requirements changed, we don't have shades per component anymore. It's per product/notification.

cosmetics-web/spec/analyzers/read_data_analyzer_spec.rb
aidayounesi: Sure, thanks.

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
broder: Fair enough! I was wondering whether we'd want to distinguish between a `ManualNotification` and a `CpnpNotification` but if these are the only fields then it's probably unnecessary 🙂

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
broder: Ah, cool - I'm assuming there's a separate ticket to update the manual journey?

aidayounesi: Yes, https://regulatorydelivery.atlassian.net/browse/COSBETA-108

cosmetics-web/app/views/responsible_persons/notifications/tabs/_unfinished.html.slim
```diff
@@ -20,7 +20,7 @@
           dt
             | EU reference number:
           dd
-            | 31869
+            = notification.cpnp_reference
```
tom-bell-softwire: Should we also be capturing the "first notified in EU" property as part of this ticket? Might be worth having a chat with @DWRendell about this.

cosmetics-web/db/schema.rb
```diff
@@ -70,6 +70,10 @@
     t.string "import_country"
     t.bigint "responsible_person_id"
     t.integer "reference_number"
+    t.string "cpnp_reference"
+    t.boolean "cpnp_is_imported"
+    t.string "cpnp_imported_country"
+    t.string "shades"
```
tom-bell-softwire: Should this be plural?

cosmetics-web/spec/notification_file_helpers.rb
```diff
@@ -6,4 +6,16 @@ def mock_antivirus
   def unmock_antivirus
     allow(Clamby).to receive(:safe?).and_call_original
   end
+
+  def create_file_blob(filename: "testExportFile.zip", content_type: "application/zip", metadata: nil)
+    ActiveStorage::Blob.create_after_upload! io: file_fixture(filename).open, filename: filename, content_type: content_type, metadata: metadata
+  end
+
+  def create_notification_file(blob)
+    NotificationFile.create(name: blob.filename, uploaded_file: blob)
```
tom-bell-softwire: Could you use a FactoryBot factory to create `notification_file` instances for testing?

cosmetics-web/app/models/xml_info.rb
aidayounesi: Yes, in manual journey it's either country:XX or territory:XX. 
We have the list of location-autocomplete in gov-uk and list of countries available in CPNP. The following country id are available in CPNP but not exactly in gov-uk:
AN, CP, I0, SH, UM

cosmetics-web/db/migrate/20190214145420_add_cpnp_basic_info_to_notification.rb
aidayounesi: probably as you said not necessary

cosmetics-web/app/models/xml_info.rb
```diff
@@ -0,0 +1,27 @@
+class XMLInfo
```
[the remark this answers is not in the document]
aidayounesi: Sure, thanks

cosmetics-web/db/schema.rb
aidayounesi: Apparently this can be a string of space-separated shades

cosmetics-web/app/views/responsible_persons/notifications/tabs/_unfinished.html.slim
aidayounesi: Sure, I will include it, thanks

cosmetics-web/spec/notification_file_helpers.rb
aidayounesi: Yes, sure, thanks

cosmetics-web/app/models/xml_info.rb
aidayounesi: It's mapped to gov-uk countries/territories. It will save null if it cannot find any relevant in gov-uk

PR #612, opened 2019-02-15

kicferk1:


mspsds-web/db/migrate/20190215121322_fix_incorrect_pretty_ids.rb
```diff
@@ -1,5 +1,5 @@
 class FixIncorrectPrettyIds < ActiveRecord::Migration[5.2]
   def change
-    Investigation.in_batches.each(&:add_pretty_id)
+    Investigation.in_batches { |relation| relation.each(&:add_pretty_id) }
```
broder: Actually, shouldn't this be `Investigation.in_batches.each_record(&:add_pretty_id)`?

PR #616, opened 2019-02-18

kicferk1:
Last version worked great with cases from current month, but not so great with cases from before current month. 

I can't really understand the results I got by running the old migration on int, the current month ids don't seem to be in the right order(when sorting by created_at), the older ones all use the count of 5(count of cases from current month), except for one that uses 4. 

Maybe having 2 conditions on the same field in different wheres with different types doesn't generate the query the way I thought it does :( 

In any case, this should make case ids unique for older months, migration works for the limited dataset I have locally.

### DIFF
--- a/mspsds-web/db/migrate/20190218110830_make_pretty_id_work_with_old_cases.rb
+++ b/mspsds-web/db/migrate/20190218110830_make_pretty_id_work_with_old_cases.rb
@@ -1,5 +1,0 @@
-class MakePrettyIdWorkWithOldCases < ActiveRecord::Migration[5.2]
-  def change
-    Investigation.in_batches.each_record(&:add_pretty_id)
-  end
-end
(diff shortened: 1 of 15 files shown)


mspsds-web/db/migrate/20190218110830_make_pretty_id_work_with_old_cases.rb
```diff
@@ -1,4 +1,4 @@
-class FixIncorrectPrettyIds < ActiveRecord::Migration[5.2]
+class MakePrettyIdWorkWithOldCases < ActiveRecord::Migration[5.2]
```
broder: Does changing this name cause the migration to re-run? Do we want to just manually down and up the migration on int?

kicferk1: Not sure if name change will make it re-run, but timestamp is different too, as is database version.
So it should run just by merging it. 

I generated new migration, but it looked just like the old one so I removed the old one.

broder: I'm not sure that deleting migrations is the best strategy - we can chat about it after lunch.

PR #617, opened 2019-02-18
COSBETA-30: handling upload errors

raamSoftwire:
This is still a WIP.


raamSoftwire:
I have split COSBETA-30 into two tickets, this PR comprises most of the first ticket.

Still outstanding are 
- button styling on the dashboard page
- handling errors around an actual virus being uploaded
- formatting of error messages
- system tests for uploading an incorrect file type or a file that is over the size limit

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
```diff
@@ -20,6 +20,7 @@
         .form-group
           = form.label "Upload CPNP zip file", class: "govuk-label"
           = form.file_field :uploaded_file,
-                  class: "govuk-file-upload", id: "uploaded_file"
+                  class: "govuk-file-upload", id: "uploaded_file", \
```
tom-bell-softwire: Now that I've spotted this, I think you can remove the `id` here – it should be added automatically by the form helper.

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
```diff
@@ -28,7 +28,7 @@
             class: "govuk-button govuk-button", role: "button"
 
 = render "tabs", tabs: [\
-  Tab.new("errors", "Errors (0)", "Files that failed to upload",
+  Tab.new("errors", "Errors (#{@erroneous_notification_files.count})", "Files that failed to upload",
```
tom-bell-softwire: Should this be `total_entries` like the other tabs below?

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
```diff
@@ -2,9 +2,12 @@
   .govuk-grid-column-two-thirds
     ul.govuk-list
       li[class="govuk-!-padding-bottom-2"]
-        = link_to "Try uploading files again", "/", class: "govuk-link--no-visited-state"
+        = link_to "Try uploading files again", new_responsible_person_notification_file_path(@responsible_person),
```
tom-bell-softwire: How does this functionality work? Should we also delete all the erroneous `NotificationFiles` before redirecting? Would the user be surprised to see duplicate entries if not?

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
```diff
@@ -14,9 +17,16 @@
...
-            | Incorrect file type
+        - @erroneous_notification_files.each do |notification_file|
+          tr.govuk-table__row
+            td.govuk-table__cell
+              = notification_file.name
+            td.govuk-table__cell
+              / = get_error_message(notification_file.upload_error)
+              = notification_file.get_upload_error_message
+            td.govuk-table__cell
+              = link_to "Dismiss", responsible_person_notification_file_path(@responsible_person, notification_file.id),
```
tom-bell-softwire: I think it should work without the `.id` on `notification_file`

cosmetics-web/config/locales/en.yml
```diff
@@ -27,3 +27,13 @@ en:
...
+      notification_file:
+        upload_errors:
+          uploaded_file_not_a_zip: "Uploaded file is not a ZIP file"
+          unzipped_files_not_xml: "Unzipped files are not XML files"
+          unzipped_files_are_pdf: "Unzipped files are PDF files"
+          file_flagged_as_virus: "Unzipped files contain a virus"
+          file_size_too_big: "Uploaded file exceeds size limit"
+          notification_validation_error: "Unknown error"
+          notification_duplicated: "A notification for this file already exists"
+          unknown_error: "Unknown error"
```
tom-bell-softwire: Awesome!!

cosmetics-web/spec/system/responsible_person_dashboard_spec.rb
```diff
@@ -14,6 +14,15 @@
...
 
+  it "only shows user the erroneous notifications belonging to their Responsible Person" do
+    create(:notification_file, responsible_person: responsible_person_1, user: user_1,
+           upload_error: "uploaded_file_not_a_zip")
+    create(:notification_file, responsible_person: responsible_person_2,
+                                                              upload_error: "uploaded_file_not_a_zip")
+    visit responsible_person_notifications_path(responsible_person_1)
+    assert_text "Errors (1)"
+  end
+
```
tom-bell-softwire: Worth testing that it shows the erroneous notification files for other users belonging to the same Responsible Person?

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -10,22 +10,32 @@ def create
...
     end
   end
 
+  def destroy
+    NotificationFile.delete(params[:id])
+    redirect_to responsible_person_notifications_path(@responsible_person)
+  end
+
+  def destroy_all
+    NotificationFile.delete_all
```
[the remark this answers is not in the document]
raamSoftwire: Very good point! I think it actually wants to be scoped just to the current user, rather than all users for a given RP. I'll make sure to add a system test for this too.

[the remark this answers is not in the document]
raamSoftwire: (I've checked this assumption with Ralph)

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -3,11 +3,13 @@ class ResponsiblePersons::NotificationsController < ApplicationController
 
   def index
     @pending_notification_files_count =
-      @responsible_person.notification_files.where(user_id: current_user.id).count
+      @responsible_person.notification_files.where(user_id: current_user.id, upload_error: nil).count
```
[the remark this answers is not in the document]
raamSoftwire: I agree, that's much nicer. So I've now pulled that out. I have also tried changing it to use `where(user: current_user, ...)`, but it doesn't work. I get `undefined method `arel_table' for User:Class`.

cosmetics-web/app/models/notification_file.rb
```diff
@@ -2,4 +2,43 @@ class NotificationFile < ApplicationRecord
   has_one_attached :uploaded_file
   belongs_to :responsible_person
   belongs_to :user
+
+  validate :uploaded_file_is_zip?
+  validate :uploaded_file_is_within_allowed_size?
+
```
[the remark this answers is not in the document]
raamSoftwire: Sorry! Surprised this wasn't picked up by the Rubocop

cosmetics-web/app/models/notification_file.rb
```diff
@@ -2,4 +2,43 @@ class NotificationFile < ApplicationRecord
...
+    unzipped_files_not_xml: "unzipped_files_not_xml",
+    unzipped_files_are_pdf: "unzipped_files_are_pdf",
+    file_flagged_as_virus: "file_flagged_as_virus",
+    file_size_too_big: "file_size_too_big",
+    notification_validation_error: "notification_validation_error",
+    notification_duplicated: "notification_duplicated",
+    unknown_error: "unknown_error"
+  }
+
+  @max_file_size_bytes = 30.megabytes
```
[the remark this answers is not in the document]
raamSoftwire: I was following the pattern from `image_upload.rb`,

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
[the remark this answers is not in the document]
broder: It's something to do with it being an ActiveHash rather than an actual database table - probably best to stick with the `user_id` version. I get caught out by this all the time too 😞

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
raamSoftwire: Done, thanks

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
raamSoftwire: Yes it should, changed it now.Thanks

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
raamSoftwire: So I spoke to Ralph about this, and the rows in the errors table should be thought of as attempts to upload, rather than relating to files. So in that case it would be less surprising to leave previous attempts on show.

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
raamSoftwire: Cool thanks

cosmetics-web/spec/system/responsible_person_dashboard_spec.rb
raamSoftwire: I've discussed this with Ralph and erroneous notification files are scoped to the current user (and RP).

cosmetics-web/app/models/notification_file.rb
```diff
@@ -2,4 +2,43 @@ class NotificationFile < ApplicationRecord
...
+
+  def self.get_max_file_size
+    @max_file_size_bytes
+  end
+
+  def get_upload_error_message
+    I18n.t("activerecord.attributes.notification_file.upload_errors.#{self.upload_error}")
```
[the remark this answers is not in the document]
raamSoftwire: I guess so, for completeness? But we shouldn't be calling this method unless we already know that the notification file has an `upload_error`. I've added handling for it, not sure what we should show to the user. I've gone with "No error" for now.

cosmetics-web/app/models/notification_file.rb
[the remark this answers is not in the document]
raamSoftwire: FYI, this is already due to be refactored into a concern by COSBETA-26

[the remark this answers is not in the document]
tom-bell-softwire: Great, thanks!

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
tom-bell-softwire: That makes sense. Thanks for clarifying.

cosmetics-web/app/assets/application/stylesheets/responsible_persons.scss
```diff
@@ -29,22 +47,17 @@
   }
 }
 
-.loading-spinner__container {
-  border-color: govuk-colour("blue");
+.dismiss-link {
```
tom-bell-softwire: Do we need this extra styling? We should be trying to use the Design System classes wherever possible.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -5,27 +5,36 @@ class ResponsiblePersons::NotificationFilesController < ApplicationController
...
-        format.html { redirect_to responsible_person_notifications_path(@responsible_person) }
-        format.json { render :show, status: :created, location: @notification_file }
-      else
-        format.html { render :new }
-        format.json { render json: @notification_file.errors, status: :unprocessable_entity }
-      end
+    @notification_file.name = notification_file_params[:uploaded_file].original_filename
+    @notification_file.responsible_person = @responsible_person
+    @notification_file.user = User.current
+    @notification_file.uploaded_file.attach(notification_file_params[:uploaded_file])
```
tom-bell-softwire: Is this needed? I thought the `NotificationFile.new(notification_file_params)` below would already attach it.

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
```diff
@@ -14,9 +17,16 @@
...
-          td.govuk-table__cell
-            | Example product
-          td.govuk-table__cell
-            | Incorrect file type
+        - @erroneous_notification_files.each do |notification_file|
+          tr.govuk-table__row
+            td.govuk-table__cell
+              = notification_file.name
+            td.govuk-table__cell
+              / = get_error_message(notification_file.upload_error)
```
tom-bell-softwire: Can this be deleted now?

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
```diff
@@ -14,9 +17,16 @@
...
+              = notification_file.upload_error_message
+            td.govuk-table__cell
+              = link_to "Dismiss", responsible_person_notification_file_path(@responsible_person, notification_file),
+                      class: "govuk-link--no-visited-state", method: :delete
```
tom-bell-softwire: We're using `method: :delete` here, but a form with a submit button for the "Dismiss all" action above. Will this have implications for users without JavaScript?

shared-web/app/analyzers/shared/web/master_analyzer.rb
```diff
@@ -7,16 +7,14 @@ def self.accept?(_blob)
...
-      end
-
-    private
-
-      def analyzers
-        Rails.application.config.document_analyzers
-            .select { |analyzer_class| analyzer_class.accept? @blob }
-            .collect { |analyzer_class| analyzer_class.new(@blob) }
+        collected_metadata = []
+        Rails.application.config.document_analyzers.each do |analyzer_class|
```
tom-bell-softwire: Nice!

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -5,27 +5,36 @@ class ResponsiblePersons::NotificationFilesController < ApplicationController
...
     end
   end
 
+  def destroy
+    NotificationFile.delete(params[:id])
+    redirect_to responsible_person_notifications_path(@responsible_person)
+  end
+
+  def destroy_all
+    @responsible_person.notification_files.where(user_id: current_user.id).where.not(upload_error: nil).delete_all
```
tom-bell-softwire: I think you'll need to change `current_user` to `User.current`

cosmetics-web/app/assets/application/stylesheets/responsible_persons.scss
raamSoftwire: Yes we do, this is used to make a submit button look like a link. I've spoken to Ralph and he doesn't want dismissing errors to feel like a big deal with a big button, so he's asked for this to be rendered as a link instead.

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
raamSoftwire: Oops! Yes

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: Done, thanks

cosmetics-web/app/analyzers/anti_virus_analyzer.rb
```diff
@@ -0,0 +1,19 @@
...
+        { safe: true }
+      else
+        Rails.logger.warn "#{@blob.id} detected as virus, removing."
+        notification_file = get_notification_file_from_blob(@blob)
+        notification_file.update(upload_error: :file_flagged_as_virus)
+        attachments = ActiveStorage::Attachment.where(blob_id: @blob.id)
+        attachments.each(&:destroy)
+        @blob.purge_later
+        { safe: false }
+      end
```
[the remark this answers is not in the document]
raamSoftwire: That's a good point, thanks

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: Ahh, I didn't understand that. Thanks

cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
raamSoftwire: I guess it will? I've now rewritten it as a form as above, is that the correct approach?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -8,7 +8,8 @@ def initialize(blob)
   end
 
   def self.accept?(given_blob)
-    return false unless given_blob.present? && given_blob.metadata["safe"]
+    return false unless given_blob.present? && given_blob.metadata["safe"] &&
+      NotificationFile.get_content_types.include?(given_blob.content_type)
 
     # this analyzer only accepts notification files which are zip
```
tom-bell-softwire: This comment feels misleading here – perhaps best to remove it.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -53,6 +62,18 @@ def get_xml_file_content
     end
   end
 
+  def contains_pdf?
+    download_blob_to_tempfile do |file|
```
tom-bell-softwire: I think this means we're downloading and unzipping the file twice (once here and once in `get_xml_file_content`), which isn't ideal.

Could we instead combine these checks into one method and raise errors to handle the two cases above? Something along the lines of
```ruby
def get_product_xml_file
  download_blob_to_tempfile do |zip_file|
    Zip::File.open(zip_file.path) do |files|
      files.each do |file|
        if file_is_pdf?(file)
          raise UnexpectedPdfFileError
        elsif file_is_product_xml?(file)
          yield file.get_input_stream.read
        end
      end
      raise ProductFileNotFoundError
    end
  end
end
```

You can then `rescue` these specific errors within `create_notification_from_file` and attach the relevant error to the `NotificationFile`.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -28,23 +29,31 @@ def create_notification_from_file
...
-      @xml_info = CPNPExport.new(xml_file_content)
-    end
+      if xml_file_content.blank?
+        if contains_pdf?
+          notification_file.update(upload_error: :unzipped_files_are_pdf)
+        else
+          notification_file.update(upload_error: :unzipped_files_not_xml)
+        end
+      else
+        @xml_info = CPNPExport.new(xml_file_content)
```
tom-bell-softwire: Not something you've added, but just spotted this – might be good to call this variable something like `cpnp_export` instead.

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Done

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: I've renamed the class `CpnpExport` which follows the same pattern as `NotificationFile`

PR #621, opened 2019-02-18
MSPSDS-1026: Remove constraint and index on company number

kicferk1:
Should it be merged to master or staging?


broder:
Let's go for staging.

PR #626, opened 2019-02-19
MSPSDS-918: Clear autocomplete

kicferk1:
I've looked into getting` simpleAccessibleAutocomplete` called inside slim, but I didn't figure out anything nicer than what we have now.


mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
+- returned_value = local_assigns[:returned_value] || :itself
+- displayed_value = local_assigns[:displayed_value] || :itself
+.govuk-grid-row
+  .govuk-grid-column [class="#{local_assigns[:column_class]}"]
+    = form.collection_select key, options, returned_value, displayed_value,
+            { include_blank: true, selected: local_assigns[:selected] },
+            class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
+  - if local_assigns[:show_delete_x]
+    span style="padding: 1% 0; font-size: x-large" id="remove-picker-#{key}"
+      span#delete-hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
```
jasiekmiko: Not sure we need `govuk-!-width-two-thirds` when we're hiding it?

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
+- returned_value = local_assigns[:returned_value] || :itself
```
jasiekmiko: What do you think about following the same convention as the shared components and placing it in
`/mspsds-web/app/views/application/form_components/_autocomplete_select.html.slim`?

There's also a question if we want this to simply be shared. @tom-bell-softwire, would you guys find this generic enough to be useful? (to be clear, this is separate from implementing a non-js, design system `select` component, which we are tracking through https://regulatorydelivery.atlassian.net/browse/MSPSDS-1004)

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
```diff
@@ -4,22 +4,22 @@
   - if current_user.teams.any?
     - someone_in_your_team = capture do
       = render "investigations/assign/assignee_selection", form: form, key: :select_team_member,
-              options: compute_select_options(entities: User.get_team_members(user: current_user)),
-              accessible_hint: "Name"
+              options: User.get_team_members(user: current_user),
+              accessible_hint: "Name", title: "Assignee", show_delete_x: true
```
jasiekmiko: Why are we passing `title` in when it's always the same?

mspsds-web/app/views/investigations/assign/_assignee_selection.html.slim
```diff
@@ -1,7 +1,7 @@
-.govuk-grid-row
-  .govuk-grid-column-one-half
-    .govuk-form-group
-      span#other-assignee-hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
-        = accessible_hint
-      = form.label key, "Assignee", class: "govuk-label"
-      = form.select key, options_for_select(options, ''), {}, class: "govuk-select", id: "assignee-picker-#{key}"
+- display_method = local_assigns[:ignore_visibility_restrictions] ? :display_name_without_restrictions : :display_name
+.govuk-form-group
+  span#hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
```
jasiekmiko: Not sure we need `govuk-!-width-two-thirds` when we're hiding it?

mspsds-web/app/views/investigations/assign/_assignee_selection.html.slim
```diff
@@ -1,7 +1,7 @@
-.govuk-grid-row
-  .govuk-grid-column-one-half
-    .govuk-form-group
-      span#other-assignee-hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
-        = accessible_hint
-      = form.label key, "Assignee", class: "govuk-label"
-      = form.select key, options_for_select(options, ''), {}, class: "govuk-select", id: "assignee-picker-#{key}"
+- display_method = local_assigns[:ignore_visibility_restrictions] ? :display_name_without_restrictions : :display_name
```
jasiekmiko: If we're already providing modal behaviour, I think it'll be easier if we just accept the method symbol as a parameter to this partial, instead of a flag which gets translated to method name

mspsds-web/app/views/investigations/assign/_assignee_selection.html.slim
```diff
@@ -1,7 +1,7 @@
...
+.govuk-form-group
```
jasiekmiko: The form-group div, hint and label should all be part of the select component

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
...
+- displayed_value = local_assigns[:displayed_value] || :itself
+.govuk-grid-row
+  .govuk-grid-column [class="#{local_assigns[:column_class]}"]
```
jasiekmiko: The row and column divs are part of the particular layout of the page and shouldn't be included in the component code I think.

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
...
+    = form.collection_select key, options, returned_value, displayed_value,
+            { include_blank: true, selected: local_assigns[:selected] },
+            class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
+  - if local_assigns[:show_delete_x]
+    span style="padding: 1% 0; font-size: x-large" id="remove-picker-#{key}"
+      span#delete-hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
+        | Click this to remove currently selected value
+      | ✖
```
jasiekmiko: Are we in danger of this character being displayed as a funky emoji in some cases? @broder, any thoughts? I'm not sure how fonts and special characters interact :P

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
+- returned_value = local_assigns[:returned_value] || :itself
+- displayed_value = local_assigns[:displayed_value] || :itself
+.govuk-grid-row
+  .govuk-grid-column [class="#{local_assigns[:column_class]}"]
+    = form.collection_select key, options, returned_value, displayed_value,
+            { include_blank: true, selected: local_assigns[:selected] },
+            class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
+  - if local_assigns[:show_delete_x]
+    span style="padding: 1% 0; font-size: x-large" id="remove-picker-#{key}"
```
jasiekmiko: It probably ought to be a button, or have the `role='button'` assigned to it

mspsds-web/app/assets/application/javascripts/autocomplete.js
```diff
@@ -32,5 +33,17 @@ export function simpleAccessibleAutocomplete(id, autocompleteOptions) {
         $enhancedElement.click();
       });
     }
+
+    // This adds ability to remove currently selected input by pressing on an X next to it
+    // This is a work-around to
+    // https://github.com/alphagov/accessible-autocomplete/issues/240
+    const removeButton = document.getElementById("remove-" + id);
+    if (removeButton) {
+      removeButton.addEventListener("click", () => {
```
jasiekmiko: Is `click`ing alone enough? Does it want to be keyboard-usable?

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,12 @@
+- returned_value = local_assigns[:returned_value] || :itself
+- displayed_value = local_assigns[:displayed_value] || :itself
+.govuk-grid-row
+  .govuk-grid-column [class="#{local_assigns[:column_class]}"]
+    = form.collection_select key, options, returned_value, displayed_value,
+            { include_blank: true, selected: local_assigns[:selected] },
+            class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
+  - if local_assigns[:show_delete_x]
```
jasiekmiko: Another benefit of having the `javascript:` be called somewhere from within the component is we could have tie this variable to `showAllEntries` instead of being passed independently. This would follow Ed's [comment](https://regulatorydelivery.atlassian.net/browse/MSPSDS-918?focusedCommentId=13683&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13683)

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
broder: Could we use an X character? Or a simple SVG?

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -0,0 +1,25 @@
...
+.govuk-form-group class=form_group_classes
+  = local_assigns[:preface_html]
+  - if local_assigns[:errorMessage].present?
+    = render "components/govuk_error_message", **errorMessage
+  div class=select_class
+    = form.collection_select key, options, returned_value, displayed_value,
+            { include_blank: true, selected: local_assigns[:selected] }, class: "govuk-select",
+            id: id, aria_describedby: local_assigns[:aria_describedby]
+  - if show_all_values
+    button id="remove-picker-#{key}" class="autocomplete__remove-button" type="button"
```
jasiekmiko: Would `clear` rather than `remove` be more accurate?
Also the logic of adding-or-not the button probably warrants an explanatory comment :)

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -7,6 +7,6 @@
             class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
   - if local_assigns[:show_delete_x]
     span style="padding: 1% 0; font-size: x-large" id="remove-picker-#{key}"
-      span#delete-hint.govuk-hint [class="govuk-!-width-two-thirds govuk-visually-hidden"]
+      span#delete-hint.govuk-hint
```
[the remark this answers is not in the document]
kicferk1: I over zealously deleted both, brought it back in a commit further down the line

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
```diff
@@ -6,7 +6,7 @@
             { include_blank: true, selected: local_assigns[:selected] },
             class: "govuk-select", id: "picker-#{key}", aria_describedby: local_assigns[:aria_describedby]
   - if local_assigns[:show_delete_x]
-    span id="remove-picker-#{key}"
+    button id="remove-picker-#{key}" class="autocomplete__remove-button" type="button"
```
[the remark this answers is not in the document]
kicferk1: Yes, without it clicking on this button will submit a form it's in :( 
I considered making it a non-button, but making it a button allows it to be selectable by tabbing through site.

mspsds-web/app/assets/packs/application.js
```diff
@@ -8,3 +8,4 @@
 
 import '../application/stylesheets/main.scss';
 import './shared';
+import '../application/javascripts/autocomplete';
```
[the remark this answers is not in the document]
kicferk1: Shouldn't really be there, I wanted to make sure something gets run so I put it in application directly, and didn't remove import :)

mspsds-web/app/views/application/_autocomplete_select_component.html.slim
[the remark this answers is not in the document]
jasiekmiko: Ah ok - that sounds like a really good reason and the right solution! :)

mspsds-web/app/assets/application/javascripts/autocomplete.js
```diff
@@ -32,5 +33,26 @@ export function simpleAccessibleAutocomplete(id, autocompleteOptions) {
         $enhancedElement.click();
       });
     }
+
+    // This adds ability to remove currently selected input by pressing on an X next to it
+    // This is a work-around to
+    // https://github.com/alphagov/accessible-autocomplete/issues/240
+    const removeButton = document.getElementById(`remove-${id}`);
+    if (removeButton) {
+      const removeCallback = () => {
```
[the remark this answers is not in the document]
kicferk1: keyPress was something I found in some dark corner of the internet, and I can't find it again. It seems to trigger just with space and enter, as opposed to keypress(notice lower case), that triggers on each key. 

But I can't find documentation for keyPress anywhere, so it's probably best to be a little more explicit.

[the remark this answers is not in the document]
jasiekmiko: sounds like a good conclusion :D

mspsds-web/app/assets/application/javascripts/autocomplete.js
```diff
@@ -39,13 +39,19 @@ function simpleAccessibleAutocomplete(id, autocompleteOptions) {
...
-      const removeCallback = () => {
+      const removeValue = () => {
         $enhancedElement.val('');
         $enhancedElement.click().focus().blur();
         $(element).parent().find('select').val('');
       };
-      removeButton.addEventListener('keyPress', removeCallback);
-      removeButton.addEventListener('click', removeCallback);
+      removeButton.addEventListener('keypress', (e) => {
+        if (e.keyCode === 13 || e.keyCode === 32) {
```
jasiekmiko: can we explain what these are?

shared-web/app/views/form_components/_autocomplete_select.html.slim
```diff
@@ -1,25 +1,36 @@
 ruby:
   returned_value = local_assigns[:returned_value] || :itself
   displayed_value = local_assigns[:displayed_value] || :itself
-  id = "picker-#{key}"
+  id = local_assigns[:id] || "picker-#{key}"
   show_all_values = local_assigns[:show_all_values]
-  select_class = show_all_values ? "govuk-select-with-remove" : ""
+  if form.object.errors
```
jasiekmiko: I think you want `form.object&.errors`, sometimes the form isn't actually passed a model (not in the cases we have so far i don't think, but it'll cost us nothing to support that)

shared-web/app/views/form_components/_autocomplete_select.html.slim
```diff
@@ -1,25 +1,36 @@
...
-    = form.collection_select key, options, returned_value, displayed_value,
+    = render "components/govuk_error_message", **local_assigns[:errorMessage]
+
+  div class=("autocomplete-select-with-remove" if show_all_values)
+    = form.collection_select key, items, returned_value, displayed_value,
             { include_blank: true, selected: local_assigns[:selected] }, class: "govuk-select",
-            id: id, aria_describedby: local_assigns[:aria_describedby]
+            id: id
+
+  / We decided to only add clear button to autocomplete that shows all values, and hence the dropdown arrow
```
jasiekmiko: I'm really struggling to parse that comment. Would 
`The clear button is only needed when the select box is showing all values (and a dropdown arrow). See https://github.com/alphagov/accessible-autocomplete/issues/240#issuecomment-455859918 for background to this decision`
be clearer?

mspsds-web/app/assets/application/javascripts/autocomplete.js
```diff
@@ -39,13 +39,19 @@ function simpleAccessibleAutocomplete(id, autocompleteOptions) {
     // https://github.com/alphagov/accessible-autocomplete/issues/240
     const removeButton = document.getElementById(`remove-${id}`);
     if (removeButton) {
-      const removeCallback = () => {
+      const removeValue = () => {
```
jasiekmiko: Oh, and you've renamed it to `clear` in the slim file, let's keep it consistent across - don't mind which way :)
(there's a few more places that use `remove` atm, have a look thruogh the changes to catch them all!)

mspsds-web/test/system/investigation_assignee_test.rb
```diff
@@ -16,7 +16,7 @@ class InvestigationAssigneeTest < ApplicationSystemTestCase
     assert_text @user.display_name
     choose @user.display_name, visible: false
     click_on "Assign"
-    assert_text "Assigned to\n#{@user.full_name}"
```
jasiekmiko: These fixes, and the activity entry fix in the same commit, seem to be fixing issues that arise when the tests run without css/js (and, e.g., all the tabs are displayed in a stack). I've noticed those on my local machine before, but I haven't seem them break on travis, so I'm a little worried about the discrepancy (and specifically this failing when css does load). @broder any idea what could be causing tests to run like this? could it be a race-condition with the test executing before the frontend bundle gets downloaded and run?

jasiekmiko: @kicferk1 rebuilding my mspsds-web docker image seems to have fixed it for me - can you try the same? I rebuilt with `--no-cache`, in case that ends up relevant.
Either way, this edit wants to be reverted...

mspsds-web/test/system/activity_entry_test.rb
```diff
@@ -5,7 +5,7 @@ class ActivityEntryTest < ApplicationSystemTestCase
     sign_in_as_admin
     @investigation = investigations(:one)
     visit investigation_path(@investigation)
-    click_on "Add activity"
+    click_on "Add activity", match: :first
```
jasiekmiko: ... and this one ...

mspsds-web/test/system/investigation_assignee_test.rb
```diff
@@ -25,7 +25,7 @@ class InvestigationAssigneeTest < ApplicationSystemTestCase
     assert_text @team.name
     choose @team.name, visible: false
     click_on "Assign"
-    assert_text "Assigned to\n#{@team.name}"
+    assert_text "Assigned to#{@team.name}"
```
jasiekmiko: ... and this one :)

mspsds-web/test/system/investigation_assignee_test.rb
kicferk1: It did fix it

PR #627, opened 2019-02-19
MSPSDS-6: Correctly throw error on ClamAV problem

broder:
There are a couple of things going on here:

- The first is that the default Clamby config value for `error_clamscan_client_error` is `false`. That means if `clamd` has crashed (which I think is what's causing the issues on PaaS), `Clamby.safe?` returns `false`, we identify the file as a virus and delete it. It should now throw an error, which causes the job to be retried later (giving us time to restart the worker or something like that).
- When we detected (or incorrectly detected, due to the above) that a file was a virus, we'd delete it. We deleted it synchronously using `.purge`, which I think meant that the currently running analyser would fail when it tried to set the metadata (on the now deleted blob). This was the `Can't modify frozen hash` error which we've been seeing a lot on the worker / sentry. We now `.purge_later` which I think should give the analyser enough time to set the metadata before it's deleted.

I'm planning to raise another ticket to deal with the UI aspects of a file detected as a virus. At the minute, you'll get a variety of view and controller errors if you upload a virus.

### DIFF
--- a/shared-web/app/analyzers/shared/web/anti_virus_analyzer.rb
+++ b/shared-web/app/analyzers/shared/web/anti_virus_analyzer.rb
@@ -2,7 +2,13 @@ module Shared
   module Web
     class AntiVirusAnalyzer < ActiveStorage::Analyzer
       def initialize(blob)
-        config = { daemonize: true }
+        config = {
+          daemonize: true,
+          error_clamscan_missing: true,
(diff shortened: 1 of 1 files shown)


DWRendell:
@raamSoftwire This might be related to your problem?

broder:
I've updated the PR to merge into staging. We probably want to include this in the next prod release.

PR #628, opened 2019-02-20
MSPSDS-887 fix page template

MatthewBurstein:
This also covers MSPSDS-886 (incorrect vertical alignment throughout site.


mspsds-web/app/views/investigations/_case_card.html.slim
```diff
@@ -1,4 +1,5 @@
-.govuk-grid-row class=(investigation.is_closed ? "mspsds-case-card mspsds-case-card--closed" : "mspsds-case-card")
+- additional_classes = (investigation.is_closed ? "mspsds-case-card--closed" : "") + (local_assigns[:card_class] || "")
```
broder: Is this concatenating the class names as one string or as an array of strings?

mspsds-web/app/assets/application/stylesheets/main.scss
```diff
@@ -30,6 +30,7 @@
     // Shifts content of body to right by width of scrollbar
     padding-left: calc(100vw - 100%);
 
+    // The use of important here is forced because it is already used in .govuk-skip-link from the design system
     .skip-link-banner {
       margin-left: -20px !important;
```
jasiekmiko: Do we know the root reason why we need to fix the styling of that component? I would hope the css provided by the govuk frontend to be sufficient, so if it's not it'd be good to know what's getting in our way here..

MatthewBurstein: This whole block prevents the content from shifting to the left on pages with a vertical scrollbar. The design system at present does not take this situation into account. @edwardhorsford has raised it with them. In the meantime, all our content which stretches across the entire screen must have this negative margin, positive padding pattern to fill the screen with their background colour.

mspsds-web/app/assets/application/stylesheets/case-card.scss
```diff
@@ -100,6 +100,7 @@
   @extend %mspsds-card;
 }
 
+// add this class when displaying the card in an index page (as opposed to a tab for example)
```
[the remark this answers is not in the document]
MatthewBurstein: the same card component is used both in the index page, and inside a tab (eg the cases tab on the product page). In the index page case, it has the search sidebar to the left and `margin-left: 0` produces too much whitespace between the sidebar and the results. When inside a tab, however, it needs to be centred within the tab. If you have a better idea for how to do this I am all ears :)

mspsds-web/app/assets/application/stylesheets/main.scss
edwardhorsford: [Here's the issue](https://github.com/alphagov/govuk-frontend/issues/1204) I raised against govuk-frontend. I have no great preference either way - it's nice if it's fixed - but we should probably make sure we test it well. We can also leave it to design system team to look at in the fullness of time.

mspsds-web/app/assets/application/stylesheets/case-card.scss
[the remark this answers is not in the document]
edwardhorsford: I might argue the issue is with the container the cards sit in on the index page, not the card themselves.

I think Tim (or someone else?) added negative margin-left to the cards so that when in a 1/4, 3/4 grid (eg filter bar and table of cases), the gutter between columns was reduced - presumably to give more space to the table. This feels a bit hacky - though I'd like to review the design of this page as a whole so don't mind much what we do now.

mspsds-web/app/views/investigations/_case_card.html.slim
MatthewBurstein: Yes it's strings - good spot. Have fixed.

mspsds-web/app/assets/application/stylesheets/case-card.scss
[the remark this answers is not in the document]
MatthewBurstein: Following discussion with Jan and Ed we have removed `.mspsds-case-card--index` and allowed the gap as this is in keeping with the design system.

mspsds-web/app/assets/application/stylesheets/main.scss
MatthewBurstein: I think it works as it is here until the design system team can get to it.

mspsds-web/app/views/investigations/_table.html.slim
```diff
@@ -2,10 +2,12 @@
...
-        = render "highlight_card", investigation: search_result[:record], highlights: search_result.highlight
+        = render "highlight_card", investigation: search_result[:record], highlights: search_result.highlight,
+                card_class: "mspsds-case-card--index"
       - else
-        = render "case_card", investigation: search_result[:record], sorted_by: sorted_by
+        = render "case_card", investigation: search_result[:record], sorted_by: sorted_by,
+                card_class: "mspsds-case-card--index"
     - else
-      = render "restricted_card", investigation: search_result[:record]
+      = render "restricted_card", investigation: search_result[:record], card_class: "mspsds-case-card--index"
```
broder: Should we remove all usages of `mspsds-case-card--index` and `card_class` or are we still using it?

MatthewBurstein: Yes sorry have now done so. 👍

PR #633, opened 2019-02-20
MSPSDS-888 Fix location of error summary within pages

jasiekmiko:
Counterpart to https://github.com/UKGovernmentBEIS/beis-mspsds/pull/632, but escalated to hotfix staging


mspsds-web/app/views/businesses/edit.html.slim
```diff
@@ -1,9 +1,10 @@
...
 - content_for :after_header do
   = link_to "Back to business", @business, class: "govuk-back-link"
-= render "business_heading", business: @business
-h2.govuk-heading-l
-  | Business details
+
 = form_with(model: @business, local: true) do |form|
   = render "form_components/govuk_error_summary", form: form
+  = render "business_heading", business: @business
+  h2.govuk-heading-l
```
broder: Some of these headings are within a `.govuk-grid-row` and some aren't - does that matter?

jasiekmiko: Yes, I've noticed that and want to raise a separate ticket to clean that up.
I think it's mostly harmless*, but would be good to standardise one approach

(*) as long as it's outside of divs, or it's inside *both* a row and a column - being in but not the other definitely produces weird layout.

broder: 👍

jasiekmiko: Turns out yes, we want the entire content of the page to be in, including the error summaries - I've raised a PR for it here: https://github.com/UKGovernmentBEIS/beis-mspsds/pull/643

PR #636, opened 2019-02-21
MSPSDS-937: Create new case button ts

kicferk1:


mspsds-web/app/views/layouts/_navbar.html.slim
```diff
@@ -12,3 +12,6 @@ nav.app-navigation.govuk-clearfix
   - if User.current.is_opss?
     .creation
       = link_to "Create new", new_investigation_path, class: "govuk-button govuk-button--start", role: "button"
+  - else
+    .creation
+      = link_to "Create new", new_ts_investigation_path, class: "govuk-button govuk-button--start", role: "button"
```
broder: Should we move the conditional to just be for the URL? i.e. `User.current.is_opss? ? new_investigation_path : new_ts_investigation_path`

shared-web/app/assets/application/stylesheets/navbar.scss
```diff
@@ -50,7 +50,7 @@
   .creation {
     position: absolute;
     right: 4px;
-    top: 4px;
+    top: 15%;
```
broder: What's the rationale on going for `%` rather than `px`?

kicferk1: There is none seeing how we are using px everywhere.

PR #643, opened 2019-02-22
MSPSDS-888 Move error summary into two-thirds columns

jasiekmiko:
Also in changed places move to using
.govuk-grid-column-two-thirds-from-desktop to use real estate on
smaller screens better

Add error summary to new activity page. It doesn't actually use a model
to record its error, so we're using the component directly, rather than
through the form glue


jasiekmiko:
Waiting on final confirmation from Ed re width of error summary on pages which use the full width for their headers

mspsds-web/app/views/investigations/meetings/context.html.slim
```diff
@@ -1,24 +1,23 @@
...
+      fieldset.govuk-fieldset
+        legend.govuk-fieldset__legend.govuk-fieldset__legend--l
+          | Who was the meeting with?
         .govuk-form-group
-          fieldset.govuk-fieldset aria-describedby="correspondence-date-hint" role="group"
-            legend.govuk-fieldset__legend.govuk-fieldset__legend--l
-              | Date of meeting
-          = render "date_input", form: form
-        = form.submit "Continue", class: "govuk-button"
+          = form.label :correspondent_name, "Name", class: "govuk-label govuk-!-width-two-thirds"
```
broder: Should we be removing more of these `govuk-!-width-two-thirds` classes?

mspsds-web/app/views/investigations/products/new.html.slim
```diff
@@ -3,10 +3,10 @@
   = link_to "Back to #{@investigation.pretty_description}", @investigation, class: "govuk-back-link"
 .investigation-product-page
```
broder: I'm not sure this is used anywhere anymore

mspsds-web/app/views/investigations/meetings/context.html.slim
jasiekmiko: urrr no I think I went a bit over the top in a couple of the forms - we still want the fields to be narrowed in relation to the form. I've reverted those now :)

mspsds-web/app/views/investigations/products/new.html.slim
jasiekmiko: good shout

mspsds-web/app/views/investigations/creation_flow/allegation_details.html.slim
```diff
@@ -8,7 +8,7 @@
 
       h1.govuk-heading-xl
         = @page_title
-
+
```
broder: Slim lint might complain about the trailing whitespace?

jasiekmiko: oh thanks! shouldn't have resolved merges within github 😅

PR #648, opened 2019-02-25
MSPSDS-961: Update text area limits

kicferk1:
Includes a fix for forms generating 2 characters(\r\n) for one end of line. I put it in relevant models, but I'm not sure it belongs there, perhaps it should be fixed in controllers instead. 

Most text areas are now 10k characters long, except tests details, corrective action details and correspondence details(which are 50k), because it's likely they will include some copy-pasting from external sources. 

Also, there was one text area without server-side length validation, attachment description. Should I add one for it?


mspsds-web/app/models/comment_activity.rb
```diff
@@ -17,4 +18,12 @@ def search_index
   def email_update_text
     "#{source&.show&.titleize} commented on the #{investigation.case_type}."
   end
+
+private
+
+  # Browsers treat end of line as one character when checking input length, but send it as \r\n, 2 characters
+  # To keep max length consistent we need to reverse that
+  def trim_end_line
+    self.body = body.gsub("\r\n", "\n") if self.body
```
broder: Does `body.gsub!(...)` work? Generally ruby methods with `!` at the end modify the variable in place

mspsds-web/app/models/complainant.rb
```diff
@@ -19,4 +20,13 @@ def can_be_displayed?
 
     false
   end
+
+private
+
+  # Browsers treat end of line as one character when checking input length, but send it as \r\n, 2 characters
+  # To keep max length consistent we need to reverse that
+  def trim_end_line
```
broder: We could do this with some metaprogramming / magic on the base `ApplicationRecord` such that we have something like like `before_validation :name_trim_line_endings, :other_details_trim_line_endings` and it does what we want, meaning we don't have to write a `def trim_end_line` everywhere... How magic are you feeling?

mspsds-web/app/views/application/_related_attachment_fields.html.slim
```diff
@@ -14,4 +14,4 @@ fieldset.govuk-fieldset
       #attachment-description.govuk-form-group
           = subform.label :description, "Attachment description", class: "govuk-label"
           = subform.text_area :description, value: file_blob&.metadata&.dig(:description),
-                class: "govuk-textarea", rows: 5, maxlength: 1000
+                class: "govuk-textarea", rows: 5, maxlength: 10000
```
broder: Is now a good chance to make components for https://design-system.service.gov.uk/components/text-input/ and https://design-system.service.gov.uk/components/textarea/? @jasiekmiko, what do you think?

mspsds-web/app/models/complainant.rb
kicferk1: Magic enough, just point me in the right direction :D

mspsds-web/app/views/application/_related_attachment_fields.html.slim
kicferk1: Definitely a good chance to make text_area component, since there will definitely be more work coming up on that(like adding character count), as for text-input, maybe? It doesn't feel the two are strongly related in our code.

mspsds-web/app/models/complainant.rb
jasiekmiko: or `before_validation trim_line_endings(:name)`? choose your magic spectrum :')

mspsds-web/app/views/application/_related_attachment_fields.html.slim
jasiekmiko: If we add a character count, we should probably use https://design-system.service.gov.uk/components/character-count/, which in turn [uses a textarea](https://github.com/alphagov/govuk-frontend/blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/character-count/template.njk#L7) component - so yes, I would be quite keen to get the a textarea component in quite soon, but I'm also conscious that it is orthogonal to the to concerns of this ticket. It'd make more sense to finish this ticket as is and actually just invest some time in https://regulatorydelivery.atlassian.net/browse/MSPSDS-1007

mspsds-web/app/models/complainant.rb
broder: Can you pass arguments in `before_validation` hooks? I was thinking of something like https://stackoverflow.com/a/38233191

jasiekmiko: I think as long as `trim_line_endings(:name)` spits out something that `before_validation` accepts we should be good, and according to https://api.rubyonrails.org/classes/ActiveRecord/Callbacks.html (Types of Callbacks) it can take symbols but also lambda etc., so there are definitely some ways to achieve that kind of syntax :)

mspsds-web/app/views/application/_related_attachment_fields.html.slim
broder: (that's assuming that MSPSDS-1007 ever gets prioritised!)

PR #653, opened 2019-02-25
COSBETA-86: Install required node packages when building Cosmetics Worker

tom-bell-softwire:
The country list from the GOV.UK country and territory autocomplete package is used in helper methods that are called by the file analyzers on Cosmetics Worker.

For now, the `package.json` file is shared with Cosmetics Web to ensure the same dependency versions are used on both.

### DIFF
--- a/cosmetics-worker/docker-entrypoint.sh
+++ b/cosmetics-worker/docker-entrypoint.sh
@@ -4,6 +4,9 @@ set -ex
 # Ensure all gems are installed.
 bin/bundle check || bin/bundle install
 
+# Ensure all node packages are installed.
+yarn install
+
 # freshclam returns code 1 when it's up-to-date...
 freshclam || EXIT_CODE=$?
 if [[ $EXIT_CODE > 1 ]]; then
--- a/cosmetics-worker/manifest.yml
+++ b/cosmetics-worker/manifest.yml
@@ -7,7 +7,7 @@ applications:
     - ruby_buildpack
   instances: 1
   path: ../cosmetics-web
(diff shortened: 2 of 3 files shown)


cosmetics-worker/docker-entrypoint.sh
```diff
@@ -4,6 +4,9 @@ set -ex
 # Ensure all gems are installed.
 bin/bundle check || bin/bundle install
 
+# Ensure all node packages are installed.
+yarn install
```
broder: Will this be run on PaaS as well?

tom-bell-softwire: Very good point! I'll reassess how best to fix this issue...

cosmetics-worker/manifest.yml
```diff
@@ -7,7 +7,7 @@ applications:
     - ruby_buildpack
   instances: 1
   path: ../cosmetics-web
-  command: freshclam --config-file=clamav/freshclam.conf && freshclam --config-file=clamav/freshclam.conf -d -c 2 && (../deps/0/apt/usr/sbin/clamd -c clamav/clamd.conf &) && bin/sidekiq -C config/sidekiq.yml
+  command: bin/yarn install && freshclam --config-file=clamav/freshclam.conf && freshclam --config-file=clamav/freshclam.conf -d -c 2 && (../deps/0/apt/usr/sbin/clamd -c clamav/clamd.conf &) && bin/sidekiq -C config/sidekiq.yml
```
tom-bell-softwire: I couldn't see a way to run just the `yarn install` part of the asset compilation step in the [Ruby buildpack](https://github.com/cloudfoundry/ruby-buildpack/blob/master/src/ruby/finalize/finalize.go#L213), so this seems like the simplest approach.

PR #658, opened 2019-02-26
MSPSDS-1007: Text area component

kicferk1:
The component borrows many ideas from autocomplete select one, especially for error display, labels and hints. Might be worth commonizing now or when doing next component with such a similar structure(form_group with validation, label, hint and some input).


shared-web/app/views/form_components/_text_area.html.slim
```diff
@@ -0,0 +1,22 @@
+ruby:
```
broder: Should we be adding something to `components` (e.g. `components/_govuk_textarea.html.slim`) and be basing it off https://github.com/alphagov/govuk-frontend/blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/textarea/template.njk?

We could then re-use e.g. `_govuk_hint` for the `local_assigns[:hint]` bit below.

jasiekmiko: Yes, definitely how we want to approach this! Happy to help out if needed! :)

shared-web/app/views/components/_govuk_textarea.html.slim
```diff
@@ -0,0 +1,37 @@
+/ Implementation based on
+/ https://github.com/alphagov/govuk-frontend/
+/ blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/textarea/template.njk
+ruby:
+  described_by = ""
+  has_hint = local_assigns[:hint].present?
+  has_hint = has_hint && (local_assigns[:hint][:text].present? || local_assigns[:hint][:html].present?) if has_hint
```
jasiekmiko: Don't think we need both `has_hint && ...` and `... if has_hint` here?

shared-web/app/views/form_components/_text_area.html.slim
```diff
@@ -1,22 +1,19 @@
+/ This partial acts as glue between rails form helpers and the component textarea
```
jasiekmiko: Can we rename this file to `_govuk_textarea` to match the system?

shared-web/app/views/components/_govuk_textarea.html.slim
```diff
@@ -0,0 +1,37 @@
...
+  has_error = local_assigns[:errorMessage].present?
+  rows = local_assigns[:rows] || 5
+
+  form_group_classes = [local_assigns.dig(:formGroup, :classes)]
+  form_group_classes.push "govuk-form-group--error" if local_assigns[:errorMessage]
+  area_classes = ["govuk-textarea", local_assigns[:classes]]
+  area_classes.push("govuk-textarea--error") if local_assigns[:errorMessage].present?
+
+.govuk-form-group class=form_group_classes id=local_assigns[:group_id]
+  = render "components/govuk_label", for: local_assigns[:id], **local_assigns[:label]
```
jasiekmiko: `id` and `label` a required (after https://design-system.service.gov.uk/components/textarea/ nunjucks options), so we can actually use them directly here with the added benefit of getting an error when they're missing:
`= render "components/govuk_label", for: id, **label`

shared-web/app/views/components/_govuk_textarea.html.slim
```diff
@@ -0,0 +1,37 @@
...
+    = render "components/govuk_hint", id: hint_id, **local_assigns[:hint]
+
+  - if has_error
+    - error_id = local_assigns[:id] + '-error'
+    - described_by = described_by.present? ? described_by + ' ' + error_id : error_id
+    = render "components/govuk_error_message", id: error_id, **errorMessage
+
+  textarea[*(local_assigns[:attributes] || {})
+                class=area_classes
+                id=local_assigns[:id]
```
jasiekmiko: ditto, this can be `id=id`

shared-web/app/views/components/_govuk_textarea.html.slim
```diff
@@ -0,0 +1,37 @@
...
+                rows=rows
+                name=local_assigns[:name]
```
jasiekmiko: ... and this can be `name=name`

shared-web/app/views/form_components/_text_area.html.slim
```diff
@@ -1,22 +1,19 @@
...
+/ With access to the form object, we can infer a lot of information from it
+
 ruby:
+  # Infer name and id_prefix from attribute key
+  local_assigns[:name] ||= get_attribute_name(form, key)
+  local_assigns[:id] ||= get_attribute_id_prefix(form, key)
+  local_assigns[:value] ||= form.object.send(key) if form.object.respond_to?(key)
+  local_assigns[:max_length] ||= 10000
```
jasiekmiko: Where does this default come from?
If it's custom to `mspsds`, then I'd be tempted to not include it in the component and simply make sure that it's supplied at use sites to keep the logic here agnostic to the domains..

shared-web/app/views/form_components/_text_area.html.slim
```diff
@@ -1,22 +1,19 @@
...
+  local_assigns[:label][:classes] ||= "govuk-label" if local_assigns[:label].present?
```
jasiekmiko: I don't think we need this - the label component [take care of it](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/e89c8d41519c45d3b3fdb814ea4255c53539ca05/shared-web/app/views/components/_govuk_label.html.slim#L6)

shared-web/app/views/form_components/_text_area.html.slim
```diff
@@ -1,22 +1,19 @@
...
+  # TODO: govuk component seems to require label, but we don't always define it. Perhaps we should?
```
jasiekmiko: Yes, we should always have a label that describes the purpose of the element. Where it's missing right now, I suggest we add one with `govuk-visually-hidden` class applied, so we can commit this component without this workaround

shared-web/app/views/components_gallery/textarea.html.slim
```diff
@@ -0,0 +1,64 @@
+/ Exmpales in this file are based on  https://github.com/alphagov/govuk-frontend/
+/           blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/radios/radios.yaml
```
jasiekmiko: Needs updating :)

shared-web/app/views/components/_govuk_textarea.html.slim
kicferk1: It works(as in, it runs and doesn't throw errors), but my ruby mine is unhappy with the name 'label', confusing it with a form method :( Weird.

jasiekmiko: silly ruby :P

mspsds-web/app/views/investigations/ts_investigations/_file_page.html.slim
```diff
@@ -21,8 +21,7 @@ fieldset.govuk-fieldset
...
               = subform.text_field :title, class: "govuk-input govuk-!-width-two-thirds", value: @file_title
-            .govuk-form-group class=("govuk-form-group--error" if form.object.errors[:description].any?)
-              = subform.label :description, class: "govuk-label"
-                | Description
-              = subform.text_area :description, class: "govuk-textarea", rows: 5, maxlength: 10000,
-                      value: @file_description
+            = render "form_components/govuk_textarea", key: :description, form: subform, value: @file_description,
+                    max_length: 10000, label: { text: "Description" },
+                    hint: { text: "For example, 'Orange, 4 socket extension lead'" },
+                    label: { text: "File description", classes: "govuk-visually-hidden" }
```
broder: Do we need 2 labels on this element?

PR #660, opened 2019-02-27
COSBETA-81: Enable logging and security fixes

tom-bell-softwire:
This applies the changes from PRs #246 and #371 to Cosmetics Web, by moving most of the relevant dependencies and configuration into Shared Web.

@jasiekmiko, @broder – this is in preparation for our upcoming pen test, so if you can think of anything else that was done for MSPSDS that we should be doing here, please let me know.


cosmetics-web/config/application.rb
```diff
@@ -33,6 +33,10 @@ class Application < Rails::Application
     config.active_job.queue_adapter = :sidekiq
     config.action_mailer.deliver_later_queue_name = 'cosmetics-mailers'
 
+    # Set the request timeout in seconds. The default set by Slowpoke is 15 seconds.
+    # Use a longer timeout on development environments to allow for asset compilation.
+    Slowpoke.timeout = Rails.env.production? ? 15 : 180
```
jasiekmiko: Do we still need this to be `180`? It feels like asset compilation has really sped up recently-  though I guess there's not really much harm in it?

tom-bell-softwire: I don't think there's much harm in allowing a longer timeout, even if it isn't hit much in practice. If it prevents the occasional system test from timing out on Travis, it'll be worth having it.

tom-bell-softwire:
Thanks for the link to those recommendations. I think we've implemented most of the broad recommendations on there. We might want to come back and look at page-level things, like preventing caching for sensitive information (such as product formulation details), but those pages don't exist yet.

PR #661, opened 2019-02-27
MSPSDS-6: Update sidekiq configuration and enable sidekiq UI

broder:
I think we've got some jobs on MSPSDS production in the dead set (after failing several times if / when the AV was down). Enabling the UI should allow us to retrigger them. I've also bumped up the number of retries to the default (25) which should hopefully reduce the number of jobs which end up in the dead set.

@tom-bell-softwire, I'm assuming that cosmetics will want to follow the same process.

https://github.com/mperham/sidekiq/wiki/Error-Handling#best-practices


mspsds-web/config/sidekiq.yml
```diff
@@ -1,6 +1,5 @@
 ---
 :concurrency: 10
-:max_retries: 3
```
jasiekmiko: @MatthewBurstein FYI look like the same change you were looking to make, so looks like no need to have different configs for different jobs for the time being :)

mspsds-web/README.md
```diff
@@ -181,6 +181,11 @@ See the Sentry account section in [the root README](../README.md#sentry) to get
 
 This enables and adds basic auth to the health check endpoint at `/health/all` which can be polled to check the site status.
 
+    cf set-env mspsds-web SIDEKIQ_USERNAME XXX
+    cf set-env mspsds-web SIDEKIQ_PASSWORD XXX
+
+This enables and adds basic auth to the sidekiq monitoring UI at `/sidekiq` which can be used to check the worker performance.
```
jasiekmiko: Where will those be stored, the prototypes page again? If I was following this readme I'd find it helpful to have a note explaining our approach, e.g. inidicating if I'm expected to generate these credentials afresh every time, where to get/store them, etc.

broder: Yeah, we should probably rename the page to "Credentials" or something like that. I guess you would be expected to generate them fresh every time (i.e. when you're setting up a new environment from scratch)?

mspsds-web/config/sidekiq.yml
MatthewBurstein: Yes this is the same change. Thanks for flagging :)

broder:
@tom-bell-softwire this PR does also cover cosmetics - I just wanted to check that you were happy with the changes!

PR #662, opened 2019-02-27
MSPSDS-957: Add confirmation email on case creation

MatthewBurstein:


mspsds-web/app/models/investigation.rb
```diff
@@ -244,6 +245,14 @@ def create_audit_activity_for_removing_business(business)
   def assign_current_user_to_case
     self.source = UserSource.new(user: User.current) if self.source.blank? && User.current
   end
+
+  def send_confirmation_email
+    NotifyMailer.investigation_created(pretty_id,
```
broder: Should we check that `User.current` isn't `nil`?

mspsds-web/app/mailers/notify_mailer.rb
```diff
@@ -12,4 +12,19 @@ def updated_investigation(investigation_pretty_id, name, email, update_text, sub
 
     mail(to: email)
   end
+
+  def investigation_created(investigation_pretty_id, name, email, investigation_title, investigation_type)
```
broder: Should we keep the method name consistent with the one above (or change the one above if you feel strongly about it)? i.e. `updated_investigation` vs `investigation_created`

MatthewBurstein: Yes good point :)

mspsds-web/app/models/investigation.rb
MatthewBurstein: Why would `User.current` be `nil`? Surely noone should be able to access this if they aren't signed in?

broder: Yes, hopefully. The RAPEX import would break though I think. (Although it probably would break in more ways than just this, and we should probably just delete it...)

MatthewBurstein: Have added the check.

mspsds-web/app/models/investigation.rb
```diff
@@ -22,7 +22,8 @@ class Investigation < ApplicationRecord
   validates_length_of :hazard_description, maximum: 10000
 
   after_save :create_audit_activity_for_assignee,
-             :create_audit_activity_for_status, :create_audit_activity_for_visibility
+             :create_audit_activity_for_status, :create_audit_activity_for_visibility,
+             :send_confirmation_email
```
kicferk1: Is that going to send an email every time investigation is updated(and hence saved) as well? Maybe we can use after_created hook?

MatthewBurstein: yes good spot :)

broder:
@MatthewBurstein, could you add a line to the changelog to mention this functionality?

PR #663, opened 2019-02-27
MSPSDS-6: Add PR template and changelog

broder:
I'd expect people to add a note to the "Unreleased" section of the changelog when they open their pull request. After a production release, we'd take a cut of the "Unreleased" and label it with the date of the release (I'm not sure we want to start doing version numbers). Does that process seem sensible?


broder:
@tom-bell-softwire I guessed that this isn't useful for cosmetics yet - is that correct or should I include the recent cosmetics changes?

PR #665, opened 2019-02-28
COSBETA-125: Handle additional errors for file upload

raamSoftwire:
[COSBETA-125](https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-125)

This tickets adds 4 new errors for CPNP file upload

- notification missing data error (eg. no product name)
- notification validation error
- notification duplication error (eg. notification with CPNP reference # already exists for this RP)
- unknown error

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
                                           responsible_person: @notification_file.responsible_person)
         notification.notification_file_parsed!
         notification.save
+
+        if notification.errors.messages.present?
+          if notification.errors.messages[:cpnp_reference].include? "Notification duplicated"
```
tom-bell-softwire: Is there a risk that the text of this error could change, in which case we would no longer catch the error?

Could we instead modify the validation to add an error with a different key (`:duplicate_cpnp_reference` or something), that we could check for, so we wouldn't need to compare the error message itself?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
...
+            Rails.logger.error "File Upload Error: A notification for this product already
```
tom-bell-softwire: Worth including the CPNP reference number of the duplicate entry?

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
...
+          end
+        end
       end
     rescue UnexpectedPdfFileError
       @notification_file.update(upload_error: :unzipped_files_are_pdf)
     rescue ProductFileNotFoundError
       @notification_file.update(upload_error: :product_file_not_found)
-    rescue UnexpectedFileError
-      @notification_file.update(upload_error: :unzipped_files_not_xml)
+    rescue NotificationDuplicationError
```
tom-bell-softwire: Minor thing, but `DuplicateNotificationError` feels a bit more readable.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
...
         notification.notification_file_parsed!
         notification.save
+
+        if notification.errors.messages.present?
+          if notification.errors.messages[:cpnp_reference].include? "Notification duplicated"
+            Rails.logger.error "File Upload Error: A notification for this product already
+              exists for this responsible person"
+            raise NotificationDuplicationError
+          elsif notification.errors.messages.values.any? { |message| message.include?("must not be blank") }
+            missing_values = []
```
tom-bell-softwire: I thought we decided not to distinguish between missing properties and other validation errors?

While it's useful to log all the validation errors so we can investigate any issues that come up, I don't think it's useful to inform the user of this.

I'd just have a NotificationValidationError or similar, with an equivalent upload error state and generic message we present to the user.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
...
+            messages.each { |key, value| missing_values << key if value.include? "must not be blank"}
+            Rails.logger.error "File Upload Error: There is missing data in the
+              notification: #{missing_values.join(', ')}"
+            raise NotificationMissingDataError
+          else
+            Rails.logger.error "File Upload Error: Notification validation error"
```
tom-bell-softwire: Would be good to include the full list of error messages in this log.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -35,13 +35,34 @@ def create_notification_from_file
...
     rescue ProductFileNotFoundError
       @notification_file.update(upload_error: :product_file_not_found)
-    rescue UnexpectedFileError
-      @notification_file.update(upload_error: :unzipped_files_not_xml)
+    rescue NotificationDuplicationError
+      @notification_file.update(upload_error: :notification_duplicated)
+    rescue NotificationMissingDataError
+      @notification_file.update(upload_error: :notification_missing_data)
+    rescue NotificationValidationError
+      @notification_file.update(upload_error: :notification_validation_error)
```
tom-bell-softwire: Do we need a generic `rescue` block to catch any other unexpected errors?

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -45,6 +45,42 @@
...
 
+  it "shows an error when the uploaded file contains PDF files" do
+    visit new_responsible_person_notification_file_path(responsible_person)
+    page.attach_file('notification_file[uploaded_file]',
+                     Rails.root + 'spec/fixtures/testZippedPDF.zip')
+    click_button "Upload"
+    expect(page).to have_text("The unzipped files are PDF files")
+  end
+
+  it "shows an error when the uploaded file does not contain product XML files" do
```
tom-bell-softwire: `...contain a product XML file`

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -45,6 +45,42 @@
...
+    visit new_responsible_person_notification_file_path(responsible_person)
+    page.attach_file('notification_file[uploaded_file]',
+                     Rails.root + 'spec/fixtures/testNoProductFile.zip')
+    click_button "Upload"
+    expect(page).to have_text("The ZIP file does not contain product XML files")
+  end
+
+  it "shows an error when the uploaded file already exists for this RP" do
```
tom-bell-softwire: `when a product with the same CPNP reference already exists for this RP`

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -45,6 +45,42 @@
...
+    visit new_responsible_person_notification_file_path(responsible_person)
+    page.attach_file('notification_file[uploaded_file]',
+                     Rails.root + 'spec/fixtures/testExportFile.zip')
+    click_button "Upload"
```
tom-bell-softwire: Rather than upload twice, might be better to use the factory to create an existing Notification entry in the DB with the relevant CPNP reference.

cosmetics-web/spec/system/upload_single_file_spec.rb
```diff
@@ -45,6 +45,42 @@
     expect(page).to have_text("Uploaded file exceeds size limit")
   end
 
+  it "shows an error when the uploaded file contains PDF files" do
```
tom-bell-softwire: I wonder if it's worth splitting the error scenarios into a separate system spec file, to avoid this file getting too long?

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Good idea. Thanks

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: I agree

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Sure thing

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: We do indeed, I had one before then removed it during some gnarly debugging and forgot to put it back in. Great catch! Thanks

cosmetics-web/spec/system/upload_single_file_spec.rb
raamSoftwire: Done

cosmetics-web/spec/system/upload_single_file_spec.rb
raamSoftwire: Done

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Sorry, I'm not quite sure I understand here. If I updated the `validates` statement on the `notification` model to read

```
validates :cpnp_reference, uniqueness: { scope: :responsible_person, message: :duplicate_cpnp_reference }, :allow_nil => true
```

would that work?

raamSoftwire: And then pick up the `:duplicate_cpnp_reference` symbol in the analyzer instead of the string

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Ah, I misunderstood. I have now fixed that up

cosmetics-web/spec/system/upload_single_file_spec.rb
raamSoftwire: That's a much better idea, thanks

cosmetics-web/spec/system/upload_single_file_spec.rb
raamSoftwire: I agree, I've now split them out.

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Just FYI rubocop wants it to rescue `StandardError`, rather than not specifying an error type

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: That's sorted now, thanks for the help.

cosmetics-web/spec/system/upload_single_file_errors_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'rails_helper'
+
+RSpec.describe "Upload a single file errors", type: :system do
+  let(:responsible_person) { create(:responsible_person) }
+
+  before do
+    sign_in_as_member_of_responsible_person(responsible_person)
+    mock_antivirus
```
tom-bell-softwire: I think this needs changing to `mock_antivirus_api` to match the recent changes on master branch.

cosmetics-web/spec/system/upload_single_file_errors_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  after do
+    sign_out
+    unmock_antivirus
```
tom-bell-softwire: I think this needs changing to `unmock_antivirus_api` to match the recent changes on master branch.

PR #666, opened 2019-02-28
MSPSDS-979 team assignment page fixes

MatthewBurstein:
## Checklist:

### General testing
- [X] Test without javascript
- [X] Test on small screen
### Accessibility testing
- [X] Works keyboard only
- [X] Disable css - does content make sense and appear in a logical order?


mspsds-web/app/helpers/investigations_helper.rb
```diff
@@ -141,4 +141,11 @@ def teams_with_keys
   def assignee_ids_from_team(team)
     [team.id] + team.users.map(&:id)
   end
+
+  def suggested_previous_assignees
+    all_past_assignees = @investigation.past_assignees + @investigation.past_teams
+    return [] if all_past_assignees.empty? || all_past_assignees == [User.current]
+
+    all_past_assignees || []
```
kicferk1: Can all_past_assignees ever be falsy here?

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
```diff
@@ -5,19 +5,20 @@
   - if User.current.teams.any?
     - someone_in_your_team = capture do
       = render "investigations/assign/assignee_selection", form: form, key: :select_team_member,
-              items: User.get_team_members(user: User.current), accessible_hint: "Name", show_all_values: true
+              items: User.get_team_members(user: User.current), label: "Select team member", show_all_values: true
   - previously_assigned = capture do
-    = render "investigations/assign/assignee_selection", form: form, key: :select_previously_assigned,
-            items: investigation.past_assignees + investigation.past_teams, ignore_restrictions: true,
-            accessible_hint: "Select previously assigned person or team", show_all_values: true
+    - unless @suggested_previous_assignees.empty?
```
kicferk1: This is checked twice, once when generating the element(here), and again when adding element to items. Do we gain anything by that?

mspsds-web/app/helpers/investigations_helper.rb
MatthewBurstein: yes good point 👍

mspsds-web/app/views/investigations/assign/_assignee_form.html.slim
MatthewBurstein: Both operations should not be performed in the case when suggested_previous_assignees is empty. I'm not sure I see a more elegant way to structure it. Did you have something in mind?

kicferk1: I thought that we might just check it when constructing items. The result to end user should be the same, and I'm not sure capturing html is a costly enough operation to warrant checking the same thing twice. 

Alternatively, maybe you can compute previously_assigned html inside the ruby block?

jasiekmiko: I haven't tried this yet, buy I have a suspicion that the capture blocks can actually be done inside of the `ruby` blocks if that helps? I.e.
```
ruby:
  some_html = capture do
    render "something", with: options
  end
= some_html
```
would need checking that that's valid, but if it is then it could make the whole items building a little neater :)

mspsds-web/app/views/investigations/assign/_assignee_selection.html.slim
```diff
@@ -4,6 +4,5 @@ ruby:
 
 .govuk-grid-row
   .govuk-grid-column-one-half
-    = form.label key, "Assignee", class: "govuk-label"
     = render "form_components/autocomplete_select", key: key, show_all_values: local_assigns[:show_all_values],
-            items: actual_items, form: form, hint: { text: accessible_hint, classes: "govuk-visually-hidden" }
+            items: actual_items, form: form, label: { text: label, classes: "govuk-visually-hidden" }
```
jasiekmiko: Cool, this is much nicer! :)

PR #669, opened 2019-02-28
MSPSDS-1028: Team inboxes

kicferk1:
## Description
It seems difficulty getting group attributes is not due to the gem we are using, but rather keycloak api. When calling GET endpoint for specific group it gives attributes, when calling endpoint for all groups it doesn't. 

That means an api call for each team we are interested in notifying. I'm not sure how bad doing them all when loading teams could be for performance. If terrible perhaps we can call it each time we try to notify a team separately to distribute the calls. 

## Checklist:
### General testing
- [x] Test without javascript(not relevant for this ticket)
- [x] Test on small screen(not relevant for this ticket)
### Accessibility testing
- [x] Works keyboard only(not relevant for this ticket)
- [x] Tested with one screen reader(not relevant for this ticket)
- [x] Zoom page to 400% - content still visible(not relevant for this ticket)
- [x] Disable css - does content make sense and appear in a logical order?(not relevant for this ticket)


mspsds-web/app/models/activity.rb
```diff
@@ -34,17 +34,33 @@ def sensitive_body?
   end
 
   def notify_relevant_users
```
broder: This is a little hard to follow. Do you think we could do something like `entities_to_notify` which is an array of `{name: "xxx", email: "xxx"}`? Then in this method we'd just `uniq` and iterate over that array, sending an email to each.

`users_to_notify` and `teams_to_notify` would be responsible for giving us the entities in that format.

mspsds-web/app/models/audit_activity/investigation/update_assignee.rb
```diff
@@ -29,16 +29,27 @@ def email_subject_text
 
   def users_to_notify
     previous_assignee_id = investigation.saved_changes["assignable_id"][0]
```
broder: It feels like there should be some way to commonise this method and `teams_to_notify`...

mspsds-web/app/models/activity.rb
kicferk1: I think `teams_to_notify` is better off returning a team. That way `users_to_notify` doesn't need to know about team logic.

shared-web/app/clients/shared/web/keycloak_client.rb
```diff
@@ -58,7 +67,8 @@ def all_teams
         teams = []
         organisations["subGroups"].reject(&:blank?).each do |organisation|
           organisation["subGroups"].reject(&:blank?).map do |team|
-            teams << { id: team["id"], name: team["name"], path: team["path"], organisation_id: organisation["id"] }
+            team_recipient_email = group_attributes(team["id"])["teamRecipientEmail"]&.first
```
broder: Should the `teamRecipientEmail` attribute be snake_case? That's what we used for `mobile_number` but I don't know what the standard is for keycloak.

Should we also add an example `teamRecipientEmail` to some of our teams in our `initial-setup.json`?

kicferk1: All json fields names from keycloak export use camelCase, but I'm not strongly attached to that. If you think it's more consistent to have our attributes be stored in the format they will be called eventually then I can change it.

broder: Happy to leave it as it is.

PR #670, opened 2019-02-28
MSPSDS-987 breadcrumbs component

MatthewBurstein:
## Description

- Added component for breadcrumbs to shared-web
- mspsds-web requires a wrapper around this component because in some cases the breadcrumbs are replaced by a `back_to_investigation` link

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


shared-web/app/views/components/_govuk_breadcrumbs.html.slim
```diff
@@ -0,0 +1,16 @@
+/ Implementation based on
+/ https://github.com/alphagov/govuk-frontend/
+/     blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/breadcrumbs/template.njk
+
+- attributes = (local_assigns[:attributes] || {})
+.govuk-breadcrumbs *attributes class=local_assigns[:classes]
+  ol class="govuk-breadcrumbs__list"
+    - local_assigns[:items].each do |item|
```
jasiekmiko: Since `items` are required, this can just say `- items.each do |item|`.
The local variables we pass in to partials are available directly like this. `local_assigns` is a special way of accessing them that doesn't blow up if they're `nil` :)

shared-web/app/views/components/_govuk_breadcrumbs.html.slim
```diff
@@ -0,0 +1,16 @@
...
+      - item_content = capture do
+        = item[:html.presence] || item[:text]
```
jasiekmiko: `item[:html] || item[:text]`
The `.presence` is in the wrong place atm, but it's also just not needed :)

shared-web/app/views/components/_govuk_breadcrumbs.html.slim
```diff
@@ -0,0 +1,16 @@
...
+      - if item[:href]
+        li class="govuk-breadcrumbs__list-item"
+          = link_to item_content, item[:href], class: "govuk-breadcrumbs__link", **attributes
```
jasiekmiko: `attributes` here should be `(item[:attributes] || {})`, not the same attributes that go on the breadcrumbs div!
(Annoyingly, that option is missing from the docs here https://design-system.service.gov.uk/components/breadcrumbs/#options-example-default--items :(. I believe this is an omission [here](https://github.com/alphagov/govuk-frontend/blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/breadcrumbs/breadcrumbs.yaml#L6), in case you wanted to fix that upstream :) )

shared-web/app/views/components/_govuk_breadcrumbs.html.slim
MatthewBurstein: 👍

shared-web/app/views/components/_govuk_breadcrumbs.html.slim
MatthewBurstein: 👍

PR #671, opened 2019-02-28
MSPSDS-1046 List team members

jasiekmiko:
## Description
Add teams pages with membership lists.

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


jasiekmiko:
(Pending content / styling tweaks from design)

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,24 @@
+class TeamsController < ApplicationController
+  before_action :set_my_teams, only: :my_teams
+  before_action :set_team, only: :show
+
+  def index
+    redirect_to :my_teams
+  end
+
+  def my_teams
+    render :index
```
kicferk1: Why do we need my_teams action(that only stets teams and renders index)? Does it generate a different(prettier) url, or is there some other reason for it?

mspsds-web/app/views/teams/show.html.slim
```diff
@@ -0,0 +1,21 @@
...
+    h2.govuk-heading-l Members
+    ul.govuk-list
+      - @team.users.each do |user|
+        li.teams--user class="govuk-!-padding-top-3 govuk-!-padding-bottom-3"
+          h3.govuk-heading-s class="govuk-!-margin-bottom-2"
+            span.teams--user-name = user.full_name
+            '
+            span.govuk-caption-m.teams--user-email = user.email
+          ul.govuk-list class="govuk-!-margin-0"
+            li
```
kicferk1: Why do we need a new list with one element here? Styling?

mspsds-web/test/controllers/organisation_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class OrganisationControllerTest < ActionDispatch::IntegrationTest
```
kicferk1: Is this gonna get populated later in this pull request? :)

mspsds-web/app/controllers/teams_controller.rb
jasiekmiko: Yeah, it's for the sake of having the url as `/my-teams`. But because we're also exposing `/teams/:id`, `/teams` shouldn't be a 404, so I'm including both. Effectively `my_teams` is the important action, with `index` being just a redirect

mspsds-web/app/views/teams/show.html.slim
jasiekmiko: Semantics more than anything, and inclination that in the future that list of privileges would grow

mspsds-web/test/controllers/organisation_controller_test.rb
jasiekmiko: oops, no, not enough here to test meaningfully. Didn't meant add this file here, but I'm also about to use it on the followup tickets, so will probably just sneakily leave it in 😅

mspsds-web/app/views/layouts/_navbar.html.slim
```diff
@@ -1,14 +1,17 @@
...
       = link_to "Products", products_path, "data-topnav": "products", class: "govuk-link--no-visited-state"
-  .creation
-    = link_to "Create new", User.current.is_opss? ? new_investigation_path : new_ts_investigation_path,
-            class: "govuk-button govuk-button--start", role: "button"
+    li.right-aligned class=(params[:controller] == "teams" ? "app-navigation--current-page" : nil)
+      - if User.current.teams.count > 1
+        = link_to "My teams", my_teams_path, "data-topnav": "products", class: "govuk-link--no-visited-state"
+      - else
+        - team = User.current.teams.first
+        = link_to team.display_name, team_path(team), "data-topnav": "products", class: "govuk-link--no-visited-state"
```
kicferk1: Many tests broken here, possibly because of team-less user? Pretty sure all users in tests should have teams, so it might be that our stubbing is causing keycloak to throw an error half-way through importing teams or team memberships. 

In any case, team-less users can probably happen, if not intentionally then by accident, and we should probably not crash the site when that happens :D

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,24 @@
+class TeamsController < ApplicationController
+  before_action :set_my_teams, only: :my_teams
+  before_action :set_team, only: :show
+
+  def index
```
broder: Have we split out `index` and `my_teams` because we're going to do something different with `index` later?

mspsds-web/app/views/teams/show.html.slim
```diff
@@ -0,0 +1,21 @@
...
+    h1.govuk-heading-xl = @team.display_name
+    h2.govuk-heading-l Members
+    ul.govuk-list
+      - @team.users.each do |user|
+        li.teams--user class="govuk-!-padding-top-3 govuk-!-padding-bottom-3"
+          h3.govuk-heading-s class="govuk-!-margin-bottom-2"
+            span.teams--user-name = user.full_name
+            '
+            span.govuk-caption-m.teams--user-email = user.email
+          ul.govuk-list class="govuk-!-margin-0"
```
broder: Having the list within a list seems a bit weird to me but I'm guessing that this has come from Notify?

mspsds-web/test/controllers/organisation_controller_test.rb
broder: We can probably get rid of this.

shared-web/app/clients/shared/web/keycloak_client.rb
```diff
@@ -124,8 +127,13 @@ def user_info
         { id: user["sub"], email: user["email"], groups: user["groups"], first_name: user["given_name"], last_name: user["family_name"] }
       end
 
-      def has_role?(role)
-        @client.has_role? role
+      def has_role?(user_id, role)
+        if User.current && (User.current.id == user_id)
+          # This is faster, as it uses the already fetched claims
+          @client.has_role? role
```
broder: Neat!

shared-web/app/clients/shared/web/keycloak_client.rb
```diff
@@ -124,8 +127,13 @@ def user_info
...
+        else
+          @internal.has_role? user_id, role
```
broder: Do you think we should do some kind of caching on this? How often are we hitting it?

mspsds-web/app/controllers/teams_controller.rb
broder: Should we just route `/my-teams` to `index` then?

jasiekmiko: As in have one action, but in routes have `get `my-teams` => `teams#index`?
Hmm, yeah, that does seem like a better way of what I'm trying to do 😅
I think `teams` might become a different page in the future (listing all teams, not just "mine"), but that's not why I've split it out - definitely worth crossing that when we get there!

broder: 👍

mspsds-web/app/views/teams/show.html.slim
broder: I was leaning more towards the question that Michal asked below about a list with one element - happy with both of your answers!

shared-web/app/clients/shared/web/keycloak_client.rb
broder: I had thought we could use the `Rails.cache` like we do for some of the other methods in this class. It's probably not going to be hit very often though...

mspsds-web/app/views/layouts/_navbar.html.slim
jasiekmiko: Yeah, the test were tricky - one simple solution (which you identified well!) and actually uncovering a problem with how investigations policy tests were set up!

shared-web/app/clients/shared/web/keycloak_client.rb
jasiekmiko: I'll leave it as is for now, I'm wary that introducing caching may complicate some test further (like the need for this cache clear does: https://github.com/UKGovernmentBEIS/beis-mspsds/blob/b218a3f62560b7a9ea50da339cfd93eb4d7944e7/mspsds-web/test/test_helper.rb#L235). Happy to come back to this if performance manifests itself as an issue

jasiekmiko:
This now includes the design tweaks as well as instructions for setting up the correct permission on the client user. Final form ready for review.

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  def show; end
+
+private
+
+  def set_user_teams
+    @teams = User.current.teams
+  end
+
+  def set_team
+    @team = Team.find(params[:id])
```
broder: Do we need any authorisation on this action?

jasiekmiko: Ahh... I went back and forth on it given there were thougths from Ed that we might want to expose these pages in general down the line. Eventually it slipped off my radar, but as long as we don't have a way of accessing them, they should probably be restricted to members only

PR #672, opened 2019-03-01
MSPSDS-6: Move antivirus to separate service

broder:
## Description
This should hopefully make everything more stable. 🤞

I'll add deployment from travis later

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


docker-compose.yml
```diff
@@ -4,16 +4,14 @@ services:
   db:
     build:
       context: ./db
-      dockerfile: Dockerfile
```
jasiekmiko: Not sure I understand this change? Is it just something that turned out to be superfluous? convention?

broder: "Dockerfile" is the default value so we don't need it. In other places, the `dockerfile` isn't in the same folder as the context. I can explain it in person if you still have any questions.

PR #674, opened 2019-03-01
MSPSDS-NONE: Research bugfix

kicferk1:
Remove string sorting to prevent weird ES behaviour that I don't fully understand.


mspsds-web/app/models/concerns/searchable.rb
```diff
@@ -14,12 +14,7 @@ module Searchable
           strings: {
             match_mapping_type: "string",
             mapping: {
-              type: "text",
-              fields: {
-                sort: {
-                  type: "icu_collation_keyword"
-                }
-              }
+              type: "text"
```
broder: @tom-bell-softwire, does this change mean that we can remove `analysis-icu` from the elasticsearch Dockerfile?

tom-bell-softwire: If we're no longer using this to fix the sort order, then yes – that ES plugin can go. (Which I think means we can return to using a vanilla image and get rid of the Dockerfile entirely.)

PR #677, opened 2019-03-03
COSBETA-32: Add initial pages for Poison Centre users

tom-bell-softwire:
This adds dedicated product list and detail pages for all registered products, which are only accessible to Poison Centre users (determined by their membership of the "Poison Centres" group in Keycloak).

The existing new and edit notification actions have also been moved to the nested notifications resource, so that they are more clearly associated with a Responsible Person.

## Checklist:
### General testing
- [x] Test without Javascript
- [x] Test on small screen
### Accessibility testing
Not applicable, as current page layouts are simply placeholders until we have detailed designs.


cosmetics-web/app/views/application/_unfinished_product.html.slim
```diff
@@ -0,0 +1,19 @@
...
+    dd
+      = notification.reference_number_for_display
+
+    - if notification.cpnp_reference.present?
+      dt
+        | EU reference number:
+      dd
+        = notification.cpnp_reference
+  p.add-documents
+    = link_to "Add formulation document to complete registration", "#",
```
broder: Have we got this page yet?

cosmetics-web/app/views/notifications/_product_details.html.slim
```diff
@@ -0,0 +1,45 @@
+- notification = local_assigns[:notification]
```
broder: Can we not just use `notification` without this?

cosmetics-web/app/policies/notification_policy.rb
```diff
@@ -0,0 +1,23 @@
+class NotificationPolicy < ApplicationPolicy
```
broder: Should we have a separate `ResponsiblePersonNotificationPolicy` and `NotificationPolicy` (for the poison centres)?

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -9,16 +9,30 @@ class ApplicationController < ActionController::Base
...
   before_action :set_current_user
   before_action :set_raven_context
+
+  before_action :redirect_poison_centre_user
   before_action :create_or_join_responsible_person
 
-  def create_or_join_responsible_person
+private
+
+  def redirect_poison_centre_user
```
broder: Should we just do this redirect once on login? And then let the poison centres see 401s for the other routes?

cosmetics-web/app/views/responsible_persons/notifications/edit.html.slim
```diff
@@ -0,0 +1,98 @@
+- content_for :page_title, "Check your answers"
+- content_for :after_header do
+  = link_to "Back", responsible_person_notifications_path(@responsible_person, anchor: "unfinished"),
+          class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      | Check the cosmetic product details before registering
+    = render "error_summary", errors: @notification.errors
```
broder: Should we be using `govuk_error_summary` here instead? Also, we've recently updated MSPSDS such that the errors appear above the heading (e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/views/investigations/tests/details.html.slim#L6) - do we want to do that here as well?

cosmetics-web/app/views/responsible_persons/notifications/edit.html.slim
```diff
@@ -0,0 +1,98 @@
...
+  .govuk-grid-column-full
+    table.govuk-table
```
broder: Should we re-use the `_responsible_person_details` partial for this?

cosmetics-web/config/routes.rb
```diff
@@ -4,40 +4,42 @@
 
   root 'landing_page#index'
 
-  resources :notifications, only: %i[new edit] do
-    member do
-      get :confirmation
-    end
-
+  resources :notifications, param: :reference_number, only: %i[index show] do
```
broder: I wonder if we should explicitly put this into a `poison_centres` `namespace` https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing. It might make it more obvious when we're using poison centre routes in RP journeys and vice versa?

cosmetics-web/config/routes.rb
```diff
@@ -4,40 +4,42 @@
...
     resources :build, controller: :notification_build, only: %i[show update new]
```
broder: And then maybe just nest this under RPs. It'll make the URL quite complicated but hopefully will keep more separation.

shared-web/app/models/shared/web/user.rb
```diff
@@ -36,6 +36,10 @@ def self.current=(user)
         RequestStore.store[:current_user] = user
       end
 
+      def has_role?(role)
```
broder: @jasiekmiko has also made some changes to `has_role?` but I don't think there will be too much conflict: https://github.com/UKGovernmentBEIS/beis-mspsds/pull/671/files#diff-eb638bc6d1bfc552d7c5966134f91f8cR130

jasiekmiko: The change also includes adding a method like this one, but with a matching signature, so that should be the one to come out of the merge: https://github.com/UKGovernmentBEIS/beis-mspsds/pull/671/files#diff-7ff7566e7b2458de1d390f3b49f9c7f1R43

cosmetics-web/app/views/application/_unfinished_product.html.slim
tom-bell-softwire: No, it's being added as part of COSBETA-26.

cosmetics-web/app/views/notifications/_product_details.html.slim
tom-bell-softwire: Yep! I think I added this to make the syntax highlighting nicer in RubyMine, but that's not a reason to keep it!

cosmetics-web/app/policies/notification_policy.rb
tom-bell-softwire: Is Pundit clever enough to know which one to use? If so, that would be much neater.

cosmetics-web/app/views/responsible_persons/notifications/edit.html.slim
tom-bell-softwire: I'll update to use that instead.

cosmetics-web/app/policies/notification_policy.rb
broder: Possibly not but you can manually use `policy_class`

cosmetics-web/config/routes.rb
tom-bell-softwire: I was trying to avoid too many deeply nested resources, but perhaps it's cleaner to keep them separated from the Poison Centre stuff.

shared-web/app/models/shared/web/user.rb
tom-bell-softwire: 👍

cosmetics-web/app/views/notifications/show.html.slim
```diff
@@ -0,0 +1,74 @@
...
+            | Shades
+          td.govuk-table__cell
+            = @notification.components.first&.shades&.join(", ") || @notification.shades || "N/A"
+        tr.govuk-table__row
+          th.govuk-table__header
+            | Label image
+          td.govuk-table__cell
+            - @notification.image_uploads.each do |image|
+              - if image.file_exists? && image.marked_as_safe?
+                = link_to image.filename, url_for(image.file), target: "_blank", rel: "noopener"
```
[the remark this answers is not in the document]
tom-bell-softwire: Good spot! Thanks.

cosmetics-web/spec/controllers/responsible_persons/notifications_controller_spec.rb
```diff
@@ -29,6 +29,15 @@
       expect(assigns(:pending_notification_files_count)).to eq(1)
     end
 
+    it "excludes pending notification files for other users" do
```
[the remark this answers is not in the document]
tom-bell-softwire: This is just the count of files that are still being processed. I think Raam discussed this with Ralph at the time. Once the files have been processed, any user under the same RP should be able to see them in the Errors/Unfinished tabs.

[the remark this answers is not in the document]
DWRendell: Oh right, that's fine then, nevermind :D

cosmetics-web/spec/controllers/notification_build_controller_spec.rb
```diff
@@ -4,8 +4,15 @@
   let(:responsible_person) { create(:responsible_person) }
   let(:notification) { create(:notification, responsible_person: responsible_person) }
 
-  let(:file) { fixture_file_upload('/testImage.png', 'image/png') }
-  let(:text_file) { fixture_file_upload('/testText.txt', 'application/text') }
+  let(:image_file) { fixture_file_upload("testImage.png", "image/png") }
+  let(:text_file) { fixture_file_upload("testText.txt", "application/text") }
+
+  let(:params) {
```
broder: Nice!

cosmetics-web/config/routes.rb
tom-bell-softwire: Yeah, great idea.

PR #680, opened 2019-03-04
MSPSDS-918: Clear autocomplete

kicferk1:
Make button focus like other elements, update hint text.

## Checklist:
### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/assets/application/stylesheets/form.scss
```diff
@@ -3,7 +3,7 @@
 }
 
 .autocomplete-select-with-clear {
-  width: 90%;
+  width: calc(100% - 48px);
```
jasiekmiko: These values are linked to those in `autocomplete.scss`. Can we either move this there, of it that doesn't make sense can we make this link explicit with a couple of comments

PR #681, opened 2019-03-04
COSBETA-31: Bulk file upload

raamSoftwire:
Link to ticket: [COSBETA-31](https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-31)

## Description

- Add functionality to upload multiple files
- Add notification files to queue as before
- Add error handling if too many files are uploaded at once

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


raamSoftwire:
Still to do
- add tests for multiple file upload
- add JS to fail early if they select too many files at once

raamSoftwire:
I'm happy to raise a separate ticket for the JS work, so I think this is ready to go.

cosmetics-web/app/models/notification_file.rb
```diff
@@ -20,6 +20,7 @@ class NotificationFile < ApplicationRecord
 
   @allowed_content_types = %w[application/zip application/x-zip-compressed].freeze
   @max_file_size_bytes = 30.megabytes
+  @no_of_files_limit = 100
```
tom-bell-softwire: Might be clearer to call this `max_number_of_files` (same for the method below).

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -9,20 +8,30 @@ def show
   end
 
   def create
-    unless notification_file_params && notification_file_params[:uploaded_file]
-      @notification_file.errors.add :uploaded_file, "No file selected"
+    @errors = []
+
+    unless uploaded_files_params
+      @errors << { text: "No files selected", href: "#" }
```
tom-bell-softwire: It would be good to use the anchor link to jump to the file input field in the view, which is how the Design System error summary behaves. Same for the other error message added below.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -9,20 +8,30 @@ def show
...
+    if uploaded_files_params.length > NotificationFile.get_no_of_files_limit
+      @errors << { text: "Too many files selected. Please select no more than #{NotificationFile.get_no_of_files_limit} files", href: "#" }
       return render :new
     end
 
-    @notification_file.name = notification_file_params[:uploaded_file].original_filename
-    @notification_file.responsible_person = @responsible_person
-    @notification_file.user = User.current
+    uploaded_files_params.each do |uploaded_file|
+      notification_file = NotificationFile.new
```
tom-bell-softwire: Can you set the name, responsible person and user as part of this line? Something like:
```
notification_file = NotificationFile.new(name: uploaded_file.original_filename, responsible_person: @responsible_person, user: User.current)
```

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -9,20 +8,30 @@ def show
...
+      notification_file.name = uploaded_file.original_filename
+      notification_file.responsible_person = @responsible_person
+      notification_file.user = User.current
 
-    if @notification_file.save
-      redirect_to responsible_person_notifications_path(@responsible_person)
-    else
-      render :new
+      unless notification_file.save
+        render :new
```
tom-bell-softwire: What happens in this scenario? Is the user given an error message?

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -38,19 +47,15 @@ def destroy_all
...
     @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
     authorize @responsible_person, :show?
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
-  def notification_file_params
-    if params.has_key?(:notification_file)
-      params.require(:notification_file).permit(:name, :uploaded_file)
+  def uploaded_files_params
+    if params.has_key?(:uploaded_files)
```
tom-bell-softwire: Can you just use `params.permit(:uploaded_files)` instead of this if-block?

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.describe ResponsiblePersons::NotificationFilesController, type: :controller do
   let(:responsible_person) { create(:responsible_person) }
   let(:valid_attributes) {
-    { uploaded_file: fixture_file_upload("/testExportFile.zip") }
+    [fixture_file_upload("/testExportFile.zip")]
```
tom-bell-softwire: Is the "/" needed in the filename?

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
```diff
@@ -20,7 +22,7 @@
         .govuk-form-group
```
tom-bell-softwire: I think we should add the "govuk-form-group--error" class here if there are any errors.

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
```diff
@@ -20,7 +22,7 @@
...
           .form-group
```
tom-bell-softwire: Also, I don't think this additional "form-group" element is needed.

cosmetics-web/app/models/notification_file.rb
raamSoftwire: Naming things is hard. I spent ages thinking about that one, and still hated it. Yours is much better.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: Done. I've used an ID on the form group element `#file-upload-form-group`. Is this the best way to do this?

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: Done, thanks

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
tom-bell-softwire: Yep, sounds great!

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: So here there should only be three scenarios
- no files selected (handled above by adding to `@errors`)
- too many files (handled above by adding to `@errors`)
- everything else (handled on the following page in the errors tab)

If the notification file saves, then we go ahead to the next page. If not then we must be in one of the first two cases, and so if we reload the page (by rendering `:new`) then `@errors` will be shown

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: That didn't seem to work, but `params.require(:uploaded_files)` did.

cosmetics-web/spec/controllers/responsible_persons/notification_files_controller_spec.rb
raamSoftwire: I guess not, good spot.

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
raamSoftwire: I agree, thanks

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
tom-bell-softwire: If the `save` call fails, it's likely because there is some other error we haven't anticipated in the above scenarios. In this case, is it worth adding those errors to the `@errors` array so that they can be displayed on the page? Something like:
```
@errors.concat(notification_file.errors.full_messages.map { |message| { text: message, href: "#" } })
```
might work.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
tom-bell-softwire: Will that throw an exception if there are no uploaded files?

raamSoftwire: Actually, that doesn't seem happy in the case where the params are empty. I think we need to keep the if block for this initially check

cosmetics-web/app/views/responsible_persons/notification_files/new.html.slim
raamSoftwire: Done

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
raamSoftwire: Sure thing, thanks

PR #690, opened 2019-03-05
MSPSDS-1044 Radio buttons default to unchecked in TS flow

samlionheart:
## Description
- Clear repeat_step and session step every time you go to show so that it isn't selected by default
- Renamed file_type to further_page_type to be clearer about what the variable contains
- Changed repeatable_info and corrective action to call @investigation errors like in the controller
- Changed govuk_radios so that it's turning symbol into string to avoid errors around name
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -53,6 +53,8 @@ def show
     when :corrective_action, *other_information_types
       return redirect_to next_wizard_path unless @repeat_step == "Yes"
     end
+    @repeat_step = nil
+    session.delete further_key(step)
```
MatthewBurstein: This feels less consistent to me than using a different `further_key` for the `:has_corrective_action` step. Happy to talk it over if you like :)

jasiekmiko: > Clear repeat_step and session step every time you go to show so that it isn't selected by default

I'm a little worried that deleting it might disrupt the flow in unexpected ways. What happens if you refresh the page without clicking "next" - I'd expect that this change would cause it to skip our step!

Perhaps we need to sets of variables: steps to be completed (including multiples of the same step) and the ones that have been done, then display the current step based on the first difference in those lists?

shared-web/app/views/components/_govuk_radios.html.slim
```diff
@@ -2,7 +2,7 @@
 / https://github.com/alphagov/govuk-frontend/
 /       blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/radios/template.njk
 ruby:
-  id_prefix = local_assigns[:idPrefix].presence || name
+  id_prefix = local_assigns[:idPrefix].presence || name.to_s
```
jasiekmiko: neat catch! I think we might want to apply the same to the LHS of the `||`?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
jasiekmiko: Good fix for that @samlionheart, thanks!

PR #692, opened 2019-03-05
MSPSDS-575 alert trading standards

MatthewBurstein:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


mspsds-web/Gemfile
```diff
@@ -13,6 +13,9 @@ gem "cf-app-utils"
 gem "elasticsearch-model"
 gem "elasticsearch-rails"
 gem "jbuilder", "~> 2.8"
+gem "lograge"
```
broder: I think `lograge` and `mini_magick` are being pulled in from `shared-web`

mspsds-web/Gemfile
```diff
@@ -13,6 +13,9 @@ gem "cf-app-utils"
...
+gem "mini_magick"
+gem "notifications-ruby-client", ">= 3.1.0"
```
broder: Do we need this as well as `govuk_notify_rails` in `shared-web`?

mspsds-web/app/controllers/investigations/alerts_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+  end
+
+private
+
+  def clear_session
+    session.delete(:alert)
+  end
+
+  def alert_valid?
+    @alert.valid?
```
broder: I don't think we need both this and `@alert.description.blank?` / `@alert.summary.blank?` below

mspsds-web/app/controllers/investigations/alerts_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+  def set_alert
+    @alert = Alert.new alert_params.merge(
+      investigation_id: @investigation.id,
+      source: UserSource.new(user: User.current)
+    )
+  end
+
+  def set_default_field_values
+    @default_summary = "Product safety alert: "
+    @default_description = "\r\n\r\n\r\nMore details can be found on the case page: #{investigation_url @investigation}"
```
broder: Do we need `\r\n` rather than just `\n`?

mspsds-web/app/controllers/investigations/alerts_controller.rb
```diff
@@ -0,0 +1,105 @@
...
+    params.require(:alert).permit(:summary, :description)
+  end
+
+  def set_user_count
+    @user_count = number_with_delimiter(User.all.length, delimiter: ',')
+  end
+
+  def get_preview
+    @preview = NotificationsClient.instance.generate_template_preview(
+      '47fb7df9-2370-4307-9f86-69455597cdc1',
```
broder: It would be good to share this with the value in `NotifyMailer` so we only have to change it in one place.

mspsds-web/app/jobs/send_alert_job.rb
```diff
@@ -0,0 +1,12 @@
+class SendAlertJob < ApplicationJob
+  def self.perform(email_subject, email_body)
+    User.all.each do |user|
```
broder: Does the worker have access to the users? I don't think the keycloak details are shared with the worker...

mspsds-web/db/migrate/20190222165902_add_alert_id_field_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddAlertIdFieldToActivities < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :activities, :alert, index: false, foreign_key: true
```
broder: Do we need to link the alert to an activity? The current relations that we have are to generate e.g. "View product" links or to check if the current user can view the correspondence.

mspsds-web/app/models/alert.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  validates :summary, presence: true
+  validates :description, presence: true
+
+  after_save :create_audit_activity
+
+  after_save :send_alert_email
+
+  def send_alert_email
+    SendAlertJob.perform summary, description
```
broder: `perform` makes this run on the web server - was that intentional?

mspsds-web/app/models/audit_activity/alert/add.rb
```diff
@@ -0,0 +1,25 @@
+class AuditActivity::Alert::Add < AuditActivity::Base
+  extend ActionView::Helpers::NumberHelper
+  belongs_to :alert
+  belongs_to :investigation
+
+  def self.from(alert)
+    self.create(
+      title: "Product safety alert sent",
+      body: build_body(alert),
+      source: UserSource.new(user: current_user),
```
broder: Should this be `User.current` rather the `current_user`?

mspsds-web/app/models/audit_activity/alert/add.rb
```diff
@@ -0,0 +1,25 @@
...
+      investigation: alert.investigation,
+      alert: alert
+    )
+  end
+
+  def self.build_body(alert)
+    [
+        "From: **DUMMY EMAIL ADDRESS**",
```
broder: Is it worth displaying this?

mspsds-web/app/models/audit_activity/alert/add.rb
```diff
@@ -0,0 +1,25 @@
...
+        "To: **All users** (#{number_with_delimiter(User.all.length, delimiter: ',')} people)",
+        "Subject: **#{alert.summary}**",
+        "Date sent: **#{alert.created_at.strftime('%d/%m/%Y')}**",
+        "<br>" + alert.description
```
broder: Could we use an empty string rather than adding another `<br> here? i.e.
```
...
"Date sent: **#{alert.created_at.strftime('%d/%m/%Y')}**",
"",
alert.description
].join("<br>"
```

mspsds-web/app/assets/application/stylesheets/timeline.scss
```diff
@@ -6,7 +6,7 @@ $timeline-item-margin-bottom: 20px;
   padding: 10px;
   margin: 0 auto $timeline-item-margin-bottom;
 
-  ul {
+  .timeline-list {
```
broder: We could also do this using `> ul` and `> li` below

mspsds-web/app/views/investigations/alerts/about_alerts.html.slim
```diff
@@ -0,0 +1,42 @@
+- content_for :page_title, "Alert trading standards"
+- content_for :after_header do
+  .govuk-grid-row
+    .govuk-grid-column-one-third
```
broder: I don't think we wrap "Back to ..." links in in rows / columns anywhere else - was there a specific reason for using it here?

mspsds-web/app/views/investigations/alerts/about_alerts.html.slim
```diff
@@ -0,0 +1,42 @@
...
+      = link_to "Back to #{@investigation.pretty_description}", @investigation, class: "govuk-back-link"
+
+- if @investigation.is_private
```
broder: Is there anything that stops me from going directly to the `compose` page on a private case?

mspsds-web/app/views/investigations/alerts/about_alerts.html.slim
```diff
@@ -0,0 +1,42 @@
...
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      h1.govuk-heading-xl
+        p.govuk-caption-xl
+          = @investigation.pretty_description
+        | You cannot send an alert about a restricted case
+      p.govuk-body
```
broder: I don't think we need to put `.govuk-body` on `p` tags because of https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/shared-web/app/assets/application/stylesheets/main.scss#L1

mspsds-web/app/views/investigations/alerts/preview.html.slim
```diff
@@ -0,0 +1,47 @@
...
+            tr.email-message-meta.govuk-table__row
+              th.govuk-table__header Subject
+              td.govuk-table__cell
+                = @alert.summary
+        .email-message-body
+          == @preview.html
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      p.govuk-body
+        | If you’re not happy you can
```
broder: Does `' If you’re not happy you can` do the same think as having `'` on the next line?

mspsds-web/app/views/investigations/alerts/preview.html.slim
```diff
@@ -0,0 +1,47 @@
...
+        '
+        = link_to "edit your message", previous_wizard_path, class: "govuk-link"
+        | .
+      p.govuk-body
+        | Alerts may take up to 5 minutes to send.
+
+  // TODO calculate number of people
```
broder: Is this TODO still needed?

mspsds-web/db/schema.rb
```diff
@@ -149,7 +159,9 @@
     t.boolean "is_closed", default: false
     t.boolean "is_private", default: false, null: false
     t.text "non_compliant_reason"
+    t.text "planned_corrective_action_description"
```
broder: This change is only on the research branch for now, right?

mspsds-web/test/system/alert_test.rb
```diff
@@ -0,0 +1,71 @@
...
+    assert_text "Alert sent XXXX"
+  end
+
+  test "requires a restricted case be derestricted before raising an alert" do
+    @private_investigaton = investigations :private
+    fill_in_activity_selection
+    assert_selector :h1, text: "You cannot send an alert about a restricted case"
+
+    click_on "Change case visibility"
+    assert_selector :h1, text: "Legal privelege"
```
broder: Should it be "privilege"?

shared-web/app/views/components/_govuk_textarea.html.slim
```diff
@@ -34,4 +34,7 @@ ruby:
                 maxlength=local_assigns[:max_length]
                 aria-describedby=described_by
                 autocomplete=local_assigns[:autocomplete]]
+        / If the first new line of the textarea is empty, HTML swallows it, so here we add it back in. TODO check this
```
broder: What does this fix?

mspsds-web/Gemfile
MatthewBurstein: It depends what you are using. The version pulled in by govuk_notify_rails does not include the functionality to request an html preview of the email template. I think all other functionality is included though.

mspsds-web/app/controllers/investigations/alerts_controller.rb
MatthewBurstein: 👍

mspsds-web/app/controllers/investigations/alerts_controller.rb
MatthewBurstein: 👍

mspsds-web/Gemfile
broder: https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/Gemfile.lock#L232 suggests that we are actually pulling in version 3.1.0

mspsds-web/app/views/investigations/alerts/preview.html.slim
MatthewBurstein: Ooh very nice. Yes it does.

mspsds-web/Gemfile
MatthewBurstein: It does, doesn't it. But then I don't understand why the .html method was not available on the response. Has that been updated since this branch diverged from master?

broder: Yeah, potentially!

shared-web/app/views/components/_govuk_textarea.html.slim
MatthewBurstein: If this isn't in place, then when the content is added to the text area, it loses a new line from the front of the string. I noticed it because if you submit and fail the validation on the other field, when the page renders again the line has gone. It's because `<textarea>` elements don't have a value attribute so this ends up as `<textarea>\r\nsome content</textarea>` and this handles carriage returns weirdly.

mspsds-web/app/controllers/investigations/alerts_controller.rb
MatthewBurstein: At first I was just using \n, but \r\n is what the html needs to display correctly. I think it might be connected with the html it is converted in to as per [this comment](https://github.com/UKGovernmentBEIS/beis-mspsds/pull/692#discussion_r262573138)

mspsds-web/db/schema.rb
MatthewBurstein: Yes it is good spot.

mspsds-web/Gemfile
MatthewBurstein: Have removed from this gemfile

mspsds-web/app/assets/application/stylesheets/timeline.scss
MatthewBurstein: This doesn't work because if the user has used markdown to style the email, the email preview will contain `ul` and `li` elements, which will then be given this styling in the activity body.

MatthewBurstein: Actually scratch that sorry, I misunderstood how `>`  worked.

mspsds-web/app/models/audit_activity/alert/add.rb
MatthewBurstein: I checked with Ed and he said to keep it in.

PR #693, opened 2019-03-05
MSPSDS-1062 Remove pdf case export

samlionheart:
## Description

Removed the link to pdf export on case page as well as the pdf template. In the future we may still want to use pdf export so the pdf gem can remain.
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/views/investigations/ts_investigations/has_corrective_action.html.slim
```diff
@@ -8,7 +8,7 @@
         h1.govuk-heading-xl
           span.govuk-caption-xl Report an unsafe or non-compliant product
           | Has any corrective action been agreed or taken?
-      = render "form_components/govuk_radios", form: form, key: :further_corrective_action,
+      = render "form_components/govuk_radios", form: form, key: :further_has_corrective_action,
```
broder: Is this change related to this ticket?

docker-compose.yml
```diff
@@ -102,6 +102,7 @@ services:
       - mspsds-worker
     env_file:
       - .env.keycloak.mspsds
+    command: bash -c "rm -f tmp/pids/server.pid && bundle exec rails s -p 3000 -b '0.0.0.0'"
```
broder: Should we add this to `docker-entrypoint.sh` for cosmetics and mspsds?

samlionheart: This was accidentally popped in this branch, was something to do with docker-sync-stack but ended up being fixed by doing docker-sync-stack clean, as the server.pid was hidden very well somewhere

samlionheart: Although this would be helpful to have as it's good for when things don't close down properly

PR #696, opened 2019-03-05
COSBETA-33: Elasticsearch for poison centres

kicferk1:
This branch is rebased on top of COSBETA-32, so probably best if that gets merged in first. 

It adds elasticsearch indexing to notifications, as well as a way to search and see results on notifications index page. It works locally(after down and build). It ought to work after we deploy it once we add 'cosmetics-elasticsearch' service to relevant environment, mirroring mspsds-elasticsearch one. 

It probably warrants readme update, changelog update, as well as some tests(and fix old ones), but the main bulk of logic should be done. 

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


kicferk1:
Do we store product category anywhere? The ticket says we should search by it, but I can't find it anywhere. 

If it gets merged in after this ticket, updating ES to index it as well should be a simple edit of `as_indexed_json` on Notification model. 

Also, before it gets merged in I should probably create cosmetics-elasticsearch on whichever environment we deploy to.

cosmetics-web/Gemfile
```diff
@@ -3,9 +3,15 @@ source "https://rubygems.org"
 
 ruby File.read(".ruby-version").strip
 
+# Will Paginate must be installed before elasticsearch
+# https://github.com/elastic/elasticsearch-rails/issues/239
+gem 'will_paginate', '~> 3.1.0'
```
broder: Could we put these in `shared-web.gemspec`?

cosmetics-web/app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,76 @@
+# rubocop:disable Metrics/BlockLength
+module Searchable
```
broder: Should we put this in `shared-web`?

cosmetics-web/app/helpers/query_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+      multi_match: {
+        query: @query,
+        fuzziness: "AUTO"
+      }
+    }
+  end
+
+  def filter_params
+    {
+      term: { "state": "notification_complete" }
```
broder: Should we be indexing non-complete notifications?

cosmetics-web/spec/controllers/poison_centres/notifications_controller_spec.rb
```diff
@@ -40,6 +44,25 @@
       end
     end
 
+    describe "search on #index" do
+      before do
+        create(:registered_notification, responsible_person: responsible_person_1, product_name: "bbbb")
```
broder: Could we do this with `let`s? Then we might not need to do `Notification.find_by` in the tests below

cosmetics-web/app/helpers/query_helper.rb
kicferk1: We can prevent actual indexing by checking state in `as_indexed_json`, but I'm not sure there is a way of making ES return correct things without filtering. 

From what I run locally it seems ES will index every last record, so when returning all records(no search value), we will have to filter out non-complete notifications anyway.

broder: I think we can use scopes so that we never `import` the non-complete notifications e.g. https://github.com/elastic/elasticsearch-rails/blob/6069226bf25d7be91bc088e90d59bc37d496e9bf/elasticsearch-model/spec/elasticsearch/model/adapters/active_record/import_spec.rb#L88 & https://github.com/elastic/elasticsearch-rails/blob/6069226bf25d7be91bc088e90d59bc37d496e9bf/elasticsearch-model/spec/support/app/import_article.rb#L21

cosmetics-web/app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,14 @@
...
+  extend ActiveSupport::Concern
+  included do
+    include Shared::Web::Concerns::Searchable
+
+    def self.full_search(query)
+      # This line makes sure elasticsearch index is recreated before we search
+      # It fixes the issue of getting no results the first time case list page is loaded
```
tom-bell-softwire: Could this also go into `Shared::Web` or does it require something app-specific?

Also, "case list page" -> "product list page"

cosmetics-web/app/controllers/poison_centres/notifications_controller.rb
```diff
@@ -1,6 +1,9 @@
 class PoisonCentres::NotificationsController < ApplicationController
+  include QueryHelper
   def index
-    @notifications = get_registered_notifications(10)
+    @query = query_params[:q]
+    result = Notification.full_search(build_query).paginate(page: params[:page], per_page: 10)
```
tom-bell-softwire: Worth replacing the existing `get_registered_notifications` method with an equivalent `search_registered_notifications` method that does this?

cosmetics-web/app/models/notification.rb
```diff
@@ -28,6 +33,23 @@ def self.duplicate_notification_message
...
+  end
+
+  def as_indexed_json(*)
+    as_json(
+      only: %i[product_name state],
+      include: {
+        responsible_person: {
+          only: %i[name]
+        }
+      }
```
tom-bell-softwire: The `root_category`, `sub_category` and `sub_sub_category` values on the associated `components` should also be indexed. These are enum values, but for now we should probably index their display name to allow text search.

cosmetics-web/app/models/notification.rb
```diff
@@ -1,4 +1,7 @@
 class Notification < ApplicationRecord
+  include Searchable
+
+  index_name [Rails.env, "notifications"].join("_")
```
tom-bell-softwire: Might be good to move this down next to the `scope` declaration below.

cosmetics-web/app/views/poison_centres/notifications/index.html.slim
```diff
@@ -3,6 +3,15 @@
   .govuk-grid-column-two-thirds
     h1.govuk-heading-xl
       | Search cosmetic products
+
+.govuk-grid-row
+  .govuk-grid-column-one-quarter class="govuk-!-margin-bottom-2"
+    = form_with(method: :get, class: "mspsds-search") do |form|
```
tom-bell-softwire: Does the class need to be "mspsds-search" ?

cosmetics-web/config/environments/production.rb
```diff
@@ -78,4 +79,19 @@
   # Store uploaded files on the local file system on test, and in S3 on production.
   # (see config/storage.yml for options)
   config.active_storage.service = :amazon
+
+  config.after_initialize do
```
tom-bell-softwire: Worth adding a brief comment to explain why we're doing this?

cosmetics-web/spec/controllers/poison_centres/notifications_controller_spec.rb
```diff
@@ -18,20 +22,24 @@
 
     describe "GET #index" do
       it "gets all registered notifications" do
+        notifications = rp_1_notifications + rp_2_notifications
+        Notification.elasticsearch.import force: true
```
tom-bell-softwire: Can this be moved to a `before` block at the top of this `describe` block, to avoid duplicating it in each case below?

cosmetics-web/spec/controllers/poison_centres/notifications_controller_spec.rb
```diff
@@ -40,6 +48,25 @@
       end
     end
 
+    describe "search on #index" do
+      before do
+        distinct_notification
```
tom-bell-softwire: I think you can move the `let` declarations for these three notifications to the top of this `describe` block, and if you use `let!` it should force them to be created immediately (rather than lazily), so that you don't need to invoke them in this `before` block.

cosmetics-web/spec/system/poison_centre_user_spec.rb
```diff
@@ -21,24 +21,26 @@
   end
 
   it "can view a list of all registered notifications" do
+    notifications = rp_1_notifications + rp_2_notifications
+    Notification.elasticsearch.import force: true
```
tom-bell-softwire: Again, worth DRYing out these calls with a `before` block?

cosmetics-worker/manifest.yml
```diff
@@ -13,4 +13,5 @@ applications:
   services:
     - cosmetics-database
     - cosmetics-queue
+    - cosmetics-elasticsearch
```
tom-bell-softwire: Does the Worker also need to use the Elasticsearch service?

shared-web/lib/shared/web.rb
```diff
@@ -1,8 +1,11 @@
 # requires all dependencies
 Gem.loaded_specs['shared-web'].dependencies.each do |d|
-  require d.name
+  require d.name unless d.name.include? "elasticsearch"
 end
 
+require "elasticsearch/model"
```
tom-bell-softwire: Do you know why we need to require these separately? If so, might be worth adding a comment to explain.

cosmetics-web/app/models/concerns/searchable.rb
kicferk1: Full search in mspsds builds the query with a bunch of options like highlighted fields and exact fields that we don't need. So if I push this method into shared web I will need to build query before that in mspsds, using methods that belong in mspsds searchable :/

tom-bell-softwire: Ah, fine to leave this as app-specific, then.

shared-web/lib/shared/web.rb
kicferk1: I don't know the underlying issue that causes it(maybe a conscious choice), but the gem name is "elasticsearch-model", whereas when you require it you need to use "elasticsearch/model", so we need to separate it from loading other gems which don't have that issue.

tom-bell-softwire: Fair enough! That sounds like a sufficient explanation of the issue to me.

cosmetics-worker/manifest.yml
kicferk1: David says it is very likely going to crash if it doesn't have access to ES, due to re-using Web's code

PR #709, opened 2019-03-06
MSPSDS-938 Remove case count from business and product headings

samlionheart:
Case count is no longer needed as products and businesses should only be linked to one case
Ticket: [MSPSDS-1051](https://regulatorydelivery.atlassian.net/browse/MSPSDS-1051)
## Description
Removing the elements from each page. 
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


samlionheart:
I don't know when and when not to do all the checks, but I just wanted to go through every check on this ticket regardless of relevance to see how to do screen reading and disabling css.

PR #724, opened 2019-03-08
MSPSDS-1006: Text input component

kicferk1:
Add a govuk style text input component. 

## Checklist:
### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


shared-web/app/views/components/_govuk_textinput.html.slim
```diff
@@ -0,0 +1,35 @@
...
+  has_hint = local_assigns[:hint].present?
+  has_hint = local_assigns[:hint][:text].present? || local_assigns[:hint][:html].present? if has_hint
+  has_error = local_assigns[:errorMessage].present?
+
+  form_group_classes = [local_assigns.dig(:formGroup, :classes)]
+  form_group_classes.push "govuk-form-group--error" if local_assigns[:errorMessage]
+  input_classes = ["govuk-input", local_assigns[:classes]]
+  input_classes.push("govuk-input--error") if local_assigns[:errorMessage].present?
+
+.govuk-form-group class=form_group_classes id=local_assigns[:group_id]
```
jasiekmiko: `group_id` is not part of the component interface, I don't think we want to add it here. If there are places this is trying to cater for then we should find other ways to resolve them, e.g. with another div wrapper where needed.

shared-web/app/views/components/_govuk_textinput.html.slim
```diff
@@ -0,0 +1,35 @@
+/ Implementation based on
+/ https://github.com/alphagov/govuk-frontend/
+/ blob/943ff14752f0a8a765ee3f90bc3e1ecd9205e36c/src/components/input/template.njk
+ruby:
+  described_by = ""
+  has_hint = local_assigns[:hint].present?
+  has_hint = local_assigns[:hint][:text].present? || local_assigns[:hint][:html].present? if has_hint
```
jasiekmiko: Not sure we need this. If `hint` is provided, then the interface requires either one of `text` or `html` as its children. Additionally it'll be up to the `govuk_hint` component to decide that. I'd suggest we just remove the `has_hint` variable and go for `- if local_assigns[:hint]` as the guard later down

shared-web/app/views/components/_govuk_textinput.html.slim
```diff
@@ -0,0 +1,35 @@
...
+    - described_by = described_by.present? ? described_by + ' ' + error_id : error_id
+    = render "components/govuk_error_message", id: error_id, **errorMessage
+
+  input[*(local_assigns[:attributes] || {})
+                class=input_classes
+                id=id
+                name=name
+                type=(local_assigns[:type] || 'text')
+                value=local_assigns[:value]
+                aria-describedby=described_by
```
jasiekmiko: We probably want to use `described_by.presence` so we don't have a bunch of empty `aria-descirbedby` attributes

mspsds-web/app/views/application/_date_input.html.slim
```diff
@@ -6,29 +6,19 @@
...
       / https://guides.rubyonrails.org/form_helpers.html#using-date-and-time-form-helpers-model-object-helpers
-      .govuk-form-group
-        = form.label :day, "Day", class: "govuk-label govuk-date-input__label"
-        = form.text_field :day,
-                class: "govuk-input govuk-date-input__input govuk-input--width-2" \
-                            + (form.object.errors[:day].present? ? " govuk-input--error" : ""),
-                pattern: "[0-9]*",
-                type: "number",
-                value: form.object.day
+      - error_class_day = form.object.errors[:day].present? ? " govuk-input--error" : ""
```
jasiekmiko: Why do we need to specify this class? It should be added by the form glue?
Same for the month and year.

If it's something relatively sticky don't worry about it much, since this whole file ought to become a component itself at some point, but would be good to keep the repetition away if possible.

mspsds-web/app/views/businesses/_form.html.slim
```diff
@@ -1,54 +1,43 @@
+- label_class = "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
 fieldset.govuk-fieldset
-  .govuk-form-group class=[("govuk-form-group--error" if business.errors[:trading_name].any?)]
-    = form.label :trading_name, "Trading name",
-            class: "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
-    span.govuk-hint#business-trading-name-hint
-      |  The name the business is known by
-    = form.text_field :trading_name, class: "govuk-input", aria_describedby: "business-trading-name-hint"
+  - label_html = capture do
```
jasiekmiko: This feels overkill to use html here - would `text` and `classes` params on `label` not be enough?

Same for all fields below?

shared-web/app/views/components/_govuk_textinput.html.slim
```diff
@@ -0,0 +1,35 @@
+/ Implementation based on
```
jasiekmiko: On second thought - can we move the file to `_govuk_input.html.slim`? It'll match the nunjucks macro better and potentially make it easier to integrate their tests to our implementations of the components in the future.
It's also not a text-only input, it just defaults to that type!

shared-web/app/views/form_components/_govuk_textinput.html.slim
```diff
@@ -0,0 +1,14 @@
+/ This partial acts as glue between rails form helpers and the component textinput
```
jasiekmiko: ditto filename

mspsds-web/app/views/investigations/creation_flow/complainant_details.html.slim
```diff
@@ -11,17 +11,16 @@
...
-        .govuk-form-group class=("govuk-form-group--error" if @complainant.errors[:name].any?)
-          = form.label :name, class: "govuk-label"
-          = form.text_field :name, class: "govuk-input govuk-!-width-two-thirds", maxlength: 1000
-        .govuk-form-group class=("govuk-form-group--error" if @complainant.errors[:phone_number].any?)
-          = form.label :phone_number, class: "govuk-label"
-          = form.text_field :phone_number, class: "govuk-input govuk-!-width-two-thirds", maxlength: 1000
-        .govuk-form-group class=("govuk-form-group--error" if @complainant.errors[:email_address].any?)
-          = form.label :email_address, class: "govuk-label"
-          = form.text_field :email_address, class: "govuk-input govuk-!-width-two-thirds", maxlength: 1000
+        = render "form_components/govuk_textinput", key: :name, form: form, max_length: 1000,
```
jasiekmiko: `max_length` is not a property on this component. You want to use something like `attributes: { maxlength: 1000 }` here (and in the instances below)

mspsds-web/app/views/investigations/project/new.html.slim
```diff
@@ -10,11 +10,9 @@
       fieldset.govuk-fieldset
         legend.govuk-fieldset__legend.govuk-fieldset__legend--l
           | Please provide a title
-        .govuk-form-group class=("govuk-form-group--error" if @investigation.errors[:user_title].any?)
-          span.govuk-hint#enquiry-title-hint
-            | This will help other users identify this project
-          = form.text_field :user_title, class: "govuk-input", "aria-describedby": "project-title-hint",
-                  maxlength: 100
+        = render "form_components/govuk_textinput", key: :user_title, form: form, max_length: 100,
```
jasiekmiko: ditto maxlength

mspsds-web/app/views/investigations/tests/_form.html.slim
```diff
@@ -13,8 +13,9 @@ fieldset.govuk-fieldset
           = link_to "Add a product", new_investigation_product_path(investigation)
           |  to send it for testing.
       - elsif @investigation.products.size == 1
-        = form.text_field :product_name, value: @investigation.products.first.name,
-                disabled: true, class: "govuk-input govuk-!-width-two-thirds"
+        = render "form_components/govuk_textinput", key: :product_name, form: form, disabled: true,
```
jasiekmiko: Same problem with `disabled` here, you'll need to pass it through the `attributes` property

mspsds-web/app/views/locations/_form.html.slim
```diff
@@ -1,36 +1,36 @@
+- label_class = "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
 fieldset.govuk-fieldset
-  .govuk-form-group class=[("govuk-form-group--error" if @location.errors[:name].any?)]
-    = form.label :name, "Name", class: "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
-    span.govuk-hint#location-name-hint
-      |  E.g. Main office, Warehouse
-    = form.text_field :name, class: "govuk-input search-term", aria_describedby: "location-name-number-hint"
+  - label_html = capture do
```
jasiekmiko: Again, not convinced by the need for `html` here?

mspsds-web/app/views/products/_form.html.slim
```diff
@@ -1,72 +1,71 @@
...
-  .govuk-form-group class=("govuk-form-group--error" if product.errors[:product_type].any?)
-    = form.label :product_type, "Product type",
-            class: "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
-    span#report-product-type-hint.govuk-hint
-      | In plain English, what is it?
-      br
-      | For example, ‘dishwasher, soft toy, chainsaw...’
-    = form.text_field :product_type, class: "govuk-input govuk-!-width-two-thirds",
-            aria_describedby: "report-product-type-hint"
+  - label_html = capture do
```
jasiekmiko: same `html` query about this file :)

shared-web/app/views/components/_govuk_textinput.html.slim
```diff
@@ -31,5 +31,8 @@ ruby:
                 name=name
                 type=(local_assigns[:type] || 'text')
                 value=local_assigns[:value]
+                pattern=local_assigns[:pattern]
+                disabled=local_assigns[:disabled]
+                maxlength=local_assigns[:max_length]
```
jasiekmiko: Neither `disabled` nor `maxlength` are part of the interface as detailed in https://design-system.service.gov.uk/components/text-input/#options-text-input-example, so we shouldn't be adding them here. They can be provided using the `attirbutes` param.
If you feel a strong need to, then I'm happy modifying the `glue` file to understand them and convert them into the right `attributes` format, but I think this file should match the source as close as possible

mspsds-web/test/system/create_ts_investigation_test.rb
```diff
@@ -161,8 +161,8 @@ def fill_in_business_form business
     fill_in "business_trading_name", with: business.trading_name
     fill_in "business_legal_name", with: business.legal_name
     fill_in "business_company_number", with: business.company_number + "unique company number"
-    fill_in "business_contacts_attributes_0_name", with: business.contacts.first.name
-    fill_in "business_locations_attributes_0_address_line_1", with: business.locations.first.address_line_1
+    fill_in "business[contacts_attributes][0]_name", with: business.contacts.first.name
```
jasiekmiko: Hmmm.. The need to change this file is probably pointing to the fact that the glue isn't following the rails convetion as well it should. Is this looking by `id`? If so we should probably make sure the `id` our glue sets on the field is consistent with rails.

mspsds-web/app/views/application/_date_input.html.slim
kicferk1: It's because error messages are empty(and we don't really want to display a message for each field anyway, because it breaks the layout), but the previous implementation provides a nice feature of only marking as red inputs that have a problem in them. 

So really, this might be a rare case of the view(date_input) being better off handling errors of its inputs directly rather than using default behaviour(which wants to display an error message).

mspsds-web/app/views/businesses/_form.html.slim
kicferk1: They are not. Due to implementation of label it seems `govuk-label` class is applied last. If I change it so that it is applied first, I can make the need for html go away, but should I?

jasiekmiko: wait what, class order shouldn't matter?
Ohhh... Looks like this form is just a mess... let's chat in person! :P

mspsds-web/app/views/application/_date_input.html.slim
jasiekmiko: Ah yeah, i was expecting them to be empty, but didn't realise they would still show the error element. Yes, definitely needs to be a little custom - as I said, it'll be up for this component to do things properly in anyhow :)

mspsds-web/app/views/businesses/_form.html.slim
jasiekmiko: as dicussed, let's just fix the classes to be proper and not use the html. feel free to remove the superflous fieldsets if you're feeling like cleaning up a bit more :D

shared-web/app/helpers/shared/web/components_helper.rb
```diff
@@ -6,7 +6,8 @@ def get_attribute_name(form, attribute)
       end
 
       def get_attribute_id_prefix(form, attribute)
-        form.object_name.present? ? "#{form.object_name}_#{attribute}" : attribute.to_s
+        initial_name = form.object_name.present? ? "#{form.object_name}_#{attribute}" : attribute.to_s
+        initial_name.gsub(/[\[\]_]+/,"_")
```
jasiekmiko: Any chance you've still got the source for this? would be good to put as a reference in a comment :)

jasiekmiko:
```suggestion
        initial_name.gsub(/[\[\]_]+/, "_")
```

PR #732, opened 2019-03-11
MSPSDS fix failing system tests

MatthewBurstein:
Fixes for a couple of the system tests which were failing for trivial reasons.


broder:
Are these fixes related to another specific change?

PR #733, opened 2019-03-11
COSBETA-57: Validate phone registration

kicferk1:
Fixes a bug where users could provide invalid phone number.

Changes a returned value of `isPhoneNumberValid` to false when it cannot be interpreted as an international or possible UK phone number. It used to be true, which allowed pretty much any badly invalid number to go through. 

## Checklist:

### DIFF
--- a/keycloak/providers/registration-form/src/main/java/uk/gov/beis/mspsds/keycloak/providers/RegistrationMobileNumber.java
+++ b/keycloak/providers/registration-form/src/main/java/uk/gov/beis/mspsds/keycloak/providers/RegistrationMobileNumber.java
@@ -61,7 +61,7 @@ public class RegistrationMobileNumber implements FormAction {
         } else if (isInternationalNumber(formattedPhoneNumber)) {
             region = null;
         } else {
-            return true; // If the number cannot be interpreted as an international or possible UK phone number, do not attempt to validate it.
+            return false; // If the number cannot be interpreted as an international or possible UK phone number, do not attempt to validate it.
         }
 
         try {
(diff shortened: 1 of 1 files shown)


keycloak/providers/registration-form/src/main/java/uk/gov/beis/mspsds/keycloak/providers/RegistrationMobileNumber.java
```diff
@@ -61,7 +61,7 @@ private static boolean isPhoneNumberValid(String phoneNumber) {
         } else if (isInternationalNumber(formattedPhoneNumber)) {
             region = null;
         } else {
-            return true; // If the number cannot be interpreted as an international or possible UK phone number, do not attempt to validate it.
+            return false; // If the number cannot be interpreted as an international or possible UK phone number, do not attempt to validate it.
```
tom-bell-softwire: This means that any non-mobile UK number (e.g. 020 7204 3245) is marked as invalid. I think that's probably fine, as we specifically ask for a mobile number, but it would be good to rename the `isPossibleNationalNumber` method to make it clearer that we're checking for a national mobile number.

tom-bell-softwire:
Actually, looking at the comments on the ticket, I wonder if we should also be using the `getNumberType()` method of `libphonenumber` to check if it's a number that can receive SMS messages.

Perhaps we want to check if the parsed number is valid and its type is either `MOBILE` or `FIXED_LINE_OR_MOBILE`? More details [here](https://github.com/googlei18n/libphonenumber/blob/master/FAQ.md#what-types-of-phone-numbers-can-smss-be-sent-to).

PR #734, opened 2019-03-11
COSBETA-173: Fix bug where uploading formulation file didn't update notification state

DWRendell:
…tification state

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


aidayounesi:
Thank you.

PR #738, opened 2019-03-11
Add JS polling to Cosmetics

raamSoftwire:
Adds polling for Webpacker so you shouldn't need to restart web before seeing JS changes anymore


jasiekmiko:
For reference, this is the mspsds counterpart - sorry for not including the cosmetics change straight away!
https://github.com/UKGovernmentBEIS/beis-mspsds/pull/659

PR #739, opened 2019-03-13
MSPSDS-1047 Make keycloak test mocking setup better

jasiekmiko:
## Description
KeycloakClient is a much better boundary at which to set our mocks,
rather than reverse-engineering the internals of the gem. It gives us
a clear idea of how far which test should care re: set-up and it makes
us less reliant on the particular implementation

I'd like to push this in separately to the rest of MSPSDS-1047 to get early feedback on it and make sure it doesn't get out of sync too much with everyone else's tickets :)

## Checklist:

### DIFF
--- a/mspsds-web/test/test_helper.rb
+++ b/mspsds-web/test/test_helper.rb
@@ -18,6 +18,11 @@ require "rspec/mocks/standalone"
 class ActiveSupport::TestCase
   include ::RSpec::Mocks::ExampleMethods
 
+  def initialize *args
+    @keycloak_client_instance = Shared::Web::KeycloakClient.instance
+    super *args
+  end
+
   # Setup all fixtures in test/fixtures/*.yml for all tests in alphabetical order.
   fixtures :all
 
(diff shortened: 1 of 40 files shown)


mspsds-web/test/test_helper.rb
```diff
@@ -37,90 +42,91 @@ def setup
...
-      groups = teams.map(&:id)
-      groups << organisation.id
-      user_groups = user_groups.map { |group| group[:id] == user.id ? { id: user.id, groups: groups } : group }.to_json
-    end
+  # On top of mocking out external services, this method also sets the user to an initial,
+  # sensible value, but it should only be run once per test.
+  # To change currently logged in user afterwards call `sign_in_as(...)`
+  def mock_out_keycloak_and_notify(user_name: "User_one", team_admin: false)
+    @users = [admin_user,
+              test_user(name: "User_one", id: "75fda9a1-786d-4ace-ad20-76afa4f39111"),
```
broder: Is the specific `id` here significant?

mspsds-web/test/test_helper.rb
```diff
@@ -37,90 +42,91 @@ def setup
...
+              test_user(name: "User_two"),
+              test_user(name: "User_three")].map(&:attributes)
+    @organisations = organisations.map(&:attributes)
+    @teams = all_teams.map(&:attributes)
+    @team_users = [
```
broder: We could maybe use `@team_users = [...].each_with_index.map { ... }` to avoid needing the `next_id` method.

jasiekmiko: ooh yeah, much nicer!

mspsds-web/test/test_helper.rb
jasiekmiko: turns out no :)

PR #740, opened 2019-03-13
COSBETA-176: Index document in the after_save hook

kicferk1:
Add elasticsearch indexing when a notification is updated from non-complete to complete. 

## Checklist:

### DIFF
--- a/cosmetics-web/app/models/notification.rb
+++ b/cosmetics-web/app/models/notification.rb
@@ -24,7 +24,6 @@ class Notification < ApplicationRecord
 
   before_save :add_product_name, if: :will_save_change_to_product_name?
   before_save :add_import_country, if: :will_save_change_to_import_country?
-  after_save :update_elasticsearch_index
 
   def self.duplicate_notification_message
     "Notification duplicated"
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/models/notification.rb
```diff
@@ -24,6 +24,7 @@ class Notification < ApplicationRecord
 
   before_save :add_product_name, if: :will_save_change_to_product_name?
   before_save :add_import_country, if: :will_save_change_to_import_country?
+  after_save :update_elasticsearch_index
```
broder: Could / should we do this in an `aasm` callback in the `submit_notification` event? https://github.com/aasm/aasm#callbacks

kicferk1: We could. I was worried that if we do that, and then add a different event that transitions(from a possibly different state) to `notification_complete` this bug could come back. Is that likely to happen?

broder: I don't think there will be another event to go to `notification_complete` but @DWRendell or @tom-bell-softwire might know more?

tom-bell-softwire: I think all notifications should transition via the `submit_notification` event (even if submitting an update/correction in the future), so this is probably the best place to add the callback.

PR #741, opened 2019-03-13
MSPSDS-6: Send logs to S3

broder:
## Description
This contains the code for a new service which will aggregate the logs from all of our applications and then forward them onto Logit and S3.

## Checklist:


infrastructure/fluentd/plugins/out_logit.rb
```diff
@@ -0,0 +1,94 @@
...
+      def prepare_data_to_send(tag, time, record)
+        # Just forward on the message
+        "#{record["message"]}\n"
+      end
+  
+      def connect()
+        Timeout.timeout(@connect_timeout) do
+          socket = TCPSocket.open("#{resolved_host()}", @port)
+          ssl_context = OpenSSL::SSL::SSLContext.new()
+          # TODO implement mutual tls
```
jasiekmiko: Is this part of this work?

infrastructure/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Fluentd
+
+We're using [fluentd](https://www.fluentd.org/) to aggregate the logs and send them to both an [ELK stack](https://www.elastic.co/elk-stack) and S3 bucket for long term storage.
+
+#### Deployment
+
+Create or target a common space using `cf create-space common` or `cf target -o beis-mspsds -s common`.
+
+Deploy the fluentd app by running `cf push --no-start --hostname <fluentd hostname>` from the `fluentd` folder.
```
jasiekmiko: Where would `<fluentd hostname>` and `<fluentd domain>` come from?

infrastructure/fluentd/plugins/out_logit.rb
broder: Unfortunately, that's verbatim from https://rubygems.org/gems/fluent-plugin-logit 😢 Do you think I should remove it to avoid confusion?

infrastructure/README.md
broder: You'd make something up for `<fluentd hostname>` and the command would then return `<fluentd domain>` - should I make that clearer?

jasiekmiko: I think it's worth it - the less one has two guess/assume when following these the more helpful I find they are :)

infrastructure/fluentd/plugins/out_logit.rb
jasiekmiko: ah.. Are there *any* changes? If so worth highlighting what our adaptation is, otherwise - how come we're not just pulling in that gem?

PR #742, opened 2019-03-13
COSBETA-164: Populate first notified date from CPNP

kicferk1:
Extract earliest dateModified from cpnp export file when notification is uploaded, and store it in our model. 

## Checklist:


cosmetics-web/app/views/application/_registered_product.html.slim
```diff
@@ -12,17 +12,28 @@ li.product
       | First registered in the UK:
     dd
       = notification.created_at.strftime("%d/%m/%Y")
+    br
```
tom-bell-softwire: Rather than add `<br>` elements, I think we should be using something like the `.govuk-clearfix` class to ensure that `<dt>` blocks appear on the next line.

Also, I've just spotted that we're missing a `<dl></dl>` block wrapping the [description list](https://www.w3schools.com/tags/tag_dl.asp). Mind adding it?

cosmetics-web/app/views/application/_registered_product.html.slim
```diff
@@ -12,17 +12,28 @@ li.product
...
     - if notification.cpnp_reference.present?
       dt
         | EU reference number:
       dd
         = notification.cpnp_reference
+      br
+
+    - if notification.cpnp_notification_date.present?
+      dt
+        | EU registration date:
```
tom-bell-softwire: Can you check with Ralph/Adrie what copy we should be using, and whether it should be shown here or on the edit/show page (or both)? I suspect it would be something like "First notified in the EU", but it would be good to confirm with them.

cosmetics-web/app/views/application/_unfinished_product.html.slim
```diff
@@ -8,12 +8,21 @@ li.product
...
+    br
 
     - if notification.cpnp_reference.present?
       dt
         | EU reference number:
       dd
         = notification.cpnp_reference
+      br
+
+    - if notification.cpnp_notification_date.present?
```
tom-bell-softwire: Same here – do we want to show it on the unfinished product entries?

PR #745, opened 2019-03-13
COSBETA-194 first run through manual journey

TiagoSusanoPinto:
## Description
Added nano-materials page (only the pre-defined nano-elements and the expose route and condition)
Added sub-sub-category page for the component
Added CMRs page
Added page for formulation type, frame formulation, and upload a formulation file
Added multi-component page (where it's possible to add new components and remove old ones)

It wasn't added system tests

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/notification_build_controller.rb
+++ b/cosmetics-web/app/controllers/notification_build_controller.rb
@@ -2,7 +2,12 @@ class NotificationBuildController < ApplicationController
   include Wicked::Wizard
   include Shared::Web::CountriesHelper
 
-  steps :add_product_name, :is_imported, :add_import_country, :single_or_multi_component, :add_product_image
+  steps :add_product_name,
+        :is_imported,
+        :add_import_country,
+        :single_or_multi_component,
+        :add_new_component,
+        :add_product_image
 
   before_action :set_notification
   before_action :set_countries, only: %i[show update]
@@ -17,19 +22,10 @@ class NotificationBuildController < ApplicationController
       render_single_or_multi_component_step
     when :is_imported
       render_is_imported_step
+    when :add_new_component
+      render_add_new_component_step
     when :add_product_image
-      if params[:image_upload].present?
-        params[:image_upload].each do |image|
-          image_upload = @notification.image_uploads.build
-          image_upload.file.attach(image)
-          image_upload.filename = image.original_filename
-        end
-        @notification.add_product_image
-        render_wizard @notification
-      else
-        @notification.errors.add :image_uploads, "You must upload at least one product image"
-        render step
-      end
+      render_add_product_image_step
     else
       @notification.update(notification_params)
 
@@ -74,14 +70,20 @@ private
   def render_single_or_multi_component_step
     case params[:single_or_multi_component]
     when "single"
-      @notification.components.build
-      @notification.save
-      redirect_to new_responsible_person_notification_component_build_path(@notification.responsible_person, @notification, @notification.components.first)
+      @notification.components.destroy_all if @notification.is_multicomponent?
+      single_component = @notification.components.empty? ? @notification.components.create : @notification.components.first
+      redirect_to new_responsible_person_notification_component_build_path(@notification.responsible_person, @notification, single_component)
     when "multiple"
-      # TODO COSBETA-10 Implement multiple components
-      @notification.components.build
-      render_wizard @notification
-    when ""
+      if @notification.is_multicomponent?
+        render_wizard @notification
+      else
+        @notification.components.destroy_all
+        @notification.components.build
+        @notification.components.build
+        @notification.save
+        redirect_to new_responsible_person_notification_component_build_path(@notification.responsible_person, @notification, @notification.components.first)
+      end
+    else
       @notification.errors.add :components, "Must not be nil"
       render step
     end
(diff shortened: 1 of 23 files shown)


cosmetics-web/config/webpacker.yml
```diff
@@ -69,7 +69,7 @@ development:
     headers:
       'Access-Control-Allow-Origin': '*'
     watch_options:
-      poll: true
+#      poll: true
```
tom-bell-softwire: We probably don't want to commit this...

TiagoSusanoPinto: I've end up delete this line, so I could run the application. Should I revert it? and just delete locally when needed?

cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -101,4 +103,37 @@ def render_is_imported_step
...
+
+  def render_add_new_component_step
+    if params.key?(:remove_component)
+      remove_component_id = params[:remove_component].to_i
+      componet_to_remove = @notification.components.select { |component| component.id == remove_component_id }
+      @notification.components.delete(componet_to_remove)
+      @notification.components.create if @notification.components.length < 2
+      render step
+    elsif params.key?(:add_component) && params[:add_component]
+      unvalid_multicomponents = @notification.get_unvalid_multicomponents
```
DWRendell: 'invalid_components' or 'non_valid_components' might be a little clearer here

cosmetics-web/app/models/component.rb
```diff
@@ -66,13 +66,21 @@ def formulation_required?
...
+  def self.get_parent_category(category)
+    PARENT_OF_CATEGORY[category&.to_sym]
+  end
+
+  def self.get_parent_of_categories
+    PARENT_OF_CATEGORY
+  end
+
+  def is_valid_multicomponent?
+    !(name.nil? || name == "")
```
DWRendell: You can replace this with `name.present?`, it will return false if `name` is nil or the empty string

cosmetics-web/app/views/component_build/add_cmrs.html.slim
```diff
@@ -0,0 +1,44 @@
...
+    p
+      = link_to "View carcinogenic, mutagenic or toxic for reproduction substances (opens in a new tab).",
+              cmr_list_url, target: :_blank, rel: 'noopener'
+    p
+      | Apart from in exceptional cases, these substances are banned for use in cosmetic products.
+      | &nbsp
+      = link_to "Read about carcinogenic, mutagenic or toxic for reproduction substances that aren’t banned" + \
+                " (opens in a new tab).", cmr_read_url, target: :_blank, rel: 'noopener'
+
+      = form_for(@component, url: wizard_path, method: :put) do |form|
```
DWRendell: This whole form block should be indented one more tab left I think, there's something a little odd going on where it's sort of inside the paragraph tag above, and an empty paragraph tag is getting added after the form.

cosmetics-web/app/views/component_build/select_category.html.slim
```diff
@@ -0,0 +1,19 @@
...
+          responsible_person_notification_component_build_path(@component.notification.responsible_person,
+                  @component.notification, @component, previous_step),
+          class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      | What category of cosmetic is the product?
+      = form_for(@component, url: wizard_path, method: :put) do |form|
+        = render "form_components/autocomplete_select", key: :sub_sub_category, form: form, show_all_values: true,
+                items: get_sub_sub_categories.map { |category_key, _category_text| \
```
DWRendell: You could replace this with `get_sub_sub_categories.keys.map { |category_key| \` to get rid of the unused `category_text` variable

cosmetics-web/app/views/component_build/select_formulation_type.html.slim
```diff
@@ -0,0 +1,17 @@
...
+- content_for :after_header do
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_for(@component, url: wizard_path, method: :put) do |form|
+      - heading = capture do
+        h1.govuk-heading-xl How do you want to give the formulation of the product?
+      = render "form_components/govuk_radios", form: form, key: :notification_type,
+              errorMessage: @no_exposure_route_selected && { text: "Please select an option before continuing" },
```
DWRendell: This needs updating, looks like a leftover from the nano materials form, and at the moment a 500 error is thrown if you don't select a formulation type

cosmetics-web/app/views/component_build/select_frame_formulation.html.slim
```diff
@@ -0,0 +1,14 @@
+- content_for :page_title, "Add formulation"
+- content_for :after_header do
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      | Choose frame formulation
+    = form_for(@component, url: wizard_path, method: :put) do |form|
+      = render "form_components/autocomplete_select", key: :frame_formulation, form: form, show_all_values: true,
+              items: Component.frame_formulations.map { |formulation_key, _formulation_text| \
```
DWRendell: You can use `.keys.map` here as well to get rid of the unused `_formulation_text`

cosmetics-web/app/views/component_build/number_of_shades.html.slim
```diff
@@ -1,9 +1,12 @@
 - content_for :page_title, "Single or multi component"
 - content_for :after_header do
-  = link_to "Back",
-          responsible_person_notification_build_path(@component.notification.responsible_person,
-                                                     @component.notification, :single_or_multi_component),
-          class: "govuk-back-link"
+  - if @component.notification.is_multicomponent?
```
DWRendell: It might be a little cleaner to do this in the way you've done it in add_product_image.html.slim, where you set the step in a short ternary operator and avoid repeating the whole call to `link_to`

cosmetics-web/spec/controllers/notification_build_controller_spec.rb
```diff
@@ -78,7 +78,7 @@
     # TODO COSBETA-10 Update this test
```
DWRendell: Could you please get rid of this TODO, and rename this test to something like "creates two components if single_or_multi_component set to multiple"? (the name was already inaccurate, I think that was my fault, sorry)

cosmetics-web/app/views/component_build/number_of_shades.html.slim
TiagoSusanoPinto: Since both 'Back' are in different paths (if it's not a multicomponent, we want to back to the notification path, otherwise we want to keep in the component path), I've tried to create a previous_path_link with a ternary operator, but since it's multiple lines, the rubocop started shouting. End up reverted back to this

cosmetics-web/spec/controllers/notification_build_controller_spec.rb
TiagoSusanoPinto: Done

cosmetics-web/app/views/component_build/select_formulation_type.html.slim
TiagoSusanoPinto: Good catch!

PR #748, opened 2019-03-14
MSPSDS-981 feat: on case creation assign to current user by default

MatthewBurstein:
## Checklist:


mspsds-web/test/models/investigation_test.rb
```diff
@@ -29,7 +29,7 @@ def pundit_user
   end
 
   test "should create activities when investigation is created" do
-    assert_difference"Activity.count" do
+    assert_difference"Activity.count", 2 do
```
broder: Do we definitely want 2 activities here?

Also, we should put a space between `assert_difference` and `"Activity.count"`.

MatthewBurstein: Yes good spot. @edwardhorsford has asked that they be one activity.

mspsds-web/app/models/investigation.rb
```diff
@@ -21,8 +21,9 @@ class Investigation < ApplicationRecord
   validates_length_of :non_compliant_reason, maximum: 10000
   validates_length_of :hazard_description, maximum: 10000
 
-  after_save :create_audit_activity_for_assignee,
-             :create_audit_activity_for_status, :create_audit_activity_for_visibility
+  after_save :create_audit_activity_for_status, :create_audit_activity_for_visibility
```
broder: Should we make all of these callbacks on `after_update`?

MatthewBurstein: I spoke to @edwardhorsford about this. He said leave it as is for the moment as this is something he intends to come back and review anyway.

PR #750, opened 2019-03-14
MSPSDS-6: Setup blue-green deployments

broder:
## Description
This should enable zero downtime deploys.

I'd need to apply similar changes to cosmetics-web and keycloak before this goes in. I wanted to see if anyone has any feedback on the approach before updating the cosmetics and keycloak scripts though!

## Checklist:


mspsds-worker/README.md
```diff
@@ -22,36 +20,25 @@ described in [the root README](../README.md#deployment).
...
-    cf set-env mspsds-worker NOTIFY_API_KEY XXX
+There's one final environment variable which is the URL for the website and is used for sending emails:
 
-See the GOV.UK Notify account section in [the root README](../README.md#gov.uk-notify) to get this value.
-
-    cf set-env mspsds-worker SENTRY_DSN XXX
-    cf set-env mspsds-worker SENTRY_CURRENT_ENV [int|staging|prod]
-
-See the Sentry account section in [the root README](../README.md#sentry) to get this value.
+    cf set-env mspsds-worker MSPSDS_HOST XXX
```
jasiekmiko: I can't quite pin down why this isn't set using the `cf cups` method too?

broder: I thought that the `MSPSDS_HOST` variable should be stored as environment variable for `mspsds-web` such that we can set it easily in the deployment scripts (although I guess we could just bind and unbind something like `mspsds-host-env` and `mspsds-host-temp-env`). This was then for consistency with `mspsds-web`.

What do you think about using an env variable vs having `mspsds-host-env` and `mspsds-host-temp-env`?

mspsds-web/deploy.sh
```diff
@@ -8,27 +8,54 @@ set -ex
...
 fi
+
+# TODO smoke test or manual confirmation?
+
+# Unmap the temporary hostname from the new app
+cf unmap-route $NEW_APP $DOMAIN --hostname $NEW_HOSTNAME
+
+# Map the live hostname to the new app
+cf set-env $NEW_APP MSPSDS_HOST "https://$HOSTNAME.$DOMAIN"
+cf restage $NEW_APP
```
broder: If the environment variable isn't used at build time, it's safe to `restart` rather than `restage`
```suggestion
cf restart $NEW_APP
```

mspsds-worker/README.md
jasiekmiko: It's a bit weird to have it in a different source than all the rest if we can do it with the services, so I'd opt for services unless it becomes more complicated. If it does then worth explaining why it's different here :)

tom-bell-softwire:
Looks great, thanks!

As discussed, we'll need to consider how we handle `MSPSDS_HOST` once we have a domain name for the service (same for Cosmetics, too, which may be more fiddly).

broder:
I'm probably going to park this for the rest of the week to focus on bugs and the like. The things still left to do are think about blue-green deployments for Keycloak (which I'm not 100% convinced about yet) and testing that everything works in different situations (e.g. if the last deployment fails)

cosmetics-web/deploy.sh
```diff
@@ -8,27 +8,63 @@ set -ex
...
-HOSTNAME=cosmetics-$SPACE
-COMPONENT=cosmetics-web
+DOMAIN=submit-cosmetic-product-notifications.service.gov.uk
+if [[ $SPACE == "prod" ]]; then
+    HOSTNAME=www
+else
+    HOSTNAME=$SPACE
+fi
+APP=cosmetics-web
+APP_PREEXISTS=$(cf app $APP && echo 0 || echo 1)
```
jasiekmiko: what's the `echo 0 || echo 1` doing? Is it a make-your-own ternary statement?

cosmetics-web/deploy.sh
```diff
@@ -8,27 +8,63 @@ set -ex
...
-    # Route to the maintenance page
-    cf map-route maintenance $DOMAIN --hostname $HOSTNAME
-    cf unmap-route $COMPONENT $DOMAIN --hostname $HOSTNAME
+if [[ ! $APP_PREEXISTS ]]; then
+    # We don't need to go any further
+    exit 0
 fi
 
-cf push -f ./$COMPONENT/manifest.yml $( [[ $APP_DOES_NOT_EXIST ]] && printf %s "--hostname $HOSTNAME" || printf %s "--no-route" ) $( [[ $NO_START ]] && printf %s "--no-start" )
+# TODO smoke test or manual confirmation?
```
jasiekmiko: Is that a separate ticket?

mspsds-web/deploy.sh
```diff
@@ -8,27 +8,63 @@ set -ex
 # The caller should have the following environment variables set:
 #
 # SPACE: the space to which you want to deploy
-# If NO_START is set the app won't be started
 
-DOMAIN=london.cloudapps.digital
-HOSTNAME=mspsds-$SPACE
-COMPONENT=mspsds-web
+DOMAIN=product-safety-database.service.gov.uk
```
jasiekmiko: Are we moving other envs to e.g. `staging.product-safety-database.service.govuk.uk` too? neat! :)

cosmetics-web/deploy.sh
broder: Yeah, true (0) and false (1) are a little bit backwards in bash...

mspsds-web/deploy.sh
broder: Yep!

PR #752, opened 2019-03-14
MSPSDS-1047 Invite user to team

jasiekmiko:
## Description

A couple of outstanding bits an pieces:
- [x] Email content
- [x] Test notify email (non-KC)
- [x] Flash message for success
- [x] Error messages content finalisation

but it'd be great to get a first pass at this point!

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -1,19 +1,61 @@
...
 private
 
   def set_user_teams
     @teams = User.current.teams
   end
 
   def set_team
+    User.all(force:true)
     @team = Team.find_by!(id: params[:id])
     authorize @team
```
kicferk1: Not sure if there is any nice(or necessary) solution, but it is surprising we are authorising team when we really mean authorising user to invite someone to team.

jasiekmiko: I *think* that's a quirk of pundit's method names. Much like in 
https://github.com/UKGovernmentBEIS/beis-mspsds/blob/141b7c5f6cd3000a87cb18c5fa7fb0224f5e7581/mspsds-web/app/controllers/investigations_controller.rb#L124

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -1,19 +1,61 @@
...
+  before_action :set_team, only: %i[show invite_to]
+  before_action :set_new_user, only: :invite_to
 
+  # GET /teams, GET /my-teams
   def index; end
 
   def show; end
 
+  # GET /teams/:id/invite, PUT /teams/:id/invite
+  def invite_to
```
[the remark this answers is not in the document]
jasiekmiko: I've gone for a combination of that and splitting out methods :)

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -1,11 +1,32 @@
...
```
broder: Is there something which stops us from calling the action `invite`?

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -16,4 +37,38 @@ def set_team
...
+    invite_user user
+  end
+
+  def invite_user(user)
+    @team.add_user user
+    email = NotifyMailer.user_added_to_team user.email,
+                                            name: user.full_name,
+                                            team_page_url: team_url(@team),
+                                            team_name: @team.name,
+                                            service_name: "Product safety database",
```
broder: Should we hardcode this in the template for now?

mspsds-web/app/mailers/notify_mailer.rb
```diff
@@ -47,4 +48,10 @@ def investigation_created(investigation_pretty_id, name, email, investigation_ti
 
     mail(to: email)
   end
+
+  def user_added_to_team(email, personalisation)
```
broder: I think we should explicitly define the arguments here. Otherwise, if someone were to use this method in future, they'd need to infer from other usages or look at the template/

mspsds-web/app/views/teams/invite_to.html.slim
```diff
@@ -0,0 +1,25 @@
+- content_for :page_title, "My teams"
+- content_for :after_header do
+  = link_to "Back", team_path(@team), class: "govuk-back-link"
+
+= form_with model: @new_user, url: invite_to_team_path(@team), method: :put do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds-from-desktop
+      - if form.object.errors.added? :email_address, :member_of_another_organisation
```
broder: It feels a little crap that we're treating this case differently to all other error messages... I'm guessing that you've already tried to push back on the designs?

mspsds-web/app/controllers/teams_controller.rb
```diff
@@ -16,4 +37,38 @@ def set_team
...
+                                            inviting_team_member_name: User.current.full_name
+    email.deliver_later
+  end
+
+  def create_and_invite_new_user
+    user = User.create_new @new_user.email_address
+    @team.add_user user
+    Shared::Web::KeycloakClient.instance.send_required_actions_welcome_email user.id, root_url
```
broder: Should this be part of the `User.create_new` method? It feels a little weird that the controller needs to know about keycloak and our signup flow...

mspsds-web/app/controllers/teams_controller.rb
jasiekmiko: It makes the helpers read much more naturally, e.g. `invite_to_team_url` - worth the discrepancy imo, but happy to be convinced otherwise :)

mspsds-web/app/controllers/teams_controller.rb
jasiekmiko: I don't think so - i also don't think the model should know about these, really. In architectures I'm more familiar with, I'd be tempted to move this logic to some kind of "user service" - not sure what the best equivalent here would be.
In particular, I think we only want to send the email once the user has been made a member of the team.

mspsds-web/app/views/teams/invite_to.html.slim
jasiekmiko: No, I haven't actually - guess I focused on "can" rather than "should" here.
I could do that, or potentially push `\n` detection and treatment like this into the components "glue" to have it be applied universally. Do you have a preference?

mspsds-web/app/mailers/notify_mailer.rb
jasiekmiko: Good shout :)

mspsds-web/app/views/teams/invite_to.html.slim
broder: I'd rather have it as a universal thing that having this special case. It might be worth checking with Ed if this is something that we'll have in other places. Also, does the design system support new lines in error messages?

mspsds-web/app/controllers/teams_controller.rb
broder: I'd generally try and stick to the framework conventions as closely as possible but this is a fairly minor deviation so it's probably fine.

mspsds-web/app/controllers/teams_controller.rb
broder: If we keep it in the controller, then both the controller and the model know about keycloak. I do also think that it _ideally_, the model wouldn't know about keycloak at all but that doesn't align very well with rails' default ways of doing things [i.e. `User.all`](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/user.rb#L22).

Given that the model already knows about keycloak, I'd put this code in the model, rather than adding another layer that knows about keycloak. Happy to chat further about it!

mspsds-web/app/views/teams/invite_to.html.slim
```diff
@@ -0,0 +1,29 @@
...
+        / We treat this message specially due to the new line characters
+        / The extra logic here isn't ideal - moving it to a generic nl2br helper or removing the long-winded error
+          message in the future would both be advisable.
+        - message = form.object.errors.full_messages_for(:email_address).first
```
jasiekmiko:
```suggestion
        / TODO MSPSDS-1113 Standardise line ending handling
        - message = form.object.errors.full_messages_for(:email_address).first
```

PR #756, opened 2019-03-15
MSPSDS-931 Confirm assignment

samlionheart:
Assignment confirmation page where you can add a comment
## Description

Turned assignment journey into a wizard, where you can see who you have selected and then add an optional comment. Future work to be done on adding a "check" to see if you will lose the ability to assign the case after confirmation.
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    render_wizard
+  end
+
+  def new
+    clear_session
+    redirect_to wizard_path(steps.first, request.query_parameters)
+  end
+
+  def update
+    redirect_to next_wizard_path
```
broder: Should this be something like:
```
if assignee_valid?
  redirect_to next_wizard_path
else
  render step
end
```
?
We could then take that logic out of `store_assignee`

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+                                               params[:investigation][:select_other_team]
+                                             when "someone_else"
+                                               params[:investigation][:select_someone_else]
+                                             else
+                                               params[:investigation][:assignable_id]
+                                             end
+    params.require(:investigation).permit(:assignable_id)
+  end
+
+  def store_assignee
```
broder: This is doing quite a bit more than just storing the assignee. Would it work if we moved the valid and respond logic to `update` and then made this:
```
def store_assignee
  session[:assignable_id] = filter_assignee_params[:assignable_id]
end
```
?

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+  def clear_session
+    session[:assignable_id] = nil
+  end
+
+  def set_investigation
+    @investigation = Investigation.find_by!(pretty_id: params[:investigation_pretty_id])
+    authorize @investigation, :show?
+  end
+
+  def filter_assignee_params
```
broder: Should this just be `assignee_params`?

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    session[:assignable_id] = params[:investigation][:assignable_id]
+    if session[:assignable_id].blank?
+      @investigation.errors.add(:assignable_id, :invalid, message: "Select assignee")
+      respond_to_invalid_data
+      return
+    end
+  end
+
+  def find_potential_assignee
+    @potential_assignees = User.where(id: session[:assignable_id]) + Team.where(id: session[:assignable_id])
```
broder: Should this return a single `User` or `Team` rather than an array?

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+                                             else
+                                               params[:investigation][:assignable_id]
+                                             end
+    params.require(:investigation).permit(:assignable_id)
+  end
+
+  def store_assignee
+    filter_assignee_params
+    session[:assignable_id] = params[:investigation][:assignable_id]
+    if session[:assignable_id].blank?
```
broder: Should we be validating on more than just there being a string? Should we make sure that it corresponds to a `User` or `Team`?

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def respond_to_update
+    respond_to do |format|
+      if @investigation.save
+        format.html { redirect_to @investigation, notice: "#{@investigation.case_type.titleize} was successfully updated." }
+        format.json { render :show, status: :ok, location: @investigation }
+      else
+        @investigation.restore_attributes
+        format.html { render step }
```
broder: I don't think `step` would be defined on the `create` action

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def update
+    redirect_to next_wizard_path
+  end
+
+  def create
+    @investigation.assignee = @potential_assignees.first
+    @investigation.assignee_comment = params[:investigation][:assignee_comment]
+    @investigation.save
+    respond_to_update
```
broder: We're saving the investigation in `respond_to_update` as well as on the line above - should we get rid of `respond_to_update` and just put the `redirect_to` line in here?

mspsds-web/app/models/investigation.rb
```diff
@@ -6,6 +6,7 @@ class Investigation < ApplicationRecord
 
   attr_accessor :status_rationale
   attr_accessor :visibility_rationale
+  attr_accessor :assignee_comment
```
broder: Should we call this `assignee_rationale` to be consistent with the above?

mspsds-web/app/views/investigations/assign/confirm_assignment_change.html.slim
```diff
@@ -0,0 +1,21 @@
...
+              url: investigation_assign_index_path(@investigation), method: :post) do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds-from-desktop
+      = render "form_components/govuk_error_summary", form: form
+      = render "minimal_investigation_heading", investigation: @investigation, title: "Confirm assignment change"
+      p
+        | You are assigning this project to:
+        - @team_name = @potential_assignees.first.respond_to?(:name) ? "#{@potential_assignees.first.name} — " : nil
+        strong = " #{@team_name}#{@potential_assignees.first.display_name}"
+  .govuk-form-group
```
jasiekmiko: This element ought to be inside the `.govuk-grid-column-two-thirds-from-desktop` so it gets the proper sizing. It might render the `classes` parameter below no longer needed?

mspsds-web/test/system/investigation_assignee_test.rb
```diff
@@ -15,18 +15,32 @@ class InvestigationAssigneeTest < ApplicationSystemTestCase
...
     choose @team.name, visible: false
-    click_on "Assign"
+    click_on "Continue"
+    click_on "Confirm change"
     assert_text "Assigned to\n#{@team.name}"
     click_on "Activity"
     assert_text "Assigned to #{@team.name}"
   end
+
+  test "should add comment current to assignment activity" do
```
jasiekmiko: I think that's an unnecessary `current` in there? :)

mspsds-web/app/models/audit_activity/investigation/update_assignee.rb
```diff
@@ -20,7 +21,14 @@ def title
   end
 
   def email_update_text
-    "#{investigation.case_type.titleize} was assigned to #{investigation.assignee.display_name} by #{source&.show&.titleize}."
+    "#{investigation.case_type.titleize} was assigned to #{investigation.assignee.display_name} by #{source&.show&.titleize}.\n" + assignee_rationale_email_text.to_s
```
broder: We don't display the rationale for e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/audit_activity/investigation/update_visibility.rb#L9 or https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/audit_activity/investigation/update_status.rb#L8 - should we be consistent?

mspsds-web/app/models/audit_activity/investigation/update_assignee.rb
```diff
@@ -1,7 +1,8 @@
 class AuditActivity::Investigation::UpdateAssignee < AuditActivity::Investigation::Base
   def self.from(investigation)
     title = investigation.assignee.id.to_s
-    super(investigation, title)
+    body = investigation.assignee_rationale
```
broder: We sanitize `status_rationale` https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/audit_activity/investigation/update_status.rb#L4. Should we do the same here?

We should also update https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/models/audit_activity/investigation/update_visibility.rb#L5 if we do want them to be consistent.

mspsds-web/app/views/investigations/assign/choose.html.slim
```diff
@@ -0,0 +1,28 @@
...
+    - if policy(@investigation).assign?
+      = render "investigations/assign/assignee_form", investigation: @investigation do
+        = heading
+    - else
+      = heading
+      p
+        = "You do not have permission to assign this #{@investigation.case_type}."
+        br
+        | Contact
+        strong = " #{@investigation.assignee.display_name} "
```
broder: Rather than add a space before and after, you could use `' Contact` (which adds a trailing space) above and `|  to change assignment.` (which adds a leading space) below. This would then become `= "@investigation.assignee.display_name"`

mspsds-web/app/controllers/investigations/assign_controller.rb
```diff
@@ -54,36 +58,21 @@ def filter_assignee_params
...
     end
+    @investigation.errors.empty?
   end
 
-  def respond_to_invalid_data
-    respond_to do |format|
-      format.html { render step }
-      format.json { render json: @corrective_action.errors, status: :unprocessable_entity }
-    end
+  def find_potential_assignee
```
broder: Would this work without making it an array?
```
def potential_assignee
  User.find_by(id: session[:assignable_id]) || Team.find_by(id: session[:assignable_id])
end
```
?

You could then use it in the `assignee_valid?` method above.

PR #758, opened 2019-03-15
COSBETA-167 Change sign out links to form submits to avoid them faili…

DWRendell:
…ng when users has no JS

Fixing JS in IE11 was done in https://github.com/UKGovernmentBEIS/beis-mspsds/pull/755

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/mspsds-web/test/system/keycloak_test.rb
+++ b/mspsds-web/test/system/keycloak_test.rb
@@ -12,7 +12,7 @@ class KeycloakTest < ApplicationSystemTestCase
   test "can login" do
     visit root_path
     sign_in email: "user@example.com", password: "password"
-    assert_selector "a", text: "Sign out"
+    assert_selector "form input[type=submit][value=\"Sign out\"]"
   end
 
   test "can logout" do
--- a/shared-web/app/assets/application/stylesheets/header.scss
+++ b/shared-web/app/assets/application/stylesheets/header.scss
@@ -1,15 +1,5 @@
 $navbar-link-colour: #ffffff;
 
-@mixin de-buttonify() {
-  background: none;
-  border: none;
-  outline: none;
-  box-shadow: none;
-  -webkit-box-shadow: none;
-  font-family: inherit;
-  cursor: pointer;
-}
-
 .sign-out-link {
   font-size: 16px;
   font-weight: 700;
@@ -22,7 +12,7 @@ $navbar-link-colour: #ffffff;
     text-decoration: underline;
   }
 
-  @include de-buttonify();
+  @include remove-default-button-styling;
   @include govuk-link-common;
   padding: 0;
 }
--- a/shared-web/app/assets/application/stylesheets/header.scss
+++ b/shared-web/app/assets/application/stylesheets/header.scss
@@ -1,17 +1,28 @@
 $navbar-link-colour: #ffffff;
 
-.sign-out-link {
+@mixin de-buttonify() {
   background: none;
   border: none;
-  color: $navbar-link-colour;
-  text-decoration: none;
+  outline: none;
+  box-shadow: none;
+  -webkit-box-shadow: none;
+  font-family: inherit;
+  cursor: pointer;
+}
+
+.sign-out-link {
   font-size: 16px;
-  font-family: "nta", Arial, sans-serif;
   font-weight: 700;
+
+  &:not(:focus) {
+    color: $navbar-link-colour;
+  }
 
   &:hover {
     text-decoration: underline;
   }
 
-  cursor: pointer; 
+  @include de-buttonify();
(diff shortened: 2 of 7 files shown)


shared-web/app/assets/application/stylesheets/header.scss
```diff
@@ -0,0 +1,17 @@
+$navbar-link-colour: #ffffff;
+
+.sign-out-link {
```
tom-bell-softwire: I think you might need to change the selected/highlighted stying, too. The other links on the navbar are highlighted in orange.

tom-bell-softwire:
Looks like some of the tests in `keycloak_test.rb` may need to be updated, too.

shared-web/app/assets/application/stylesheets/header.scss
broder: Would it be possible to do this by `extend`ing some govuk styles or using mixins?

We could also maybe break it into:
* Removing the default button styles using something like https://fvsch.com/styling-buttons/#reset
* Adding the `.govuk-link` and `.govuk-header__link` classes
* Filling the gaps caused by the explicit references to `a` and `:link` in https://github.com/alphagov/govuk-frontend/blob/master/package/components/header/_header.scss

I wonder how much we'd then have to manually keep in sync with the design system though...

broder: I'll ask on x-gov slack and see if anyone else has done anything similar.

broder: People pointed me in the direction of https://github.com/alphagov/govuk-frontend/blob/master/src/components/accordion/_accordion.scss#L70-L83 (as a button that acts like a link) and
```
@mixin de-buttonify() {
  background: none;
  border: none;
  outline: none;
  box-shadow: none;
  -webkit-box-shadow: none;
  font-family: inherit;
  cursor: pointer;
}

.o-text-button {
  @include de-buttonify();
  padding: 0;
  color: inherit;
}
```

They thought that having it as a `POST` was best practice but ultimately, using a `GET` wouldn't be the end of the world.

Does that help?

DWRendell: Thanks, I've updated it so the styling works on hover and tab focus. Is it worth having the de-buttonify mixin in a separate file somewhere?

broder: Maybe just in the shared-web `main.scss`?

DWRendell: That doesn't seem to work, I think it would need to be somewhere we can import into header.scss, and I think importing main would lead to a circular dependency.

I could move everything in header.scss into main?

broder: What about `remove-default-button-styling.scss`? Basically, I want to put it somewhere that people will notice that we have this code, rather than them doing something like this in a different way...

Speaking of which, are we also "de-buttoning" here: https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/cosmetics-web/app/assets/application/stylesheets/responsible_persons.scss#L19?

DWRendell: Good point, I'll use the mixin there too. It looks like those links aren't highlighted orange on tab, so should be able to fix that.

mspsds-web/test/system/keycloak_test.rb
```diff
@@ -5,7 +5,7 @@ class KeycloakTest < ApplicationSystemTestCase
     visit root_path
     sign_in email: "admin@example.com", password: "password"
     assert_current_path root_path
-    assert_selector "a", text: "Sign out"
+    assert_selector "form input[type=submit][value=\"Sign out\"]"
```
broder: Does `assert_selector "input", value: "Sign out"` work?

DWRendell:
Oops, just pushed a load of stuff that doesn't belong in this review, please ignore it I'll try and fix it...

mspsds-web/test/system/keycloak_test.rb
DWRendell: No joy :(

shared-web/app/assets/application/stylesheets/remove-default-button-styling.scss
```diff
@@ -0,0 +1,9 @@
+@mixin remove-default-button-styling() {
+  background: none;
+  border: none;
+  outline: none;
+  box-shadow: none;
+  -webkit-box-shadow: none;
```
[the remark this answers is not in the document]
DWRendell: Yep, looks like it.

PR #766, opened 2019-03-18
MSPSDS-786: Add server metric exporting

broder:
## Description
`metric-exporter` captures all metrics from spaces that it has the "Space auditor" role for (which is currently int, staging, prod and research) which are then sent to fluentd and onto Logit and S3. I'll add some dashboards soon!

## Checklist:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "keycloak/providers/sms-authenticator"]
 	path = keycloak/providers/sms-authenticator
 	url = https://github.com/UKGovernmentBEIS/keycloak-sms-authenticator-sns.git
+[submodule "infrastructure/paas-metric-exporter"]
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "keycloak/providers/sms-authenticator"]
 	path = keycloak/providers/sms-authenticator
 	url = https://github.com/UKGovernmentBEIS/keycloak-sms-authenticator-sns.git
+[submodule "infrastructure/paas-metric-exporter"]
```
jasiekmiko: Do we need to add it explicitly to list of things watched by dependabot, or does it pick up all of our submodules?

broder: I think it'll just work - https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/.dependabot/config.yml#L5

PR #769, opened 2019-03-18
MSPSDS-1094 ts intro slides

MatthewBurstein:
## Checklist:

### General testing
- [X] Test without javascript
- [X] Test on small screen
### Accessibility testing
- [X] Works keyboard only


mspsds-web/app/controllers/introductions_controller.rb
```diff
@@ -0,0 +1,19 @@
+class IntroductionsController < ApplicationController
+  include Wicked::Wizard
+
+  steps :overview, :report_products, :track_investigations, :share_data
+
+  before_action :set_has_viewed_introduction
```
broder: Does this set that the user has viewed the introduction even if they've only seen the first step? Is that intentional?

If it's to do with the skip link, maybe we should make the skip link go to an action which sets the `has_viewed_introduction` flag and then redirects to root?

mspsds-web/app/controllers/introductions_controller.rb
```diff
@@ -0,0 +1,19 @@
+class IntroductionsController < ApplicationController
+  include Wicked::Wizard
+
+  steps :overview, :report_products, :track_investigations, :share_data
```
broder: I think this controller is almost simple enough that we don't need a wizard. Maybe we'd just need
```
def overview; end
def report_products; end
def track_investigations; end
def share_data
  User.current.has_viewed_introduction = true
end
```

Using a wizard is also fine though 🙂

mspsds-web/app/models/user.rb
```diff
@@ -6,6 +6,12 @@ class User < Shared::Web::User
   has_many :team_users, dependent: :nullify
   has_many :teams, through: :team_users
 
+  has_one :user_attributes, dependent: :destroy
+
+  # Getters and setters for each UserAttributes column should be added here so they can be accessed directly
+  # from the User object via delegation.
+  delegate :has_viewed_introduction, :has_viewed_introduction=, to: :get_user_attributes
```
broder: Neat!

mspsds-web/app/models/user_attributes.rb
```diff
@@ -0,0 +1,11 @@
+class UserAttributes < ApplicationRecord
+  belongs_to_active_hash :user
+
+  def has_viewed_introduction=(value)
```
broder: Will we ever want to set this to `false`? Maybe we should just define a method like
```
def has_viewed_introduction!
  update(has_viewed_introduction: true)
end
```

mspsds-web/app/views/introductions/overview.html.slim
```diff
@@ -0,0 +1,15 @@
+- content_for :page_title, "Overview - Product safety database - GOV.UK"
```
broder: What does the page title end up looking like? We're adding another bit [here](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/views/layouts/application.html.slim#L16) - does that result in "Overview - Product safety database - GOV.UK - Market Surveillance"?

mspsds-web/app/views/introductions/report_products.html.slim
```diff
@@ -0,0 +1,12 @@
+- content_for :page_title, "Report products - Product safety database - GOV.UK"
+- content_for :after_header do
+  = render "skip_introduction"
+
+.flash-card
+  .govuk-grid-row
+    govuk-grid-column-two-thirds-from-desktop
```
broder: We're missing a `.` at the start of this classname

mspsds-web/app/views/introductions/overview.html.slim
```diff
@@ -0,0 +1,15 @@
...
+- content_for :after_header do
+  = render "skip_introduction"
+
+.flash-card
```
broder: Could / should we just apply this class to another one of the divs below or does it need to be on a separate div?

mspsds-web/config/routes.rb
```diff
@@ -13,6 +13,8 @@
     end
   end
 
+  resources :introduction, controller: :introductions
```
broder: Do we need to specify `only` here?

mspsds-web/test/system/keycloak_test.rb
```diff
@@ -1,9 +1,18 @@
 require "application_system_test_case"
 
 class KeycloakTest < ApplicationSystemTestCase
+  setup do
+    sign_out_if_signed_in
```
broder: Will we always be on a page when this is called? Should we call something like `Shared::Web::KeycloakClient.instance.logout` instead?

mspsds-web/app/controllers/introductions_controller.rb
MatthewBurstein: @edwardhorsford says the flag should be set as soon as they have seen any intro slide. This means that if they use the skip introduction link the first time then they won't be shown the slides the second time they log in.

mspsds-web/app/models/user_attributes.rb
MatthewBurstein: Agreed this is nicer

mspsds-web/app/controllers/introductions_controller.rb
broder: We could use custom logic for the skip link though?

mspsds-web/app/controllers/introductions_controller.rb
MatthewBurstein: 👍

mspsds-web/config/routes.rb
MatthewBurstein: This turned out much better using `namespace` once I removed the wizard.

mspsds-web/app/views/introductions/overview.html.slim
MatthewBurstein: It needs to be a separate div. Otherwise the -15 margin on `.govuk-grid-row` is filled with the blue background of the flash card

mspsds-web/app/views/introductions/overview.html.slim
MatthewBurstein: Yes it does. @edwardhorsford confirmed the formatting from this PR so I've added it to application.html.slim instead.

mspsds-web/app/controllers/introductions_controller.rb
MatthewBurstein: Ah I see what you mean. Have done so now :)

mspsds-web/test/system/keycloak_test.rb
MatthewBurstein: I can't get this to work because direct calls to methods on `KeycloakClient` raise a `ProcCookieTokenNotDefined` error. 
I've not had a test run fail on `sign_out_if_signed_in` since adding it. Where do you think the problem might arise?

broder: It might arise if this test is run first. Then you might not be on any page and therefore couldn't click the "Sign out" link. Maybe it's not something to worry about for now!

mspsds-web/test/controllers/homepage_controller_test.rb
```diff
@@ -2,16 +2,24 @@
 
 class HomepageControllerTest < ActionDispatch::IntegrationTest
   setup do
-    mock_out_keycloak_and_notify
+    @user = mock_out_keycloak_and_notify
```
jasiekmiko: I think I've stopped returning the user from this method in another branch, sorry! :S
Anything wrong with using `User.current` here?

mspsds-web/test/system/introduction_test.rb
```diff
@@ -0,0 +1,56 @@
+require "application_system_test_case"
+
+class IntroductionTest < ApplicationSystemTestCase
+  setup do
+    @user = mock_out_keycloak_and_notify
```
jasiekmiko: ditto :)

mspsds-web/test/system/keycloak_test.rb
```diff
@@ -1,9 +1,18 @@
...
+  end
+
+  teardown do
+    sign_out_if_signed_in
+  end
+
   test "can login" do
     visit root_path
     sign_in email: "admin@example.com", password: "password"
+    click_on "Skip introduction"
```
jasiekmiko: This feels irrelevant to a "can login" test.
If it was added to make the test work, perhaps logging in as an opss user would make more sense?
If you are deliberately testing the "skip introduction" link works here, it could probably use its own test?

mspsds-web/app/assets/application/stylesheets/flashcard.scss
```diff
@@ -0,0 +1,93 @@
+$black: #000000;
```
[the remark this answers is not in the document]
edwardhorsford: FWIW Black in GOV.UK services is never #000000 - so double reason to use the helpers.

mspsds-web/app/assets/application/stylesheets/flashcard.scss
```diff
@@ -0,0 +1,93 @@
...
+
+.flash-card {
```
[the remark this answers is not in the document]
MatthewBurstein: They came from @edwardhorsford but I don't think we have a meaningful place to link to.

mspsds-web/app/assets/application/stylesheets/flashcard.scss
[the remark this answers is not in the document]
MatthewBurstein: fixed :)

mspsds-web/app/views/introduction/track_investigations.html.slim
```diff
@@ -0,0 +1,12 @@
...
+- content_for :after_header do
+  = render "skip_introduction"
+
+.flash-card
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds-from-desktop
+      h1.govuk-heading-l Track and manage investigations
+      p.govuk-body-l
+        | Market surveillance authorities (MSAs) can report and manage routine product safety cases and track
+        |  their progress.
```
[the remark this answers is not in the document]
MatthewBurstein: 👍

PR #770, opened 2019-03-18
MSPSDS-1103: Create terms and conditions page

kicferk1:
Adds terms and conditions, privacy policy and about page. 
I still need to put in the actual content for terms and conditions and privacy policy pages that Laurence shared on slack, and there is no content for 'about' page yet. 

Adds pages visible to everyone(even people who aren't logged in). When they are logged in the pages include navbar. When they are opened in a new window a button appears to close them. Currently this is achieved with a url param, which means if someone copies url with the param they will get a button that doesn't work(browser will stop close from working). I'm not sure if there is a good way to fix it though. 

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/controllers/concerns/authentication_concern.rb
```diff
@@ -0,0 +1,15 @@
+module AuthenticationConcern
+  extend ActiveSupport::Concern
+  include Shared::Web::Concerns::AuthenticationConcern
+  def no_need_to_authenticate
```
broder: Would `skip_before_action authenticate_user!, authorize_user` on `PagesController` be simpler?

mspsds-web/app/views/pages/_close_page_button.html.slim
```diff
@@ -0,0 +1,10 @@
+button id="close-page" class="govuk-button" type="button" style="display: none;"
+  | Close
+javascript:
+  const element = document.getElementById("close-page");
```
MatthewBurstein: Is `element` defined on the global scope? Should we enclose this in `$(document).ready()` or an IIFE?

mspsds-web/app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,15 @@
+class PagesController < ApplicationController
```
broder: I wonder if we should call this `HelpController` instead. Then the routes would be something like `/help/about`, `/help/privacy-policy` and `/help/terms-and-conditions`. What do you think?

mspsds-web/app/views/homepage/non_opss.html.slim
```diff
@@ -27,7 +27,6 @@ hr.govuk-section-break.govuk-section-break--l.govuk-section-break--visible class
     p All cases in the product safety database
 
   .govuk-grid-column-one-third
-    / TODO add this column (uncomment these lines and check content) and set the path accordingly in MSPSDS-1095
-    / h2.govuk-heading-m
-    /   = link_to "About this service", about_path, class: "govuk-link govuk-link--no-visited-state"
-    / p How and when to use this service.
+    h2.govuk-heading-m
+      = link_to "About this service", "/about", class: "govuk-link govuk-link--no-visited-state"
```
broder: Should we use `about_page_path` or similar?

mspsds-web/app/views/pages/_close_page_button.html.slim
```diff
@@ -0,0 +1,10 @@
...
+  element.style.display = "block";
```
MatthewBurstein: Should this line be inside the `if` block below?

mspsds-web/app/views/layouts/_navbar.html.slim
```diff
@@ -1,19 +1,20 @@
...
-      = link_to "Products", products_path, "data-topnav": "products", class: "govuk-link--no-visited-state"
-    / In principle all our users belong to a team, but this saves crashes in case of a misconfiguration
-    - if User.current.teams.present?
-      li.right-aligned class=(params[:controller] == "teams" ? "app-navigation--current-page" : nil)
-        - if User.current.teams.count > 1
-          = link_to "Your teams", your_teams_path, "data-topnav": "teams", class: "govuk-link--no-visited-state"
-        - else
-          - team = User.current.teams.first
-          = link_to team.display_name, team_path(team), "data-topnav": "teams", class: "govuk-link--no-visited-state"
+- if User.current.present?
```
broder: Is it possible to return early from slim? Also, does the ticket say "no nav links for cases / businesses / products." or am I misunderstanding something?

mspsds-web/app/views/layouts/application.html.slim
```diff
@@ -93,8 +93,22 @@ html.govuk-template.app-html-class[lang="en"]
       .govuk-width-container
         .govuk-footer__meta
           .govuk-footer__meta-item.govuk-footer__meta-item--grow
+            .govuk-clearfix class="govuk-!-margin-bottom-2"
+              span
+                = link_to "Terms and conditions", "/terms-and-conditions?referred=true",
```
broder: Should we use URL helpers here?

mspsds-web/app/views/pages/_close_page_button.html.slim
broder: I think using ES6 syntax here will break IE. Either use ES5 or put this code in a JS file (which will be transpiled)

mspsds-web/test/system/public_pages_test.rb
```diff
@@ -0,0 +1,25 @@
+require "application_system_test_case"
+
+class PublicPagesHelper < ApplicationSystemTestCase
+  setup do
+  end
+
+  teardown do
+    reset_keycloak_and_notify_mocks
```
MatthewBurstein: Do we need to call `reset_keycloak_and_notify_mocks` if we haven't called `mock_out_keycloak_and_notify`?

mspsds-web/app/views/pages/_close_page_button.html.slim
```diff
@@ -0,0 +1,10 @@
+button id="close-page" class="govuk-button" type="button" style="display: none;"
+  | Close
```
broder: The passport close button consists of `Close<span class="visuallyhidden"> the current window</span>`

mspsds-web/app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  def terms_and_conditions; end
+
+  def privacy_policy; end
+
+  def about; end
+
+private
+
+  def set_referred
+    @referred = params[:referred]
```
broder: Is this parameter only used to decide whether or not to show the close button?

If so, the passports pages use
```
helpers.isJSWindow = function() {
  return null !== window.opener
}
```
to decide whether or not to display it

mspsds-web/app/views/pages/privacy_policy.html.slim
```diff
@@ -0,0 +1,9 @@
+- content_for :page_title, "Privacy Policy Page"
```
broder: Do we need "Page" appended to all of the titles?

mspsds-web/test/system/public_pages_test.rb
broder: Do we need to remove any mocks here? We can probably just remove the `setup` and `teardown` blocks.

mspsds-web/config/routes.rb
```diff
@@ -78,6 +78,10 @@
   get "your-teams" => "teams#index"
   resources :teams, only: %i[index show]
 
+  get "terms-and-conditions" => "pages#terms_and_conditions"
```
broder: Would something like:
```
namespace :about do
  get :terms_and_conditions, path: "terms-and-conditions"
...
end
```
be any neater?

mspsds-web/app/views/layouts/_navbar.html.slim
kicferk1: Ticket logic seems to be split in description of two tickets, the second one being described [here](https://regulatorydelivery.atlassian.net/secure/RapidBoard.jspa?rapidView=18&modal=detail&selectedIssue=MSPSDS-1095)

Ed isn't sure if we want to show navbar when user is logged in and the page is opened via footer link, and told me to keep it for now.

broder: Ah, thanks!

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
...
+      | We may change this privacy policy. In that case, the ‘last updated’ date at the bottom of this page will
+      |  also change. Any changes to this privacy policy will apply to you and your data immediately.
+      |  If these changes affect how your personal data is processed, GDS will take reasonable steps to let you know.
+      |  You can see previous versions of this page. Contact us or make a complaint Contact the Data Protection
+      |  Officer (DPO) if you: have any questions about anything in this document or think that your personal data
+      |  has been misused or mishandled.
+
+    p Data Protection Officer
+    p
+      | [INSERT DETAILS]
```
broder: Do we know what these are yet?

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
+- content_for :page_title, "Terms and conditions"
+- content_for :after_header do
+  = render "layouts/navbar"
+
+h1.govuk-heading-l = "Market Surveillance Product Safety Digital Service Terms of Use"
+.govuk-grid-row
+  .govuk-grid-column-one-half-from-desktop
+    p
+      | These Terms and conditions (
+      b = "Terms"
```
broder: Should we use `'` or `|` rather than `= "..."`?

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
...
+
+    h2.govuk-heading-m How are cookies deleted?
+    p
+      | To delete or stop cookies being placed on your device, please check the help menu of your internet browser.
+      |  Blocking cookies may reduce the functionality of this website.
+
+    h2.govuk-heading-m For more information
+    p
+      | To learn more about cookies and similar technologies, we recommend you visit:
+      = link_to " http://www.allaboutcookies.org/", "http://www.allaboutcookies.org/"
```
broder: Could we use `'` on the line above so we don't need a space at the start of the link?

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
...
+      | ) govern your use of the Market Surveillance Product Safety Digital Service,
+      |  the web pages the Service is made available on and the database of information forming part of the Service
+      |  (together, the
+      b = " Service"
+      |).
+    p
+      | Your use of the Service is also governed by GOV.UK's
+      = link_to " Terms of use.", "https://www.gov.uk/help/terms-conditions"
```
broder: We shouldn't include full stops / other punctuation in the link text.

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
...
+      | Who we are
+    p
+      | The Service is managed by the Office for Product Safety and Standards (
+      b = "OPSS"
+      |) on behalf of the Crown. OPSS is referred to as '
+      b = "we"
+      |' in these Terms.
+    p
+      | If you have any questions relating to these Terms or the Service you can contact us by emailing:
+      | OPSS.enquiries@beis.gov.uk
```
broder: Should we make this a `mailto` link?

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
...
+    h2.govuk-heading-m
+      | What's in these Terms?
+    p
+      | These Terms tell you the rules for using the Service.
+    p
+      | Click on the links below to go straight to more information on each area:
+    ul.govuk-list.govuk-list--bullet
+      li = link_to "Using the Service",
+              help_terms_and_conditions_path(anchor: "using_the_service")
```
broder: Should we make the IDs use dashes rather than underscores?

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
...
+              help_terms_and_conditions_path(anchor: "may_make_changes_terms")
+      li = link_to "We may make changes to the Service",
+              help_terms_and_conditions_path(anchor: "may_make_changes_service")
+      li = link_to "We may suspend or withdraw the Service",
+              help_terms_and_conditions_path(anchor: "may_withdraw_service")
+      li = link_to "The Service is only for users in the UK",
+              help_terms_and_conditions_path(anchor: "only_for_uk")
+      li = link_to "You must keep your account details safe",
+              help_terms_and_conditions_path(anchor: "keep_your_account_safe")
+      li = link_to "Our content on the Service ",
```
broder:
```suggestion
      li = link_to "Our content on the Service",
```

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
+- content_for :page_title, "Privacy policy"
+- content_for :after_header do
+  = render "layouts/navbar"
+
+.govuk-grid-row
+  .govuk-grid-column-one-half-from-desktop
+    p
+      | This Privacy Notice applies from 29 March 2019.
+    h1.govuk-heading-l Data Protection Notice -- Market Surveillance Product Safety Digital Service (MSPSDS)
```
broder: Should this be PSD rather than MSPSDS?

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
+- content_for :page_title, "Privacy policy"
+- content_for :after_header do
+  = render "layouts/navbar"
+
+.govuk-grid-row
+  .govuk-grid-column-one-half-from-desktop
```
jasiekmiko: should be 2/3s (same on t&cs)

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
...
+    h1.govuk-heading-l Data Protection Notice -- Market Surveillance Product Safety Digital Service (MSPSDS)
+
+    h2.govuk-heading-m Who we are
+    p
+      | The Office of Product Safety and Standards (OPSS) is responsible for operating the
+      |  Market Surveillance Product Safety Digital Service (MSPSDS). OPSS is the data controller for MSPSDS.
+      |  A data controller determines how and why personal data is processed. We may also share this data with
+      |  other UK and international data controllers who assist us with our product safety oversight mission
+      |  and have their own complimentary objectives such as the UK Health and Safety Executive
+      |  and local authorities working on product safety and market surveillance.
```
jasiekmiko: It doesn't really matter, but just as convenience, you don't need to repeat the `|` bar everywhere, you can use indentation to same effect:
```
  | Hello
    this is same line of text as hello
```

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -0,0 +1,237 @@
...
+    p class="govuk-!-margin-bottom-0" Monday to Friday, 9am to 4:30pm
+    p class="govuk-!-margin-bottom-0" Information Commissioner's Office
+    p class="govuk-!-margin-bottom-0" Wycliffe House
+    p class="govuk-!-margin-bottom-0" Water Lane
+    p class="govuk-!-margin-bottom-0" Wilmslow
+    p Cheshire SK9 5AF
+
+    p Last updated 18/03/2019
+
+    h1 Cookies policy
```
jasiekmiko: class missing

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
+- content_for :page_title, "Terms and conditions"
+- content_for :after_header do
+  = render "layouts/navbar"
+
+h1.govuk-heading-l = "Market Surveillance Product Safety Digital Service Terms of Use"
```
jasiekmiko: Should be inside the column (2/3s tho!)

mspsds-web/app/views/help/terms_and_conditions.html.slim
```diff
@@ -0,0 +1,392 @@
...
+      li = link_to "We are not responsible for viruses and you must not introduce them",
+              help_terms_and_conditions_path(anchor: "no_viruses")
+      li = link_to "Our responsibility for loss or damage suffered by you",
+              help_terms_and_conditions_path(anchor: "our_responsibility")
+      li = link_to "How we may use your personal information",
+              help_terms_and_conditions_path(anchor: "personal_information")
+      li = link_to "General terms",
+              help_terms_and_conditions_path(anchor: "general_terms")
+
+    h3.govuk-heading-m id="using_the_service" Using the Service
```
jasiekmiko: Are these meant to be visually the same as the `h2`s?

mspsds-web/app/views/help/privacy_policy.html.slim
jasiekmiko: actually, it'll let you not write that awkward extra leading space on subsequent lines!

mspsds-web/app/views/help/terms_and_conditions.html.slim
kicferk1: No idea, I just got a word document in which every heading has the same style :P

jasiekmiko: Let's double check this. Perhaps the `h3` below should be `h2` instead?

mspsds-web/app/views/help/privacy_policy.html.slim
kicferk1: Laurence knows some details, but he wants to make sure he gives me the right ones.

mspsds-web/app/views/help/privacy_policy.html.slim
```diff
@@ -3,48 +3,49 @@
...
 
     h2.govuk-heading-m Who we are
     p
       | The Office of Product Safety and Standards (OPSS) is responsible for operating the
-      |  Market Surveillance Product Safety Digital Service (MSPSDS). OPSS is the data controller for MSPSDS.
-      |  A data controller determines how and why personal data is processed. We may also share this data with
-      |  other UK and international data controllers who assist us with our product safety oversight mission
-      |  and have their own complimentary objectives such as the UK Health and Safety Executive
-      |  and local authorities working on product safety and market surveillance.
+        Market Surveillance Product Safety Digital Service (MSPSDS). OPSS is the data controller for MSPSDS.
```
broder: There are some more mentions of MSPSDS here and in the T&Cs

PR #775, opened 2019-03-19
MSPSDS-869 notification banner styles

MatthewBurstein:
@broder @jasiekmiko it would be good to get this reviewed and merged as soon as possible because my next ticket needs the component :) 

## Checklist:

### General testing
- [X] Test without javascript
- [X] Test on small screen
### Accessibility testing


shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
+ruby:
+  banner_classes = "banner"
+  if local_assigns[:type] && type != :information
```
broder: Do we need to do `type.to_sym` here? Or can we remove `type.to_sym` from some of the conditionals below?

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
...
+  elsif local_assigns[:classes]
+    banner_classes += " #{classes}"
+  end
+  content = capture do
+    local_assigns[:html] || local_assigns[:text]
+  end
+
+div class=banner_classes
+  .banner__message
+    -if local_assigns[:icon_fallback_test]
```
broder:
```suggestion
    -if local_assigns[:iconFallbackText]
```

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
+ruby:
```
broder: Should we call this file `_hmcts_banner` for consistency with the other components?

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
...
+    banner_classes += " #{classes}"
+  end
+  content = capture do
+    local_assigns[:html] || local_assigns[:text]
+  end
+
+div class=banner_classes
+  .banner__message
+    -if local_assigns[:icon_fallback_test]
+      span.hmcts-banner__assistive = local_assigns[:iconFallbackText]
```
broder:
```suggestion
      span.banner__assistive = local_assigns[:iconFallbackText]
```

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
+ruby:
+  banner_classes = "banner"
+  if local_assigns[:type] && type != :information
+    banner_classes += " banner--#{type.to_s}"
+  elsif local_assigns[:classes]
```
jasiekmiko: I'm not sure this wants to be an `elseif` - I can see situations where I want to use both `type` and `classes` params. Is there anything that would make them exclusive?

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
...
+    banner_classes += " #{classes}"
+  end
+  content = capture do
```
broder: Do we need to capture this or can we just inline it below?

shared-web/app/views/components/_banner.html.slim
```diff
@@ -0,0 +1,26 @@
+ruby:
+  banner_classes = "banner"
```
jasiekmiko: if you make it a singleton array that you push extra classes onto, then you don't need the awkward leading whitespaces below. slim natively accepts arrays for attributes, too!

shared-web/app/views/components/_banner.html.slim
jasiekmiko: Was this adapted from `https://github.com/hmcts/frontend/blob/master/src/components/banner/template.njk` like the css? Would be good documenting it.

Also, govuk components can rely on the design system to document the component interface. Do we have an equivalent for this? If not, we should document it ourselves here.

shared-web/app/views/components/_banner.html.slim
jasiekmiko: I think this should say "text", not "test"?

shared-web/app/views/components_gallery/banner.html.slim
```diff
@@ -0,0 +1,32 @@
+h1.govuk-heading-xl Breadcrumbs component
```
jasiekmiko: Nice! Can we add a fallback text example, too?

shared-web/app/views/components_gallery/banner.html.slim
```diff
@@ -0,0 +1,37 @@
...
```
broder: Banner component?

MatthewBurstein: 🤦‍♀️

PR #778, opened 2019-03-19
MSPSDS-1086 Add count of items to case tabs

samlionheart:
## Description

[x]Within a case, add to the products, businesses, and attachment tabs (tabs with countable things)

[x]Count shows regardless of if there are items or not (for no items, it shows "(0)")

[x]Count does not get included in the heading of that tab (h1 seen when you go to a tab) - just on the tab label.

[x]Within a product, add to the 'attachments' tab.

Within a business add to the locations, attachments and contacts tabs
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing


mspsds-web/app/views/businesses/show.html.slim
```diff
@@ -6,8 +6,11 @@
 
 = render "tabs", tabs: [\
   Tab.new("full-detail", "Full detail", lambda { render "businesses/tabs/details" }),
-  Tab.new("locations", "Locations", lambda { render "businesses/tabs/locations" }),
-  Tab.new("contacts", "Contacts", lambda { render "businesses/tabs/contacts" }),
-  Tab.new("attachments", "Attachments", lambda { render "businesses/tabs/attachments" }),
+  Tab.new("locations", "Locations", lambda { render "businesses/tabs/locations" },
```
broder: I wonder if we should just add the count to the title here i.e.
```
Tab.new("locations", "Locations (#{@business.locations.count})", lambda { render "businesses/tabs/locations" }),
```
rather than making the `Tab` model need to know about the count as well. What do you think?

samlionheart: I kept it out of the title so that it wasn't in the heading when you click into the tab, rather than removing the case count at ` h2.govuk-heading-l[class="govuk-!-margin-bottom-0"] = tab.title unless tab.hide_title` in tabs partial

broder: Cosmetics have separated the tab header and the tab title https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/cosmetics-web/app/views/application/_tabs.html.slim#L11. We could move that into shared-web and use that?

PR #781, opened 2019-03-19
MSPSDS-1064 remove old admin role

jasiekmiko:


mspsds-web/test/controllers/investigations_controller_test.rb
```diff
@@ -2,15 +2,15 @@
...
+    mock_out_keycloak_and_notify(last_name: "User_four")
 
     @assignee = User.find_by(last_name: "User_one")
     @non_opss_user = User.find_by(last_name: "User_two")
     set_user_as_non_opss(@non_opss_user)
 
     @investigation_one = investigations(:one)
     @investigation_one.created_at = Time.zone.parse('2014-07-11 21:00')
-    @investigation_one.assignee = User.find_by(last_name: "Admin")
+    @investigation_one.assignee = User.find_by(last_name: "User_four")
```
broder: Can we do better than `find_by` `last_name` everywhere? Is there a reason we can't use the user IDs or something like that?

jasiekmiko: The ids are guids, so don't think that's much nicer at all!
Happy to hear suggestions for better, but here I tried to keep to a like-for-like replacement as much as possible.

broder: That's fine! I think cosmetics have managed to do it in a slightly neater way e.g. https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/cosmetics-web/spec/system/responsible_person_dashboard_spec.rb but that might be because of FactoryBot...

jasiekmiko: Yeah, does seem less messy. Hopefully we can move towards a similar thing over the course of these incremental improvements :)

PR #786, opened 2019-03-20
MSPSDS-997: Edit case summary

kicferk1:
Added a link to 'Edit summary' on case overview page, under the summary. It takes you to a page similar to comment one, populates text area with the current summary, and allows for editing. 

I also refactored the bits of investigation controller that deal with edit actions(assign, visibility, status), because the logic in them was repetitive. This ticket mentions that eventually we will want a lot more edit pages, and when that happens I think part of that logic will be quite easy to extract and share between controllers. 

## Checklist:
### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/policies/investigation_policy.rb
```diff
@@ -23,6 +23,10 @@ def visibility?(user: @user)
     visible_to(user: user, private: true)
   end
 
+  def edit_summary?
+    show?
```
jasiekmiko: Would it be nicer to have all the `<action>?` methods call the common rule (i.e. currently `visible_to`), as opposed to referencing each other?

mspsds-web/app/views/investigations/edit_summary.html.slim
```diff
@@ -0,0 +1,14 @@
+- content_for :page_title, "Update status"
+- content_for :after_header do
+  = link_to "Back to #{@investigation.pretty_description}", @investigation, class: "govuk-back-link"
+= form_with(model: @investigation, method: :put, url: edit_summary_investigation_url(@investigation)) do |form|
```
jasiekmiko: I *think* the `put` method is inferred from modifying an existing object?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -57,59 +57,63 @@ def new
...
+    edit(model_keys: [:description], get_valid_update_hash: get_valid_update_hash,
+         error_message: "Summary can not be empty")
   end
 
-  # GET /cases/1/assign
-  # PUT /cases/1/assign
-  def assign
+private
+
+  def edit(model_keys:, error_message:, get_valid_update_hash:, custom_params: nil)
```
jasiekmiko: Should we just have a single "update" action that the four forms trigger?

mspsds-web/app/views/investigations/edit_summary.html.slim
kicferk1: Actually, it will use `patch`, but reading up on that patch does reflect what we do slightly better than put, so it's probably ok.

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: We originally separated it out because the logic of error handling was getting too complicated. 

Currently the logic for 'which action do I actually verify and re-render if invalid' is hidden under the choice of url we send stuff to. We can move it to controller I guess, but there will be a slight overhead of setting action that `respond_to_update` and `respond_to_invalid_data` take. 

Soo, the answer is 'maybe?'

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -60,55 +60,35 @@ def new
...
-      @investigation.errors.add(model_keys.first, :invalid, message: error_message)
-      respond_to_invalid_data
-      return
-    end
-
-    update_hash.each do |key, value|
-      @investigation.send("#{key}=", value) if value
+    ps = update_params
+    editable_keys.each do |key|
+      @investigation.send("#{key}=", ps[key]) if params.require(:investigation).key?(key)
```
jasiekmiko: Can we do something like `@investigation.update ps` intead of the loop??

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -128,9 +108,21 @@ def set_investigation
...
 
-  def assignee_update_params
+  def update_params
     return {} if params[:investigation].blank?
 
+    handle_assign_params
+    params.require(:investigation).permit(editable_keys)
+  end
+
+  def editable_keys
```
jasiekmiko: I'd be tempted to move this method definition underneath the `handle_assings_params` one, so order of def is same as order of calling - small thing that makes it easier to read and follow! :)

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -128,9 +108,21 @@ def set_investigation
...
+    %i[description is_closed status_rationale is_private visibility_rationale assignable_id]
+  end
+
+  def handle_assign_params
+    relevant_params = params.require(:investigation)
+    return unless relevant_params.key?(:assignable_id) || relevant_params.key?(:select_someone_else)
```
jasiekmiko: Would `relevant_params.key?(:select_someone_else)` ever be true without `relevant_params.key?(:assignable_id)` being true?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -143,12 +135,12 @@ def assignee_update_params
...
   end
 
   def respond_to_update
     respond_to do |format|
-      if @investigation.save
+      if @investigation.valid?(:edit)
+        @investigation.save
         format.html { redirect_to @investigation, notice: "#{@investigation.case_type.titleize} was successfully updated." }
         format.json { render :show, status: :ok, location: @investigation }
       else
```
jasiekmiko: In the line below, we call `@investigation.restore_attributes`. That seems like the user could lose whatever they entered... Doesn't seem right?

mspsds-web/app/models/investigation.rb
```diff
@@ -96,18 +92,11 @@ def as_indexed_json(*)
   end
 
   def assignee
-    begin
-      return User.find(assignable_id) if assignable_type == "User"
-      return Team.find(assignable_id) if assignable_type == "Team"
-    rescue StandardError
-      return nil
-    end
+    (User.where(id: assignable_id) + Team.where(id: assignable_id)).first
```
jasiekmiko: Does `User.find assignable_id` work here? It would return just the singleton, so we could do a more idiomatic
`User.find(assingable_id) || Team.find(assignable_id)`, or something akin to this?

mspsds-web/app/models/investigation.rb
```diff
@@ -96,18 +92,11 @@ def as_indexed_json(*)
...
   end
 
   def assignee=(entity)
-    self.assignable_id = entity.id
-    self.assignable_type = "User" if entity.is_a?(User)
```
jasiekmiko: How come this isn't needed anymore?

mspsds-web/app/models/investigation.rb
```diff
@@ -214,13 +210,13 @@ def create_audit_activity_for_visibility
   end
 
   def create_audit_activity_for_assignee
-    if (saved_changes.key? :assignable_id) || (saved_changes.key? :assignable_type)
+    if (saved_changes.key? :assignable_id) && User.current
```
jasiekmiko: The `User.current` seems like an odd part of the condition... what's the use? (and below)

mspsds-web/db/migrate/20190321122934_populate_assignee_and_description.rb
```diff
@@ -0,0 +1,23 @@
+class PopulateAssigneeAndDescription < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
```
jasiekmiko: What did strong migrations complain about here, was it adding indecies?
@broder do we need to be avoiding `safety_assured` calls now that we're live?

mspsds-web/test/controllers/investigations_controller_test.rb
```diff
@@ -103,16 +99,16 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
...
 
   test "should require description to not be empty" do
-    put edit_summary_investigation_url(@investigation_one), params: {
+    patch edit_summary_investigation_url(@investigation_one), params: {
       investigation: {
         description: ""
       }
     }
-    assert_includes(response.body, "Summary can not be empty")
+    assert_includes(response.body, "blank")
```
jasiekmiko: This assertion has become much less readable now - anything stopping us from checking the full error message?
Also might be worth asserting response type?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -60,55 +60,35 @@ def new
...
-      potential_assignees.first.present? ? { assignee: potential_assignees.first } : {}
-    end
-    edit(model_keys: %i[assignable_id assignable_type], get_valid_update_hash: get_valid_update_hash,
-         custom_params: assignee_update_params,
-         error_message: "Select assignee")
+    edit
   end
 
   # GET /cases/1/status
   # PUT /cases/1/status
```
jasiekmiko: These should say patch now?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,43 +56,33 @@ def new
...
+  def update
+    return if request.get?
+
+    ps = update_params
+    editable_keys.each do |key|
+      @investigation.send("#{key}=", ps[key]) if params.require(:investigation).key?(key)
+    end
+
+    viewed_page = ""
+    viewed_page = "investigations/edit_summary" if params.require(:investigation).key?(:description)
```
jasiekmiko: Hmm, wondering if there's a nicer way to deal with these... Is `request.referrer` one of them?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,43 +56,33 @@ def new
     end
   end
 
-  # GET /cases/1/assign
-  # PUT /cases/1/assign
-  def assign
-    edit
+  def update
```
jasiekmiko: Missing action path comment/annotation

mspsds-web/db/migrate/20190321122934_populate_assignee_and_description.rb
broder: We should be trying to avoid them, yes. It'll matter more when we're doing zero-downtime deploys though - some of the things it recommends are to prevent the database locking for a few seconds.

mspsds-web/app/views/investigations/edit_summary.html.slim
```diff
@@ -1,7 +1,7 @@
 - content_for :page_title, "Update status"
 - content_for :after_header do
   = link_to "Back to #{@investigation.pretty_description}", @investigation, class: "govuk-back-link"
-= form_with(model: @investigation, url: edit_summary_investigation_url(@investigation)) do |form|
+= form_with(model: @investigation, url: investigation_url(@investigation)) do |form|
```
jasiekmiko: I think `form_with(model: @investigation, url: investigation_url(@investigation))` is just equivalent to  `form_with(model: @investigation`?

mspsds-web/app/models/investigation.rb
kicferk1: Migration we use to populate assignee and description(to avoid validation problems on update later on) will trigger after_save callback on model, which will trigger this method, which will trigger creating activity, which will trigger an email to the new assignee, which we probably don't want.

mspsds-web/db/migrate/20190321122934_populate_assignee_and_description.rb
kicferk1: Pretty sure it complained cause removing a column isn't very safe in general

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -110,41 +100,26 @@ def set_investigation
...
-                                             when "someone_else"
-                                               params[:investigation][:select_someone_else]
-                                             else
-                                               params[:investigation][:assignable_id]
-                                             end
+    %i[description is_closed status_rationale is_private visibility_rationale]
   end
 
-  def respond_to_update
+  def respond_to_update(viewed_page)
```
jasiekmiko: This could be inlined now, no?

mspsds-web/app/models/investigation.rb
jasiekmiko: Yeah, makes sense - but not sure we want to keep this code change in long term - would it make sense to put a comment here and remove this in next release?

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: As in putting a `request.referrer.include? "status"` in a conditional? Maybe, but with that check we probably break any json call validation :(

mspsds-web/app/models/investigation.rb
kicferk1: It would only be needed if we had two records with different type and same id that can be assigned to a case. Keycloak groups and users use uuids, so it's not going to happen. 

However, seeing as assigning got moved to its own cosy controller I can bring it back without breaking anything.

mspsds-web/db/migrate/20190321122934_populate_assignee_and_description.rb
jasiekmiko: Ah, not even spotted the removal!
How come? Is it linked to the work or tagging along? If so, worth own migration?

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: Yes, it could be. 
Luckily assigning got its own controller now, so it's not important anymore :)

mspsds-web/app/controllers/investigations_controller.rb
jasiekmiko: More like `viewed_page = some_func(request.referrer)`, with whatever transformation necessary provided by `some_func`. This only applies to html responses, so not sure how we'd break json here?

kicferk1: Actually, json doesn't really need it to begin with, so maybe

jasiekmiko: Alternatively maybe think about setting the view name on the submit button as value? You should get that info back via `params.commit` I think?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,20 +56,29 @@ def new
...
-      @investigation.send("#{key}=", ps[key]) if params.require(:investigation).key?(key)
-    end
-
-    viewed_page = ""
-    viewed_page = "investigations/edit_summary" if params.require(:investigation).key?(:description)
-    viewed_page = "investigations/status" if params.require(:investigation).key?(:is_closed)
-    viewed_page = "investigations/visibility" if params.require(:investigation).key?(:is_private)
+    @investigation.update(update_params)
+    is_valid = @investigation.valid?(:edit)
+    saved = is_valid ? @investigation.save : false
```
jasiekmiko: From https://guides.rubyonrails.org/active_record_validations.html#on `valid` runs on `save` with `:update` passed in - would that work here, or does it clash with anything? If it does then we're at a much nicer:
```
@investigation.update(update_params)
saved = @investiagion.save
```
if it doesn't then we can at least abbreviate the last line to:
```
saved = is_valid && @investigation.save
```
?

mspsds-web/test/controllers/investigations_controller_test.rb
```diff
@@ -90,8 +90,8 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
       investigation: {
         description: ""
       }
-    }
-    assert_includes(response.body, "blank")
+    }, headers: { "HTTP_REFERER": '/edit_summary' }
+    assert_includes(response.body, "Description can&#39;t be blank")
```
jasiekmiko: Oueeh that's ugly :S is there some sort of escaping we can run on `response.body` to make that better?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -62,20 +63,18 @@ def update
...
         format.html {
           redirect_to @investigation, flash: {
             success: "#{@investigation.case_type.titleize} was successfully updated."
           }
         }
         format.json { render :show, status: :ok, location: @investigation }
       else
-        viewed_page = request.referer.split('/').last
-        format.html { render viewed_page }
+        @investigation.restore_attributes
```
jasiekmiko: Still a bit uneasy about this line... If the user enters a new description and validation fails, do they lose what they entered because of it?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,39 +57,36 @@ def new
     end
   end
 
-  # GET /cases/1/status
-  # PUT /cases/1/status
-  def status
+  # PATCH /cases/1
+  # PATCH /cases/1.json
+  def update
     return if request.get?
```
broder: Can we GET `update`?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,39 +57,36 @@ def new
     end
   end
 
-  # GET /cases/1/status
-  # PUT /cases/1/status
-  def status
+  # PATCH /cases/1
```
broder: What was the reasoning behind `PATCH`? `PUT` is generally what rails suggests for updates.

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,39 +57,36 @@ def new
...
+    respond_to do |format|
+      if @investigation.save
+        format.html {
+          redirect_to @investigation, flash: {
+            success: "#{@investigation.case_type.titleize} was successfully updated."
+          }
+        }
+        format.json { render :show, status: :ok, location: @investigation }
+      else
+        session[:viewed_page] ||= request.referer.split('/').last
```
broder: I would guess that this is a security risk. It's very close to being able to `render` any file on the server (similar to https://github.com/mpgn/CVE-2019-5418) - spoofing your referer is quite straightforward.

mspsds-web/app/models/investigation.rb
```diff
@@ -9,21 +9,17 @@ class Investigation < ApplicationRecord
   attr_accessor :assignee_rationale
 
   before_validation { trim_line_endings(:user_title, :description, :non_compliant_reason, :hazard_description) }
-  validates :user_title, presence: true, on: :enquiry_details
-  validates :description, presence: true, on: %i[allegation_details enquiry_details]
-  validates :hazard_type, presence: true, on: :allegation_details
-  validates :product_category, presence: true, on: :allegation_details
-  validates :hazard_description, presence: true, on: :unsafe
-  validates :hazard_type, presence: true, on: :unsafe
-  validates :non_compliant_reason, presence: true, on: :non_compliant
```
broder: Have we lost this validation? Do we need it?

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: No, we can't, it returns 404

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: Patch is what the form auto-generates when the model object is a record that already exists in database. 

A [https://stackoverflow.com/questions/21660791/what-is-the-main-difference-between-patch-and-put-request](stack overflow page) suggests put is 'sending a changed copy of object' and patch is 'sending the fields to change', and we are definitely doing the latter.

broder: Fair enough

mspsds-web/app/models/investigation.rb
kicferk1: I moved it to Allegation, because allegtion_details only ever runs on Allegation, and ts flow(the other 3) creates Allegations as well.

broder: Has the `on:` condition on `Investigation::Allegation` intentionally changed?

kicferk1: No, it was an oversight

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,39 +57,34 @@ def new
...
-
-    ps = status_update_params
-    if ps[:is_closed].blank?
-      @investigation.errors.add(:status, :invalid, message: "Status should be closed or open")
-      respond_to_invalid_data(:status)
-      return
+  # PATCH /cases/1
+  # PATCH /cases/1.json
+  def update
+    @investigation.update(update_params)
```
broder: Does `.update` also save to the database?

mspsds-web/app/controllers/investigations_controller.rb
```diff
@@ -56,39 +57,34 @@ def new
...
+    respond_to do |format|
+      if @investigation.save
+        format.html {
+          redirect_to @investigation, flash: {
+            success: "#{@investigation.case_type.titleize} was successfully updated."
+          }
+        }
+        format.json { render :show, status: :ok, location: @investigation }
+      else
+        session[:viewed_page] ||= request.referer&.match(/cases\/\d+-\d+\/(edit_summary|status|visibility)/)&.captures&.first
```
broder: This feels a little surprising. I wonder if it is just simpler for now to keep the separate actions for summary / status / visibility...

mspsds-web/app/controllers/investigations_controller.rb
kicferk1: Yes it does

PR #788, opened 2019-03-20
COSBETA-29: Fix Elasticsearch indexing and edit page indentation

raamSoftwire:
These fixes cover an indentation issue from COSBETA-29 but also some work on Elasticsearch indexing. These issues were raised by Nasir and are described by tickets 173 and 176.

This is a temporary fix that relies on not being able to edit/update submitted notifications. This will need to be revisited if that ever changes. Or if we want to edit RP details.

[COSBETA-173](https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24/backlog?selectedIssue=COSBETA-173)

[COSBETA-176](https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24/backlog?selectedIssue=COSBETA-176)


shared-web/app/models/shared/web/concerns/searchable.rb
```diff
@@ -7,7 +7,10 @@ module Searchable
 
         included do
           include Elasticsearch::Model
-          include Elasticsearch::Model::Callbacks
```
kicferk1: Do we know which callback caused a crash?

tom-bell-softwire: From the `elasticsearch-model` gem source code:
```ruby
module Callbacks

  # Handle index updates (creating, updating or deleting documents)
  # when the model changes, by hooking into the lifecycle
  #
  # @see http://guides.rubyonrails.org/active_record_callbacks.html
  #
  def self.included(base)
    base.class_eval do
      after_commit lambda { __elasticsearch__.index_document  },  on: :create
      after_commit lambda { __elasticsearch__.update_document },  on: :update
      after_commit lambda { __elasticsearch__.delete_document },  on: :destroy
    end
  end
end
```

My guess is that it's the `update` callback being fired, when the record doesn't already exist on Elasticsearch (because we only index notifications once they've been submitted).

PR #791, opened 2019-03-20
COSBETA-134: Improve bulk upload interface for users

raamSoftwire:
This is a WIP branch to integrate DropzoneJS for the bulk upload page.


raamSoftwire:
- Loading config file
  - Dropzone says it has attached to the element
  - But the options are not being loaded
- Error handling
- Accepting html over json - request header
  - In the request headers we see "application/json"
  - If we create a new.json, then the error goes away
  - How do we change this to "text/html"?
- Using correct template

PR #797, opened 2019-03-20
Update index.html.slim

adrievdluijt:
Updated 'register' to 'notify' where applicable.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


tom-bell-softwire:
Covered as part of #811.

PR #799, opened 2019-03-21
COSBETA-184 Add pages to route users to bulk or manual journey

DWRendell:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
@@ -20,31 +20,27 @@ class ResponsiblePersons::AddNotificationWizardController < ApplicationControlle
 
     case step
     when :have_products_been_notified_in_eu
-      case answer
-      when "yes"
+      if answer == "yes"
         redirect_to wizard_path(:do_you_have_files_from_eu_notification)
-      when "no"
+      else
         redirect_to wizard_path(:will_products_be_notified_in_eu)
       end
     when :will_products_be_notified_in_eu
-      case answer
-      when "yes"
+      if answer == "yes"
         redirect_to wizard_path(:register_on_eu_system)
-      when "no"
+      else
         redirect_to wizard_path(:was_product_on_sale_before_eu_exit)
       end
     when :do_you_have_files_from_eu_notification
-      case answer
-      when "yes"
+      if answer == "yes"
         redirect_to bulk_upload_path
-      when "no"
+      else
         redirect_to manual_journey_path(notified_before_eu_exit: false)
       end
     when :was_product_on_sale_before_eu_exit
-      case answer
-      when "yes"
+      if answer == "yes"
         redirect_to manual_journey_path(notified_before_eu_exit: true)
-      when "no"
+      else
         redirect_to manual_journey_path(notified_before_eu_exit: false)
       end
     end
--- a/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
@@ -12,26 +12,42 @@ class ResponsiblePersons::AddNotificationWizardController < ApplicationControlle
   end
 
   def update
-    map = {
-        have_products_been_notified_in_eu: {
-            "yes" => wizard_path(:do_you_have_files_from_eu_notification),
-            "no" => wizard_path(:will_products_be_notified_in_eu)
-        },
-        will_products_be_notified_in_eu: {
-            "yes" => wizard_path(:register_on_eu_system),
-            "no" => wizard_path(:was_product_on_sale_before_eu_exit)
-        },
-        do_you_have_files_from_eu_notification: {
-            "yes" => bulk_upload_path,
-            "no" => manual_journey_path(notified_before_eu_exit: false)
-        },
-        was_product_on_sale_before_eu_exit: {
-            "yes" => manual_journey_path(notified_before_eu_exit: true),
-            "no" => manual_journey_path(notified_before_eu_exit: false)
-        }
-    }
+    answer = params[:answer]
+    if answer != "yes" && answer != "no"
+      @error_text = "Please select an answer"
+      return render step
+    end
 
-    redirect_to map[step][params[:answer]]
+    case step
+    when :have_products_been_notified_in_eu
+      case answer
+      when "yes"
+        redirect_to wizard_path(:do_you_have_files_from_eu_notification)
+      when "no"
+        redirect_to wizard_path(:will_products_be_notified_in_eu)
+      end
+    when :will_products_be_notified_in_eu
+      case answer
+      when "yes"
+        redirect_to wizard_path(:register_on_eu_system)
+      when "no"
+        redirect_to wizard_path(:was_product_on_sale_before_eu_exit)
+      end
+    when :do_you_have_files_from_eu_notification
+      case answer
+      when "yes"
+        redirect_to bulk_upload_path
+      when "no"
+        redirect_to manual_journey_path(notified_before_eu_exit: false)
+      end
+    when :was_product_on_sale_before_eu_exit
+      case answer
+      when "yes"
+        redirect_to manual_journey_path(notified_before_eu_exit: true)
+      when "no"
+        redirect_to manual_journey_path(notified_before_eu_exit: false)
+      end
+    end
   end
 
   def new
--- a/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
@@ -1,0 +1,54 @@
+class ResponsiblePersons::AddNotificationWizardController < ApplicationController
+  include Wicked::Wizard
+
+  before_action :set_responsible_person, only: %i[update show]
+
+  steps :have_products_been_notified_in_eu, :will_products_be_notified_in_eu,
+      :do_you_have_files_from_eu_notification, :was_product_on_sale_before_eu_exit,
+      :register_on_eu_system
+
+  def show
+    render_wizard
+  end
+
+  def update
+    map = {
+        have_products_been_notified_in_eu: {
+            "yes" => wizard_path(:do_you_have_files_from_eu_notification),
+            "no" => wizard_path(:will_products_be_notified_in_eu)
+        },
+        will_products_be_notified_in_eu: {
+            "yes" => wizard_path(:register_on_eu_system),
+            "no" => wizard_path(:was_product_on_sale_before_eu_exit)
+        },
+        do_you_have_files_from_eu_notification: {
+            "yes" => bulk_upload_path,
+            "no" => manual_journey_path(notified_before_eu_exit: false)
+        },
+        was_product_on_sale_before_eu_exit: {
+            "yes" => manual_journey_path(notified_before_eu_exit: true),
+            "no" => manual_journey_path(notified_before_eu_exit: false)
+        }
+    }
+
+    redirect_to map[step][params[:answer]]
+  end
+
+  def new
+    redirect_to wizard_path(steps.first)
+  end
+
+private
+
+  def set_responsible_person
+    @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
+  end
+
+  def manual_journey_path(notified_before_eu_exit)
+    new_responsible_person_notification_path(@responsible_person, notified_before_eu_exit: notified_before_eu_exit)
+  end
+
+  def bulk_upload_path
+    new_responsible_person_notification_file_path(@responsible_person)
+  end
+end
--- a/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
@@ -45,7 +45,7 @@ private
   end
 
   def manual_journey_path(notified_before_eu_exit)
-    new_responsible_person_notification_path(@responsible_person, notified_before_eu_exit: notified_before_eu_exit)
+    new_responsible_person_notification_path(@responsible_person, notified_before_eu_exit)
   end
 
   def bulk_upload_path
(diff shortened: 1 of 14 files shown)


DWRendell:
Ah, I've accidentally included the commits from COSBETA-167 in here :( Please ignore them until that's merged

cosmetics-web/app/views/responsible_persons/add_notification_wizard/do_you_have_files_from_eu_notification.html.slim
```diff
@@ -0,0 +1,20 @@
...
+- content_for :after_header do
+    = link_to "Back", wizard_path(:have_products_been_notified_in_eu), class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with url: wizard_path, method: :put do |form|
+      h1.govuk-heading-xl
+        | Do you have the files from your EU notification?
+      p CPNP allows you to export ZIP files containing XML files for each notification.
+      p
+        ' To use your EU notification files, you must
```
raamSoftwire: Is there a neater way in slim to leave an empty line?

cosmetics-web/app/views/responsible_persons/add_notification_wizard/do_you_have_files_from_eu_notification.html.slim
```diff
@@ -0,0 +1,20 @@
+- content_for :page_title, "Do you have the files from your EU notification?"
+- content_for :after_header do
+    = link_to "Back", wizard_path(:have_products_been_notified_in_eu), class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with url: wizard_path, method: :put do |form|
+      h1.govuk-heading-xl
+        | Do you have the files from your EU notification?
+      p CPNP allows you to export ZIP files containing XML files for each notification.
```
raamSoftwire: Should the text start on a new line?

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -26,7 +26,13 @@ def index
   def show; end
 
   def new
-    @notification = Notification.create(responsible_person: @responsible_person)
+    was_notified_before_eu_exit = (params.has_key? "notified_before_eu_exit") && (params["notified_before_eu_exit"] == "true")
```
broder: Do we need to check for `params` having the key? Wouldn't it be `nil` otherwise and then not `== "true"`?

cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+    answer = params[:answer]
+    if answer != "yes" && answer != "no"
+      @error_text = "Please select an answer"
+      return render step
+    end
+
+    case step
+    when :have_products_been_notified_in_eu
+      case answer
+      when "yes"
```
broder: Would `if / else` be neater for checking if the answer is yes or no?

cosmetics-web/app/assets/application/stylesheets/responsible_persons.scss
```diff
@@ -17,16 +17,15 @@
...
-  border: none;
   padding: 0;
 
   // optional
   font-family: arial, sans-serif; //input has OS specific font-family
   font: inherit;
   text-decoration: underline;
   color: $govuk-link-colour;
-  cursor: pointer;
+  @include remove-default-button-styling;
```
broder: I think you've got some changes from #758 but hopefully that won't cause too many issues.

cosmetics-web/app/views/responsible_persons/add_notification_wizard/do_you_have_files_from_eu_notification.html.slim
```diff
@@ -0,0 +1,20 @@
...
+      p
+        ' To use your EU notification files, you must
+        strong
+          ' export
+        ' your data as ZIP files for all cosmetic products.
```
broder: I think you could use `'  your data...` here to avoid needing `'` within the `strong` tag. The double space at the start gives the rendered HTML a leading space.

cosmetics-web/app/views/responsible_persons/add_notification_wizard/have_products_been_notified_in_eu.html.slim
```diff
@@ -0,0 +1,14 @@
+- content_for :page_title, "Have the cosmetic products been notified in the EU?"
+- content_for :after_header do
+  = link_to "Back", responsible_person_notifications_path(@responsible_person), class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with url: wizard_path, method: :put do |form|
+      h1.govuk-heading-xl
+        | Have the cosmetic products been notified in the EU?
+      = render "components/govuk_radios",
```
broder: I _think_ you'll want to pass the heading above into the component to make it accessible (i.e. it reads out the question when you navigate to the input) https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/mspsds-web/app/views/investigations/ts_investigations/has_corrective_action.html.slim#L7

cosmetics-web/app/views/responsible_persons/add_notification_wizard/register_on_eu_system.html.slim
```diff
@@ -0,0 +1,19 @@
+- content_for :page_title, \
+  "To make UK cosmetic product notification simpler, notify your cosmetic products on CPNP first"
```
broder: Catchy.

cosmetics-web/app/views/responsible_persons/add_notification_wizard/register_on_eu_system.html.slim
```diff
@@ -0,0 +1,19 @@
...
+- content_for :after_header do
+    = link_to "Back", wizard_path(:will_products_be_notified_in_eu), class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+        | To make UK cosmetic product notification simpler, notify your cosmetic products on CPNP first
+    p
+      | Once the cosmetic products have been registered in the EU, you can use the notification file from the CPNP to
```
broder: Does something like 
```
' Once the cosmetic products have been registered in the EU, you can use the notification file from the CPNP to
  save time. Once you have the cosmetic product notification file, you can
```
have the same effect?

cosmetics-web/app/views/responsible_persons/add_notification_wizard/was_product_on_sale_before_eu_exit.html.slim
```diff
@@ -0,0 +1,14 @@
...
+    = link_to "Back", wizard_path(:will_products_be_notified_in_eu), class: "govuk-back-link"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    = form_with url: wizard_path, method: :put do |form|
+      h1.govuk-heading-xl
+        | Was this product on sale before the UK left the European Union?
+      = render "components/govuk_radios",
+          name: :answer,
+          errorMessage: @error_text.present? && { text: @error_text },
+          items: [{ text: "Yes", value: "yes", id: "answer_yes" },
```
broder: I think you can leave `id` out here and the component will work it out from the `:name` and `value`

cosmetics-web/db/migrate/20190320143555_add_was_notified_before_eu_exit_to_notifications.rb
```diff
@@ -0,0 +1,9 @@
+class AddWasNotifiedBeforeEuExitToNotifications < ActiveRecord::Migration[5.2]
+  def change
+    # safety_assured required to prevent warnings about adding a column with a
+    # non null default rewriting the entire table
+    safety_assured {
```
broder: Not a problem yet, but we won't want to do this if / when we have lots of data and zero-downtime deployments.

cosmetics-web/spec/controllers/responsible_persons/add_notification_wizard_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  describe "GET #show" do
+    it "sets the responsible person from the path parameters" do
+      get(:show, params: { responsible_person_id: responsible_person.id, id: "have_products_been_notified_in_eu" })
+      expect(assigns(:responsible_person)).to eq(responsible_person)
+    end
+  end
+
+  describe "PUT #update" do
```
broder: Should we also test the case where you don't select an option?

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
DWRendell: Ah OK, I'll change that

cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
DWRendell: Yes, that does sound neater

cosmetics-web/app/assets/application/stylesheets/responsible_persons.scss
DWRendell: I had a bit of a mixup, I'll rebase onto master and try and clean it up

cosmetics-web/app/views/responsible_persons/add_notification_wizard/have_products_been_notified_in_eu.html.slim
DWRendell: Oh yeah, I'd been meaning to change this. Do you know if there's any way to do this for the pages where there are explanatory paragraphs between the heading and the radios using the component glue? I'm guessing I'll have to hand roll it by changing the heading to a label and making it for the right element?

cosmetics-web/app/views/responsible_persons/add_notification_wizard/do_you_have_files_from_eu_notification.html.slim
DWRendell: Done

cosmetics-web/app/views/responsible_persons/add_notification_wizard/register_on_eu_system.html.slim
DWRendell: Yes that looks fine, thanks

cosmetics-web/app/views/responsible_persons/add_notification_wizard/have_products_been_notified_in_eu.html.slim
broder: Does https://design-system.service.gov.uk/components/radios/ help? Otherwise, @jasiekmiko might have an idea?

DWRendell: Ah, I think I could capture the html and pass it in as the hint?

jasiekmiko: If I'm understanding intentions well enough here, I think you're looking to use `hint` as the explanatory paragraphs here, for this effect:
https://design-system.service.gov.uk/components/radios/hint/index.html
If so, then this is how you'd get it: https://design-system.service.gov.uk/components/radios/#radio-items-with-hint-example-nunjucks

If you need multiple paragraphs in there, then I'd try the `hint: { html: }}` form.

Happy to help in person if that's easier :)

cosmetics-web/app/views/responsible_persons/add_notification_wizard/was_product_on_sale_before_eu_exit.html.slim
DWRendell: Yep looks like that works

cosmetics-web/db/migrate/20190320143555_add_was_notified_before_eu_exit_to_notifications.rb
DWRendell: OK, will note for the future

cosmetics-web/spec/controllers/responsible_persons/add_notification_wizard_controller_spec.rb
DWRendell: Good point

cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def new
+    redirect_to wizard_path(steps.first)
+  end
+
+private
+
+  def set_responsible_person
+    @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
```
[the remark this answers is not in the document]
DWRendell: Good point, though actually we're only using the responsible person to route the user to different controllers. Maybe a better solution would be to just not fetch the responsible person from the DB at all and just use the ID from the URL for routing?

cosmetics-web/app/views/responsible_persons/add_notification_wizard/do_you_have_files_from_eu_notification.html.slim
DWRendell: I think it's fine to put shortish chunks of text inline like this, slim lint doesn't seem to mind at least. Can change it if that's the convention though.

cosmetics-web/app/controllers/responsible_persons/add_notification_wizard_controller.rb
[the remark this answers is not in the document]
tom-bell-softwire: I guess it depends on whether we want to show these pages to people who aren't a member of the given RP (or any RP). While it shouldn't allow users to do anything malicious, it feels like we shouldn't even allow the page to be viewed if you go to one of these pages for an RP account you don't belong to.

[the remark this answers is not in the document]
DWRendell: OK, I'll add the authorisation

PR #800, opened 2019-03-21
MSPSDS-976 fix confidential logic

MatthewBurstein:
Fixes GDPR logic for displaying attachments in investigation attachments tab.
Adds logic for  allowing GDPR sensitive information to be visible to assignees and their teams.

## Checklist:


mspsds-web/test/system/record_phonecall_correspondence_test.rb
```diff
@@ -87,7 +92,9 @@ class RecordPhoneCallCorrespondenceTest < ApplicationSystemTestCase
     assert_current_path(/cases\/\d+/)
   end
 
-  test "conceals information on phonecalls with customer info" do
+  test "conceals information from other organisations on phonecalls with customer info" do
+    other_org_user = User.find_by last_name: "Ts_user"
+    @investigation.source.update(user_id: other_org_user.id)
```
broder: We could use `set_investigation_source!` here.

PR #801, opened 2019-03-21
COSBETA-209: Collect non legislated data from CPNP

raamSoftwire:
This ticket extracts extra info from the XML files and saves it into the DB.

The new fields are

- For children under 3
- Still on the market
- Physical form
- Special applicator
- Other information for poison centres

For special applicator, if ‘other` is selected a `Please suggest an applicator / packaging denomination‌` text field appears, but this information is not stored in the exported XML file


cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -26,6 +26,7 @@ def metadata
 
   def create_notification_from_file
     begin
+      # rubocop:disable Metrics/BlockLength
```
raamSoftwire: should this go here, or around the entire file?

cosmetics-web/db/migrate/20190320152131_add_details_to_notifications.rb
```diff
@@ -0,0 +1,10 @@
+class AddDetailsToNotifications < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
```
raamSoftwire: Rubocop throws a BulkChangeTable error if you have multiple `add_column` lines. But then StrongMigrations does not support inspecting what happens inside a change_table block. So it asks us to wrap the block with a `safety_assured` command, if we're happy that it is indeed safe.

cosmetics-web/app/analyzers/read_data_analyzer.rb
broder: Fine to go here.

cosmetics-web/app/models/concerns/cpnp_notification_physical_forms.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  PHYSICAL_FORM_ID = {
+      1 => :solid_or_pressed_powder,
+      2 => :loose_powder,
+      3 => :cream_or_paste,
+      4 => :liquid,
+      5 => :foam,
+      6 => :spray,
+      -2 => :other_physical_form
```
broder: `-2` wat.

cosmetics-web/db/migrate/20190320152131_add_details_to_notifications.rb
broder: This is fine whilst we're not live.

cosmetics-web/app/analyzers/read_data_analyzer.rb
raamSoftwire: Cool, thanks

cosmetics-web/app/models/concerns/cpnp_notification_physical_forms.rb
raamSoftwire: I know right 🤪 
image

raamSoftwire:
I'm working on a separate ticket to surface these values. COSBETA-212

PR #803, opened 2019-03-21
MSPSDS-575 fix ie11 validation

MatthewBurstein:
Override default validation message when alert subject or body are blank.
Implement textinput component in alert form.

## Checklist:


mspsds-web/app/views/investigations/alerts/compose.html.slim
```diff
@@ -10,19 +10,22 @@
         p.govuk-caption-xl
           = @investigation.pretty_description
         | Compose new alert
+      - subject_label = capture do
+        = form.label :description, "Alert subject",
```
broder: Does this put a `label` within the `label` defined in https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/shared-web/app/views/components/_govuk_label.html.slim#L6?

MatthewBurstein: Good spot - yes it does. Unfortunately, passing the legend classes through to the label as planned doesn't work because the label has `.govuk-label` by default which overrides some of the css in the legend classes. I'm not entirely sure how to fix this without creating differences between the govuk label component api and our own. Any thoughts? @jasiekmiko

mspsds-web/app/models/alert.rb
```diff
@@ -6,8 +6,10 @@ class Alert < ApplicationRecord
 
   has_one :source, as: :sourceable, dependent: :destroy
 
-  validates :summary, presence: true
-  validates :description, presence: true
+  validates_presence_of :summary, message: "Enter the alert subject"
+  validates_presence_of :description, message: "Enter the alert subject"
```
jasiekmiko: description?

mspsds-web/app/views/investigations/alerts/compose.html.slim
jasiekmiko: we shouldn't be using fieldset classes here, becuase it's not the fieldset.
I think if you set govuk-label--m you'll get the same effect.

Fielsets should be used for multiple elements consituting a single value (e.g. how we collect dates using 3 text fields). Don't think we wan them here, so let's remove both the fieldset tags below and find classes that make sense on labels :) Does that make sense?

mspsds-web/app/views/investigations/alerts/compose.html.slim
```diff
@@ -10,19 +10,22 @@
...
+        = render "form_components/govuk_input",
+                form: form,
+                key: :summary,
+                label: { html: subject_label },
+                hint: { text: subject_hint_text },
+                value: (@alert.summary.presence || @default_summary)
 
       - description_label = capture do
         = form.label :description, "Alert content",
                 class: "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
```
jasiekmiko: ditto

mspsds-web/app/views/investigations/alerts/compose.html.slim
MatthewBurstein: Yes this makes perfect sense thanks Jan :)

mspsds-web/app/models/alert.rb
MatthewBurstein: whoops 😳

mspsds-web/app/views/investigations/alerts/compose.html.slim
```diff
@@ -10,33 +10,24 @@
         p.govuk-caption-xl
```
jasiekmiko: This also wants to be a `span`, not a `p` - see https://design-system.service.gov.uk/styles/typography/#headings-with-captions-typography-example-open-html
Would you mind fixing that while you're here?? :)

mspsds-web/app/views/investigations/alerts/compose.html.slim
```diff
@@ -10,33 +10,24 @@
...
-                class: "govuk-fieldset__legend govuk-fieldset__legend--m govuk-!-margin-bottom-2"
-      - subject_hint_text = "This will be used in the email subject line."
-      fieldset.govuk-fieldset
-        = render "form_components/govuk_input",
-                form: form,
-                key: :summary,
-                label: { html: subject_label },
-                hint: { text: subject_hint_text },
-                value: (@alert.summary.presence || @default_summary)
+      = render "form_components/govuk_input",
```
jasiekmiko: Much nicer, thanks!

mspsds-web/app/models/alert.rb
```diff
@@ -21,4 +26,25 @@ def send_alert_email
   def create_audit_activity
     AuditActivity::Alert::Add.from self
   end
+
+  def summary_validation
+    if summary.empty? || summary == default_summary
+      p "in here"
```
jasiekmiko: blip

MatthewBurstein: 😞

PR #804, opened 2019-03-21
COSBETA-56 Update service name and update phase banner to BETA

DWRendell:
I need to add some changes to the keycloak-theme-govuk submodule, would anyone be able to talk me through how best to add those?

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


keycloak/configuration/initial-setup.json
```diff
@@ -2268,7 +2268,7 @@
   }, {
     "id" : "121ec424-71bc-4403-829a-44cf4fcb05c5",
     "clientId" : "cosmetics-app",
-    "name" : "Notify Cosmetics",
+    "name" : "Submit cosmetic product notifications",
```
broder: Do you want to put something in the "next release checklist" in the CHANGELOG or are you happy to remember to do this on all environments?

DWRendell: I already have some keycloak changes to make for another ticket so hopefully should remember to do this, I'll make a note

PR #806, opened 2019-03-21
COSBETA-212: Surface non legislated data from cpnp

raamSoftwire:
Added new rows to product and component tables on check your answers page.

- Still on the market
- Intended for age 3 or under?
- Physical form
- Special applicator
- Important information for poison centres


cosmetics-web/app/helpers/notification_properties_helper.rb
```diff
@@ -438,4 +446,22 @@ def get_exposure_condition_name(exposure_condition)
       rinse_off: "Rinse off",
       leave_on: "Leave on"
   }.freeze
+
+  PHYSICAL_FORM = {
+      solid_or_pressed_powder: "Solid/pressed powder",
```
broder: Do you know why we need this in addition to the `enum` here https://github.com/UKGovernmentBEIS/beis-mspsds/pull/801/files#diff-c7dc3f769fde3501115fab8ba847af58R6?

cosmetics-web/app/views/notifications/_component_details.html.slim
```diff
@@ -74,6 +74,21 @@ tr.govuk-table__row
...
+  td.govuk-table__cell
+    = get_physical_form_name(component.physical_form)
+tr.govuk-table__row
+  th.govuk-table__header[scope="row"]
+    | Special applicator
+  td.govuk-table__cell
+    = get_special_applicator_name(component.special_applicator)
+tr.govuk-table__row
+  th.govuk-table__header[scope="row"]
+    | Important information for poison centres
```
broder: Will it be obvious to poison centres that this is `acute_poisoning_info`?

cosmetics-web/app/helpers/notification_properties_helper.rb
raamSoftwire: No, not really. I was following the pattern for the other fields. But I agree it's not very DRY

cosmetics-web/app/views/notifications/_component_details.html.slim
raamSoftwire: Good point.

broder: Are you planning to update this or just feed it back to Ralph / Adrie?

raamSoftwire: Sorry, I updated it but didn't push. I have also fed it back to Adrie but no answer just yet.

PR #807, opened 2019-03-21
MSPSDS-1091 ts flow confirmation page

MatthewBurstein:
## Checklist:

### General testing
- [X] Test without javascript
- [X] Test on small screen
### Accessibility testing
- [X] Works keyboard only
- [X] Disable css - does content make sense and appear in a logical order?


mspsds-web/app/views/investigations/created.html.slim
```diff
@@ -0,0 +1,28 @@
+- content_for :page_title, "Investigation created"
```
broder: Should this be "Case created"?

PR #814, opened 2019-03-25
MSPSDS-1080 Add skip links to repeatable ts-flow pages

jasiekmiko:
@raamSoftwire as you can see in the first commit, I've moved a piece of link-styling from cosmetics to shared, let me know if you see any issues with that!

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -288,8 +294,15 @@ def store_why_reporting
   end
 
   def store_business
+    current_business_type = params.require(:business)[:business_type]
+    # Ideally we'd use the "value" of the button here, separate from the literally displayed text, but due to
+    # differences in how this is handled between ie8 and normal browsers, that's not practical
+    if params[:commit] == "Skip this page"
```
kicferk1: `if @skip`? It looks like it isn't set here, but perhaps it should be?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -74,6 +74,8 @@ def create
 
   # PATCH/PUT /xxx
   def update
+    return redirect_to next_wizard_path if @skip
```
MatthewBurstein: Would it be easier to put the redirect in a `before_action` at the top of the list of `before_action`s? It feels like we are duplicating this with `return if @skip` in a few places.

jasiekmiko: I'm pretty sure that having a method which calls `redirect_to ...` as one of the hooks doesn't stop the latter from running, so I would have to make those latter hooks conditional on `@skip` - doable, but I think this is a slightly less controversial way of getting there?

mspsds-web/app/controllers/investigations/ts_investigations_controller.rb
jasiekmiko: Yeah, the awkwardness of setting it probably warrants it to be factored out! :)

PR #815, opened 2019-03-25
MSPSDS-1031: Protected correspondence visibility issues

kicferk1:
Use the source of the actual object(email, phone call, document) to determine whether it can be shown. Previously the source of case was used, which meant a creator could always see anything, and if someone from a different organisation uploaded sensitive information, they couldn't see it. 

This required changing tests that relied on that quirk to set up conditions necessary to test gdpr. 

## Checklist:


mspsds-web/app/models/complainant.rb
```diff
@@ -15,6 +15,6 @@ def contains_personal_data?
   end
 
   def can_be_displayed?
-    investigation.can_display_child_object? contains_personal_data?
+    investigation.can_display_child_object?(contains_personal_data?, investigation.source)
```
jasiekmiko: It seems weird that we're passing an attribute of an investigation to a method on that same investigation?

mspsds-web/app/models/investigation.rb
```diff
@@ -203,10 +203,11 @@ def add_pretty_id
     self.pretty_id = "#{created_at.strftime('%y%m')}-%04d" % (cases_before + 1)
   end
 
-  def can_display_child_object?(child_has_gdpr_sensitive_data)
-    return true if self.source&.is_a? ReportSource
+  def can_display_child_object?(child_has_gdpr_sensitive_data, child_source)
```
jasiekmiko: This method has a bunch paths now and it's not very clear what the overall goal/approach is. Is there some documentation around this area that Andrew worked on last week? can we link to it?

mspsds-web/app/models/investigation.rb
```diff
@@ -203,10 +203,11 @@ def add_pretty_id
...
+    return true if child_source&.is_a? ReportSource
     return true unless child_has_gdpr_sensitive_data
-    return true if User.current.organisation == self.source&.user&.organisation
+    return true if User.current.organisation == child_source&.user&.organisation
+    return true if User.current == self.assignee
```
MatthewBurstein: I think `User.current == self.assignee` will always be true if `self.assignee && (self.assignee.teams & User.current.teams).any?`, which would make this line unnecessary?

mspsds-web/app/models/correspondence.rb
```diff
@@ -30,6 +31,10 @@ def strip_whitespace
   end
 
   def can_be_displayed?
-    investigation.can_display_child_object? has_consumer_info
+    investigation.can_display_child_object?(has_consumer_info, source)
```
jasiekmiko: Also I'm not sure what the benefit of having cases check the child's source for them? would it not be clearer and more encapsulated to have the children check their source and then ask the investigation if there are other reasons why it shouldn't be displayed? so something like
```
def can_be_displayed?
  return true if source.is_a? ReportSource
  return true if User.current.organisation == child_source&.user&.organisation
  investigation.can_display_child_object? has_consumer_info
end
```

I'm not sure if this is exactly the right form, either. But I'm generally a bit worried about the messiness of the logic around this atm. happy to pair up on making it more direct if that would help! :)

mspsds-web/app/models/investigation.rb
kicferk1: It won't be true if we have team-less users :( 
Which actually happens in our test setup with Ts_user

mspsds-web/app/models/correspondence.rb
kicferk1: I moved report-related logic to report classes, which made it easy to include it on child objects. Hopefully should be a bit more straightforward now :)

mspsds-web/app/models/investigation.rb
jasiekmiko: Let's update the test to not do that then, since a mspsds user is only valid if they have a team

mspsds-web/app/models/investigation.rb
```diff
@@ -203,10 +203,10 @@ def add_pretty_id
     self.pretty_id = "#{created_at.strftime('%y%m')}-%04d" % (cases_before + 1)
   end
 
-  def can_display_child_object?(child_has_gdpr_sensitive_data, child_source)
-    return true if child_source&.is_a? ReportSource
-    return true unless child_has_gdpr_sensitive_data
-    return true if User.current.organisation == child_source&.user&.organisation
+  def can_display_child_object?(current_user_can_see_child)
```
jasiekmiko: Nice, this method has a nicely defined scope now!
Does it apply to children only and not the investigation itself?

mspsds-web/app/views/documents/_document_card.html.slim
```diff
@@ -1,7 +1,5 @@
 ruby:
-  source = document.attachments.find_by(record_type: "Activity")&.record&.source
-  sensitive = parent.respond_to?(:can_display_child_object?) &&
-          !parent.can_display_child_object?(document.metadata[:has_consumer_info], source)
+  sensitive = document_is_sensitive(document, parent)
```
jasiekmiko: that's much nicer!

mspsds-web/app/models/complainant.rb
```diff
@@ -10,11 +10,15 @@ class Complainant < ApplicationRecord
...
+    return true if investigation.source&.current_user_has_gdpr_access?
+
+    complainant_type != "Consumer"
   end
 
+private
+
   def can_be_displayed?
-    investigation.can_display_child_object?(contains_personal_data?, investigation.source)
+    investigation.can_display_child_object? can_be_seen_by_current_user?
```
jasiekmiko: Should these say `can_be_seen_by_current_user? && investigation.can_display_child_object?`? Or is there a reason why we want to be indirect like this?

mspsds-web/app/models/investigation.rb
kicferk1: Yes, it applies only to its children. The investigation visibility lives in InvestigationPolicy, and probably should be updated accordingly too.

mspsds-web/app/models/complainant.rb
kicferk1: I think you mean `||` rather than `&&`, because what happens now is investigation asks child about its opinion on whether it can be seen, and then says "I'll try to respect that, but my assignee has to be able to see you". And no, there is no need to be so indirect indeed :)

mspsds-web/app/helpers/documents_helper.rb
```diff
@@ -71,9 +71,10 @@ def formatted_file_updated_date(file)
   end
 
   def document_is_sensitive(document, parent)
-    visible = document.attachments.find_by(record_type: "Activity")&.record&.source&.current_user_has_gdpr_access?
-    visible = visible || !document.metadata[:has_consumer_info]
-    parent.respond_to?(:can_display_child_object?) && !parent.can_display_child_object?(visible)
+    child_is_visible = document.attachments.find_by(record_type: "Activity")&.record&.source&.user_has_gdpr_access?
+    child_is_visible = child_is_visible || !document.metadata[:has_consumer_info]
```
jasiekmiko: do we need reassignement here? can we not just have 
```
child_is_visible = document.attachments.find_by(record_type: "Activity")&.record&.source&.user_has_gdpr_access?
  || !document.metadata[:has_consumer_info]
```
and avoid mutating variables?

mspsds-web/app/models/user_source.rb
```diff
@@ -5,7 +5,7 @@ def show
     user.present? ? user.display_name : "anonymous"
   end
 
-  def current_user_has_gdpr_access?
-    User.current.organisation == user&.organisation
+  def user_has_gdpr_access?(user_in_question: User.current)
+    user_in_question.organisation == user&.organisation
```
jasiekmiko: perhaps `user.organisation == self.user&.organisation` would be more canonical?

jasiekmiko: moving this to a `user_source` policy might also make sense and make naming more straight-forward?

PR #817, opened 2019-03-25
COSBETA-165 Fix requirements for pre and post brexit notifications

DWRendell:
"As decided in the kick-off session (COSBETA‌-120), the CMR and nanomaterial properties should not be shown on the edit/show pages if the product was first notified on CPNP **prior** to Brexit date.

Product images are required before a product can be registered if it was first notified on CPNP **after** Brexit date."

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/models/notification.rb
+++ b/cosmetics-web/app/models/notification.rb
@@ -4,6 +4,8 @@ class Notification < ApplicationRecord
   include AASM
   include Shared::Web::CountriesHelper
 
+  EU_EXIT_DATETIME = DateTime.parse("29-03-2019T23:00:00").in_time_zone
+
   belongs_to :responsible_person
   has_many :components, dependent: :destroy
   has_many :image_uploads, dependent: :destroy
@@ -137,11 +139,11 @@ class Notification < ApplicationRecord
   end
 
   def notified_post_eu_exit?
-    !was_notified_before_eu_exit? || (cpnp_notification_date.present? && (cpnp_notification_date > Date.parse("29-03-2019")))
+    !notified_pre_eu_exit?
   end
 
   def notified_pre_eu_exit?
-    !notified_post_eu_exit?
+    was_notified_before_eu_exit? || (cpnp_notification_date.present? && (cpnp_notification_date < EU_EXIT_DATETIME))
   end
 
 private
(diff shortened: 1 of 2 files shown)


cosmetics-web/app/models/notification.rb
```diff
@@ -136,6 +136,14 @@ def is_multicomponent?
     components.length > 1
   end
 
+  def notified_post_eu_exit?
+    !was_notified_before_eu_exit? || (cpnp_notification_date.present? && (cpnp_notification_date > Date.parse("29-03-2019")))
```
tom-bell-softwire: We probably want to make this date more easily changeable. Perhaps pull it out into a constant?

Actually, should this be a DateTime, with EU exit technically being at 11pm on 29th?

cosmetics-web/app/models/notification.rb
```diff
@@ -136,6 +136,14 @@ def is_multicomponent?
...
+  end
+
+  def notified_pre_eu_exit?
+    !notified_post_eu_exit?
```
tom-bell-softwire: I'd be tempted to switch the logic between these two `pre`/`post` methods, to reduce the amount of `!` negations. So this would become something like:
```
was_notified_before_eu_exit? || (cpnp_notification_date.present? && cpnp_notification_date < EU_EXIT_DATE)
```

cosmetics-web/app/views/notifications/_component_details.html.slim
```diff
@@ -1,26 +1,27 @@
 /TODO: if post Brexit?
```
tom-bell-softwire: Mind removing this TODO?

cosmetics-web/app/models/notification.rb
DWRendell: Good idea. Just had a thought, do we need to worry about Timezones? Are the dates in the export likely to be in Central European Time, or are we already taking that into account?

cosmetics-web/app/models/notification.rb
DWRendell: Done

PR #819, opened 2019-03-25
COSBETA-169: Add error highlight when uploading formulation fails

TiagoSusanoPinto:
Show an error, highlighting the form when occurs an error while uploading a formulation file (e.g.: no file, wrong file type)

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/views/formulation_upload/new.html.slim
+++ b/cosmetics-web/app/views/formulation_upload/new.html.slim
@@ -15,7 +15,7 @@
         p For each ingredient give the International Nomenclature of Cosmetic Ingredients (INCI) name.
         = form_with(url: responsible_person_notification_component_formulation_index_path( \
                 @responsible_person, @notification, @component), method: :post, multipart: true) do |form|
-            .govuk-form-group
+            .govuk-form-group class=("govuk-form-group--error" if @error_list.any?)
                 = form.label :formulation_file, "File type can be a ‘pdf', 'rtf' or ‘txt’.", class: "govuk-label"
                 = form.file_field :formulation_file,  class: "govuk-file-upload", \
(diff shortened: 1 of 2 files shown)


cosmetics-web/app/views/formulation_upload/new.html.slim
```diff
@@ -15,7 +15,7 @@
         p For each ingredient give the International Nomenclature of Cosmetic Ingredients (INCI) name.
         = form_with(url: responsible_person_notification_component_formulation_index_path( \
                 @responsible_person, @notification, @component), method: :post, multipart: true) do |form|
-            .govuk-form-group
+            .govuk-form-group class=("govuk-form-group--error" if @error_list.any?)
                 = form.label :formulation_file, "File type can be a ‘pdf', 'rtf' or ‘txt’.", class: "govuk-label"
```
tom-bell-softwire: We also need to display the error message here, too...

TiagoSusanoPinto: We already display the error on lines 6-7
```
        - if @error_list.any?
            = render "components/govuk_error_summary", titleText: "There is a problem", errorList: @error_list
```

Do we want to display the same error message here too?

tom-bell-softwire: Yes – here's what the Design System [has to say](https://design-system.service.gov.uk/components/error-summary/):

> When a user makes an error, you must show both an error summary and an error message next to each answer that contains an error.
> ...
> Always show an error summary when there is a validation error, even if there’s only one.

PR #821, opened 2019-03-26
MSPSDS-NONE: Fix loading users

kicferk1:
It seems Users.all was broken on int and we caught it on staging. 

The bit populating organisations needs organisations and teams loaded, or it will reject them all(or most of them) on account of having no organisation, and therefore not being mspsds. Organisations should have been loaded by this point, but as it happens, this was done in the first request we make to keycloak after starting the app(I think), and for whatever reason the first request gets rejected :( 

This wasn't caught because we load keycloak entities more often, to not need to sign out and back in after a hot reload. Removing the line `xxx.all if Rails.env.development?` from Team.rb and Organisation.rb allowed to replicate it locally.


shared-web/config/initializers/keycloak.rb
```diff
@@ -9,8 +9,15 @@
 # Realm url (only used if the installation file is not present)
 Keycloak.auth_server_url = ""
 
+begin
+  Shared::Web::KeycloakClient.instance.all_organisations
```
broder: What's this doing?

PR #824, opened 2019-03-26
MSPSDS-615: Rename MSPSDS to PSD

broder:
## Description
INB4 bugs and merge conflicts.

Things that are still left to rename are the PaaS space and the repo name.

## Checklist:


README.md
```diff
@@ -1,4 +1,4 @@
-# Market Surveillance & Product Safety Digital Service
+# Office for Product Safety & Standards Services
```
jasiekmiko: Do we want to rename the github repo, too?

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ All notable changes to this project will be documented in this file.
 - [ ] Delete mspsds `admin` role (Clients->mspsds-app->roles->admin->delete)
 - [ ] Create the new environment variable services.
 - [ ] Create the `opss-cdn-route` service with the live and deployment URLs.
+- [ ] Rename the `mspsds-app` client to `psd-app`
```
jasiekmiko: also rename the `mspsds_user` role to `psd_user`?

jasiekmiko:
Renaming the realm was a hiccup in a previous release iirc. is this running the risk of similar?

kicferk1:
Does that mean our branches and commits should be PSD-XXXX instead of MSPSDS-XXXX?

README.md
broder: We will - I was thinking about doing that a bit later though as it'll probably break Travis and other things like that.

broder: Also, I'm not 100% sure on the repo name right now and didn't really want to hold this up for much longer

CHANGELOG.md
broder: Yes, thanks!

broder:
@jasiekmiko, well remembered! The issue was that it re-created the realm because it uses the realm name to check whether or not it already exists so we should have renamed the realm before deploying. I think we'll be okay for this change because we're not touching the realm name

broder:
@kicferk1, not yet. There's probably a separate action to rename the Jira / Confluence space.

README.md
jasiekmiko: sounds sensible :)

PR #827, opened 2019-03-26
MSPSDS-356 enable processing team members to manage users

jasiekmiko:
## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,14 +15,14 @@ All notable changes to this project will be documented in this file.
 ### Next release checklist
 - [ ] Add `team_admin` role to mspsds client on keycloak
 - [ ] Allow mspsds to view clients info on keycloak
-    # Log into keycloak admin app, click on `Clients` and select `mspsds-app`
-    # In the `Service Account Roles` tab, select `realm-management` from the dropdown and assign the `view-clients` role
-    # In the `Scope` tab, select `realm-management` from the dropdown and assign the `view-clients` role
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,17 +15,24 @@ All notable changes to this project will be documented in this file.
...
+    1. Log into keycloak admin app, click on `Clients` and select `mspsds-app`
+    1. In the `Service Account Roles` tab, select `realm-management` from the dropdown and assign the `manage-users` role
+    1. In the `Scope` tab, select `realm-management` from the dropdown and assign the `manage-users` role
 - [ ] Delete mspsds `admin` role (Clients->mspsds-app->roles->admin->delete)
 - [ ] Create the new environment variable services.
 - [ ] Create the `opss-cdn-route` service with the live and deployment URLs.
+- [ ] Update keycloak setup to allow processing team members to manage users
+    1. Log into keycloak admin app (master realm)
+    1. Go to the `Groups` section, select `OPSS Processing` and press `Edit`
+        1. Go to the `Role Mappings` tab, select `admin` from the `Realm Roles` and press `Add selected`
```
broder: I don't think we'll want all members of the processing team to be `admin`s - probably just specific users.

jasiekmiko: I've double checked with Laurence and the default we should go with for now is to give the whole team access. If we were creating them accounts on the master realm, then we'd be limiting it to just individuals.

broder: Cool, thanks!

PR #830, opened 2019-03-27
MSPSDS-1116: Don't run keycloak code on sidekiq

broder:
## Description
A fix for #821 

## Checklist:


broder:
@kicferk1, I think the worker would have crashed locally if you had restarted it?

kicferk1:
I think we only restarted web with these changes. Good thing to watch out for when changing code in initialisers then.

PR #831, opened 2019-03-27
Staging release

broder:
I'd like to wait for at least #827 and #829 to be merged before we do this release.


jasiekmiko:
Looks good, but trying to approve it makes me go to the diff page which crashes my browser 😑🌴

Probably needs to be done on staging before/during/after:
- [ ] Rename the `mspsds-app` client to `psd-app`
- [ ] Rename the `mspsds_user` role to `psd_user`
- [ ] Update the `psd-app` client to use the newly renamed `govuk-psd` theme
- [ ] Rename all `mspsds-*` CF services to `psd-*`
- [ ] Update keycloak setup to allow processing team members to manage users
    1. Log into keycloak admin app (master realm)
    1. Go to the `Groups` section, select `OPSS Processing` and press `Edit`
        1. Go to the `Role Mappings` tab, select `admin` from the `Realm Roles` and press `Add selected`
    1. Go to the `Roles` section and click on the `admin` role
        1. Under `Composite Roles` select `realm-management` from the dropdown
        1. Select `manage-users`, `view-clients` and `view-realm`, press `Add selected`

PR #832, opened 2019-03-27
MSPSDS-1103: Footer bugfixes and design updates

kicferk1:
Fix the links so that invisible text is visible to screen readers, remove OGL logo and mentions, order links alphabetically. 

## Checklist:

### Accessibility testing
- [x] Works keyboard only

### DIFF
--- a/psd-web/app/assets/application/stylesheets/footer.scss
+++ b/psd-web/app/assets/application/stylesheets/footer.scss
@@ -1,0 +1,3 @@
+.govuk-footer__meta-item > :last-child {
+  margin-bottom: 0;
+}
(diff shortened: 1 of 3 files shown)


psd-web/app/assets/application/stylesheets/footer.scss
```diff
@@ -0,0 +1,3 @@
+.govuk-footer__meta-item > :last-child {
```
broder: Could we do this with `govuk-!-margin-bottom-0` on the element?

kicferk1: We could, but Ed said the footer is likely to get more items and that the margin should only be removed from the final one.

broder: Fair enough!

PR #834, opened 2019-03-27
COSBETA-34: Poison Centre view for cosmetics products

raamSoftwire:
Update the product page for poison centres

## Description
- split up the existing partials into new tables
- included product image
- add formulation data from trigger rules

## Checklist:
### General testing
- [ ] Test without javascript
- [X] Test on small screen
### Accessibility testing
- [X] Works keyboard only
- [X] Passes automated checker (automated in build or manual)


cosmetics-web/app/helpers/notification_properties_helper.rb
```diff
@@ -418,14 +418,14 @@ def get_exposure_condition_name(exposure_condition)
...
-      greater_than_75_less_than_100_percent: "> 75 - ≤ 100%"
+      less_than_01_percent: "≤ 0.1",
+      greater_than_01_less_than_1_percent: "> 0.1 - ≤ 1",
+      greater_than_1_less_than_5_percent: "> 1 - ≤ 5",
+      greater_than_5_less_than_10_percent: "> 5 - ≤ 10",
+      greater_than_10_less_than_25_percent: "> 10 - ≤ 25",
+      greater_than_25_less_than_50_percent: "> 25 - ≤ 50",
+      greater_than_50_less_than_75_percent: "> 50 - ≤ 75",
+      greater_than_75_less_than_100_percent: "> 75 - ≤ 100"
   }.freeze
```
raamSoftwire: Units are now being handled in the slim files.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
```diff
@@ -2,9 +2,27 @@
...
+      = @notification.responsible_person.name
+    h3.govuk-heading-s[class="govuk-!-font-weight-regular"]
+      - if @notification.components&.first&.sub_sub_category&.present?
+        = @notification.components.first.display_root_category + ", " + \
+            @notification.components.first.display_sub_category + ", " + \
+            @notification.components.first.display_sub_sub_category
+
+.govuk-grid-row
+  .govuk-grid-column-one-third.product-images
+    - if @notification.image_uploads.present?
```
tom-bell-softwire: I think we should loop over the `image_uploads` and show them all, so this could change to:
```
- @notification.image_uploads.each do |image|
  - if image.marked_as_safe?
    = link_to ...
```

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
```diff
@@ -2,9 +2,27 @@
...
-    = render "notifications/formulation_details", notification: @notification
-    = render "notifications/responsible_person_details", responsible_person: @notification.responsible_person
-    = render "notifications/product_details", notification: @notification
+    h2.govuk-heading-m
+      = @notification.responsible_person.name
+    h3.govuk-heading-s[class="govuk-!-font-weight-regular"]
+      - if @notification.components&.first&.sub_sub_category&.present?
+        = @notification.components.first.display_root_category + ", " + \
+            @notification.components.first.display_sub_category + ", " + \
+            @notification.components.first.display_sub_sub_category
```
tom-bell-softwire: Worth making this into a helper method, something like `full_category_display_name` ?

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
```diff
@@ -2,9 +2,27 @@
...
+.govuk-grid-row
+  .govuk-grid-column-one-third.product-images
+    - if @notification.image_uploads.present?
+      = link_to image_tag(url_for(@notification.image_uploads.first.file.variant(resize: "450x450"))),
+              url_for(@notification.image_uploads.first.file)
+  .govuk-grid-column-two-thirds
+    = render "poison_centres/notifications/poison_centre_trigger_rule_formulation", notification: @notification
+    = render "notifications/ingredient_details", notification: @notification
+    = render "notifications/frame_formulation_details", notification: @notification
+    = render "notifications/substances_details", notification: @notification
```
tom-bell-softwire: We should only show this information for post-EU exit notifications. If you merge in master, there should be a `notified_post_eu_exit?` method on the `Notification` model that you can use to check this.

cosmetics-web/app/views/poison_centres/notifications/_poison_centre_trigger_rule_formulation.html.slim
```diff
@@ -0,0 +1,34 @@
...
+                    = "#{question_element.answer} %w/w"
+              - when "inciname"
+                tr.govuk-table__row
+                  - formulation_data = get_formulation_data(question, question_element)
+                  - if formulation_data.present?
+                    th.govuk-table__header[class="govuk-!-width-one-half" scope="row"]
+                      = formulation_data.first.humanize
+                    td.govuk-table__cell.govuk-table__cell--numeric
+                      = "#{formulation_data.last.humanize} %w/w"
+
```
tom-bell-softwire: Extra empty line.

cosmetics-web/app/views/notifications/_ingredient_details.html.slim
```diff
@@ -0,0 +1,21 @@
...
+- notification.components.each do |component|
+  table.govuk-table
+    - if component.name.present?
+      caption.govuk-table__caption.govuk-heading-m
+        = component.name
+    - if component.exact? && component.exact_formulas.present?
+      = render "formulation_table", entities_list: component.exact_formulas,
+              key_name: :inci_name, value_name: :quantity
+    - elsif component.range? && component.range_formulas.present?
+      = render "formulation_table", entities_list: format_range_formulas(component.range_formulas),
```
tom-bell-softwire: I think we'll need to update the `format_range_formulas` method to include the percentage units, now that they're no longer in the `UNIT_NAME` values.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: Good thinking

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: Yes, it definitely is. Done

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: To be clear here, which information? Just substances?

cosmetics-web/app/views/poison_centres/notifications/_poison_centre_trigger_rule_formulation.html.slim
raamSoftwire: Gone

cosmetics-web/app/views/notifications/_ingredient_details.html.slim
raamSoftwire: Done, thanks

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: Also, should we be hiding information from the poison centres?

cosmetics-web/app/helpers/trigger_rules_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+  def get_inciname_incivalue_pair(question, element)
+    question_id = element.trigger_question_id
+    answer_order = element.answer_order
+    inciname_incivalue_pair = []
+    question.trigger_question_elements.where(
+      trigger_question_id: question_id, answer_order: answer_order
+    ).each do |question_element|
+      inciname_incivalue_pair << question_element.answer
+    end
+    p inciname_incivalue_pair
```
TiagoSusanoPinto: Do we want to print it?

Also, why is it called ..._pair, if the array only contains answers from the question_element? I'm probably missing something :p

cosmetics-web/app/helpers/category_helper.rb
```diff
@@ -0,0 +1,7 @@
+module CategoryHelper
+  def full_category_display_name(component)
+    component.display_root_category + ", " + \
```
TiagoSusanoPinto: When displaying the component's category name, do you want to display the root category?

Asking this because while developing the manual journey, I've noticed that on the contingency plan we only display the sub-category and sub-sub-category, separated by an '-' (e.g. "Skin care products - Face mask").

cosmetics-web/app/helpers/trigger_rules_helper.rb
tom-bell-softwire: Our handling of answers to trigger questions in the data model is based for now on the structure of the CPNP files (which we'll probably want to change longer term). For answers that expect a list of ingredients and their concentrations, the ingredient INCI name and the concentration value are stored as separate `answer` entries.

tom-bell-softwire: Good catch on the printing, BTW!

cosmetics-web/app/helpers/category_helper.rb
tom-bell-softwire: I think we do here. The way it's displayed in the manual journey will likely change as we refine those pages.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
tom-bell-softwire: Not critical, but FYI, the `if @notification.image_uploads.present?` check isn't required here, since `has_many` associations on models [always return an empty array](https://apidock.com/rails/v4.2.1/ActiveRecord/Associations/ClassMethods/has_many) rather than nil, if none are found.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
tom-bell-softwire: We don't show CMR and nanomaterial details for products notified pre-EU exit ([details](https://regulatorydelivery.atlassian.net/browse/COSBETA-165)).

That might be a follow-up question for Ralph and Euan, but for now, assume that if we aren't showing it to the business, we can't show it to Poison Centres.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: That's good to know. Thanks for the tip! I've fixed that up here.

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
raamSoftwire: Fair enough, I've added that now.

cosmetics-web/app/helpers/trigger_rules_helper.rb
raamSoftwire: Cheers for that Tiago, great spot!

PR #835, opened 2019-03-27
COSBETA-214: Change date formats to include the full month

TiagoSusanoPinto:
Change date formats to include the full month instead of dd/mm/yyyy
* In the new format, it should not display zero-padding values (e.g: display 1 Jan... instead of 01 Jan...)
* On the registered tab display the date as before (dd/mm/yyyy), but with tabular numbers

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/helpers/date_helper.rb
```diff
@@ -0,0 +1,9 @@
+module DateHelper
+  def display_full_month_date date
```
tom-bell-softwire: Can you add parentheses around the arguments in both these methods? While not required by Ruby, it's a convention we're trying to follow for method declarations (less so when calling methods).

shared-web/app/assets/application/stylesheets/text.scss
```diff
@@ -0,0 +1,3 @@
+.app-font--tabular {
```
tom-bell-softwire: Better to name this class for what it represents, perhaps something like `tabular-data` ? (Naming things is hard...)

PR #836, opened 2019-03-27
PSD-1044 repeat step error

MatthewBurstein:
## Checklist:


jasiekmiko:
I'll take this one over :)

psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
```diff
@@ -14,14 +14,11 @@
...
-      = render "components/govuk_radios",
-              name: further_key,
-              errorMessage: error_text && { text: error_text },
+      = render "form_components/govuk_radios",
+              form: form, key: further_key,
               fieldset: { legend: { text: other_text, classes: "govuk-fieldset__legend--m" } },
-              items: [{ text: "Yes", value: "Yes", checked: @repeat_step == "Yes",
-                      id: "#{further_key}_yes" },
-                      { text: "No", value: "No", checked: @repeat_step == "No", id: "#{further_key}_no" }]
+              items: [{ text: "Yes", value: "Yes", checked: @repeat_step == "Yes" },
```
broder: Are we now setting `@repeat_step` to be `true` / `false` rather that `"Yes"` / `"No"`? https://github.com/UKGovernmentBEIS/beis-mspsds/pull/836/files#diff-46445ecbbea8ff13f67c7ac5d671a650R147

psd-web/app/controllers/investigations/ts_investigations_controller.rb
```diff
@@ -331,36 +338,58 @@ def store_repeat_step
...
-    else
+    end
+  end
+
+  def repeat_step_valid?(model)
+    if @repeat_step.nil?
       further_page_type = to_item_text(step)
-      @investigation.errors.add(further_key(step), "Select whether or not you have #{further_page_type} to record")
+      model.errors.add(further_key(step), "Select whether or not you have #{further_page_type} to record")
+      false
```
broder: Do we need to do `return false` here?

jasiekmiko: Ahh, thanks! I've first put a `return` on both `false` and `true`, then figured I don't need it on `true and didn't think any better after that so removed from both 🤣

psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
jasiekmiko: Good spot! Fix proposed in https://github.com/UKGovernmentBEIS/beis-mspsds/pull/895

PR #838, opened 2019-03-27
MSPSDS-997: Make assignee and description migration not update timestamps

kicferk1:
So that the cases don't get 'updated_at' changed to 'the last time we deployed'


broder:
This is covered by #847.

PR #839, opened 2019-03-27
MSPSDS-575: Fix investigation created template_id

broder:


broder:
This is covered by #847.

PR #840, opened 2019-03-27
MSPSDS-1031: Fix broken assignee logic

kicferk1:
Make sure gdpr logic works when case is assigned to team


broder:
This is covered by #847.

PR #841, opened 2019-03-27
MSPSDS-976 fix: add activity_types to all activities that need them

MatthewBurstein:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/models/audit_activity/document/base.rb
```diff
@@ -19,4 +19,8 @@ def attachment_type
   def attached_image?
     self.attachment.image?
   end
+
+  def sensitive_title
+    'Document added'
```
broder: Should we be consistent between `'` and `"`?

broder:
This is covered by #847.

PR #842, opened 2019-03-27
COSBETA-231: Add product data fields to data model

raamSoftwire:
Add extra fields to data model

- is product mixed?
- pH min value (at product level)
- pH max value (at product level)


cosmetics-web/db/migrate/20190327170534_add_product_details_to_notifications.rb
```diff
@@ -0,0 +1,11 @@
+class AddProductDetailsToNotifications < ActiveRecord::Migration[5.2]
+  def change
+    safety_assured do
+      change_table :components, bulk: true do |_t|
```
tom-bell-softwire: Should this be `:notifications` instead?

cosmetics-web/db/migrate/20190327170534_add_product_details_to_notifications.rb
```diff
@@ -0,0 +1,11 @@
...
+        add_column :notifications, :is_mixed, :boolean
```
tom-bell-softwire: It would be neater to call this something like `components_are_mixed`

You can then use the method `notification.components_are_mixed?` in the display logic.

cosmetics-web/db/migrate/20190327170534_add_product_details_to_notifications.rb
raamSoftwire: I guess it should. I followed the pattern from `20190320152131_add_details_to_notifications.rb`, so maybe that one is incorrect?

raamSoftwire: I've changed it here, rolled back and migrated the change to check it all works.

cosmetics-web/db/migrate/20190327170534_add_product_details_to_notifications.rb
raamSoftwire: Agreed, done

raamSoftwire:
I have now surfaced these details in the product details partial. They will also need adding to the poison centre product view once that it merged in.

PR #851, opened 2019-03-28
Staging fixes

broder:


psd-web/app/models/team.rb
```diff
@@ -26,12 +26,7 @@ def add_user(user_id)
   end
 
   def self.all(options = {})
-    begin
-      self.data = Shared::Web::KeycloakClient.instance.all_teams
-    rescue StandardError => error
```
broder: @tom-bell-softwire, can you remember why we were swallowing these errors?

tom-bell-softwire: No... at a guess, it might have been to handle the issue where the initial request on startup fails, but more likely it was just a first approach to guard against any problems with getting Keycloak permissions working properly.

tom-bell-softwire: If this fails, does the whole service fall over (I can't remember if `User.all` is used when setting/getting `current_user`)?

shared-web/app/models/shared/web/user.rb
```diff
@@ -14,12 +14,7 @@ def self.find_or_create(user)
       end
 
       def self.all(options = {})
-        begin
-          self.data = Shared::Web::KeycloakClient.instance.all_users
-        rescue StandardError => error
-          Rails.logger.error "Failed to fetch users from Keycloak: #{error.message}"
-          self.data = nil
-        end
+        self.data = Shared::Web::KeycloakClient.instance.all_users
```
tom-bell-softwire: If this fails, does the whole service fall over (I can't remember if `User.all` is used when setting/getting `current_user`)?

broder: `current_user` is populated from the keycloak token, I think. Either way, I think we maybe want things to fall over? It certainly results in some weird functionality on PSD.

jasiekmiko: setting `User.current` does a `find_or_create`, so shouldn't be problematic in of itself, but agreed that without users loaded, PSD is pretty dead. If that's not so on cosmetics, i suggest we put something along the lines of `begin super rescue ...` in cosmetic's `User.all`?

psd-web/app/models/team.rb
broder: One of the things it was doing, was preventing the asset compilation failing because it can't reach keycloak 🤦 I've reverted it for now...

PR #855, opened 2019-03-28
PSD-1121 MVP design snags

jasiekmiko:
Commit-by-commit probably easiest.

Already updated the email template to not use `name`


psd-web/app/jobs/send_alert_job.rb
```diff
@@ -1,9 +1,8 @@
 class SendAlertJob < ApplicationJob
-  def perform(recipients_details, email_subject, email_body)
-    recipients_details.each do |details|
+  def perform(addresses, email_subject, email_body)
```
broder: `email_addresses` or `emails` just to be 100% clear?

psd-web/config/locales/en.yml
```diff
@@ -100,3 +100,10 @@ en:
             date_decided:
               invalid: "Enter a real date when the corrective action was decided"
               date_missing_component: "Enter the date the corrective action was decided and include a day, month and year"
+        alert:
+          attributes:
+            summary:
+              required: "Enter an alert subject"
```
kicferk1: Rails uses locales to try and compose nicer error messages? Cool 👍

psd-web/app/mailers/notify_mailer.rb
```diff
@@ -21,12 +21,11 @@ def investigation_updated(investigation_pretty_id, name, email, update_text, sub
     mail(to: email)
   end
 
-  def alert(name, email, email_text, subject_text)
+  def alert(email, email_text, subject_text)
```
kicferk1: It feels weird to have different order of `email_text` and `subject_text` here than in `SendAlertJob`, especially since they have 3 different names between model, job and mailer. Perhaps we could pass them as hashes to make it more readable?

psd-web/config/locales/en.yml
jasiekmiko: Yeah! I've found this to be a much cleaner way of setting those messages, as well as a more flexible one (see, e.g., how [this one](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/d5116df14f3ec3292f16c0ef0c4d193353903b7f/psd-web/config/locales/en.yml#L70) sets it for all email addresses, not just on specific models!)
Moving all of our text to locale files would perhaps not be the right goal right now, but I think we it's definitely an appealing direction, and error messages feel like a good start :)

psd-web/app/mailers/notify_mailer.rb
kicferk1: A lot easier to follow :)

PR #864, opened 2019-03-29
PSD-1150: Fix crash when investigation has multiple correspondence types

broder:
## Description
This was breaking search if an investigation had multiple correspondence types. As we could be calling `include?` on `nil` (e,g, when trying to get `email_subject` on a phone call)

## Checklist:


jasiekmiko:
Does this fix PSD-11150?

broder:
@jasiekmiko, yes

PR #868, opened 2019-03-30
COSBETA-239: Map submit and search hostnames to app in deploy script

tom-bell-softwire:
## Description
Update the deploy script for `cosmetics-web` to map both the "submit" and "search" subdomains to the app (prefixed by the space name, e.g. `int-submit.cosmetic-...` and `int-search.cosmetic-...`).

### DIFF
--- a/cosmetics-web/deploy.sh
+++ b/cosmetics-web/deploy.sh
@@ -9,23 +9,27 @@ set -ex
 #
 # SPACE: the space to which you want to deploy
 
-DOMAIN=submit-cosmetic-product-notifications.service.gov.uk
+DOMAIN=cosmetic-product-notifications.service.gov.uk
 if [[ $SPACE == "prod" ]]; then
-    HOSTNAME=www
+    SUBMIT_HOSTNAME=submit
+    SEARCH_HOSTNAME=search
 else
-    HOSTNAME=$SPACE
+    SUBMIT_HOSTNAME=$SPACE-submit
+    SEARCH_HOSTNAME=$SPACE-search
 fi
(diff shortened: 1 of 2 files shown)


cosmetics-web/deploy.sh
```diff
@@ -55,21 +60,25 @@ fi
...
-fi
+# TODO COSBETA-240 Uncomment when service goes live
+#if [[ $SPACE == "prod" ]]; then
+#    cf unbind-service $NEW_APP cosmetics-auth-env
+#fi
 
-# Map the live hostname to the new app
-cf set-env $NEW_APP COSMETICS_HOST "https://$HOSTNAME.$DOMAIN"
+# Map the live hostnames to the new app
+cf set-env $NEW_APP COSMETICS_HOST "$SUBMIT_HOSTNAME.$DOMAIN"
```
jasiekmiko: This now no longer has the protocol - I'm assuming that's intentional, as you've done the same in the psd deploy, but just wanted to flag up in case not :)

tom-bell-softwire: Yep – this isn't required, since we explicitly set the protocol to "https" in the config files ([here](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/4a719c3360e7a3f697f900bd6d70c62cd51b8b0d/cosmetics-web/config/environments/production.rb#L53) and [here](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/4a719c3360e7a3f697f900bd6d70c62cd51b8b0d/psd-web/config/environments/production.rb#L31)).

PR #871, opened 2019-04-01
PSD-1039: Search cases by corrective action

kicferk1:
Added corrective actions and alerts, as well as ts-based fields to elasticsearch index. 

## Checklist:


psd-web/app/models/investigation.rb
```diff
@@ -61,9 +51,20 @@ class Investigation < ApplicationRecord
 
   after_create :create_audit_activity_for_case, :send_confirmation_email
 
+  # Elasticsearch index name must be declared in children and parent
+  index_name [Rails.env, "investigations"].join("_")
```
jasiekmiko: Thanks for grouping these together - much nicer! :D Would it be worth splitting it out into a separate module added to investigations, to enforce that grouping?

kicferk1: It would have made my job easier when I added a field to index but not to highlighted fields and was trying to figure out why I don't see any highlights :D

So yes, probably worth it.

jasiekmiko: nice!

PR #873, opened 2019-04-01
PSD-976: editing sensitive attachment

kicferk1:
When a sensitive attachment is edited or removed from a case, it should remain sensitive, and all activities that include it should not show it to unauthorised users. 

This adds a source user to document metadata, adds a migration to add source users to current documents, puts logic behind document visibility in a single helper. 

Also, since documents are editable, I added authorisation check for other actions as well. 

## Checklist:


psd-web/app/controllers/documents_controller.rb
```diff
@@ -40,6 +41,7 @@ def destroy
   def set_file
     @errors = ActiveModel::Errors.new(ActiveStorage::Blob.new)
     @file = file_collection.find(params[:id]) if params[:id].present?
+    raise Pundit::NotAuthorizedError if document_is_sensitive(@file, @parent)
```
jasiekmiko: Hmm, what's the intent behind this line? Because rn it reads as "if the document is sensitive, throw an error". That seems counter-intuitive? Should it say something along the lines of `... if ! authorized_to_view`?  And if so, would using a policy in this case make more sense?

kicferk1: Yeah, should probably rename `if document_is_sensitive` to `unless can_be_displayed?`, to follow pattern on other objects.

psd-web/app/views/documents/_document_card.html.slim
```diff
@@ -1,5 +1,5 @@
 ruby:
-  sensitive = document_is_sensitive(document, parent)
+  sensitive = !can_be_displayed?(document, parent)
```
jasiekmiko: Step in the right direction, but I thin we want to propagate it further - see how this line still equates the two concepts?

psd-web/app/models/audit_activity/document/base.rb
```diff
@@ -24,6 +24,6 @@ def attached_image?
   def sensitive_title; end
 
   def sensitive?
-    document_is_sensitive(self.attachment, self.investigation)
+    !can_be_displayed?(self.attachment, self.investigation)
```
jasiekmiko: ditto

psd-web/app/helpers/gdpr_helper.rb
```diff
@@ -1,9 +1,9 @@
...
-  def document_is_sensitive(document, parent)
+  def can_be_displayed?(document, parent)
     attachment_creator = User.find_by(id: document.metadata[:created_by])
     child_is_visible = UserSource.new(user: attachment_creator).user_has_gdpr_access? || !document.metadata[:has_consumer_info]
     parent_forces_to_be_visible = parent.respond_to?(:child_should_be_displayed?) && parent.child_should_be_displayed?
-    !(child_is_visible || parent_forces_to_be_visible)
+    child_is_visible || parent_forces_to_be_visible
   end
 
   def document_sensitive_title
```
jasiekmiko: In a similar vein, this probably wants to be called something along the lines of `restricted_title`, `title_when_restricted`, `public_title`, or similar!

PR #874, opened 2019-04-01
MSPSDS-1139 Move all heading sizes to new, smaller standard

jasiekmiko:
By default:
+ h1 -> size l
+ h2 -> size m
+ h3 -> size s
+ fieldset / label -> size m


psd-web/app/views/contacts/_form.html.slim
```diff
@@ -1,10 +1,15 @@
...
-      = render "form_components/govuk_input", key: :phone_number, form: form, label: { text: "Phone number" }
-      = render "form_components/govuk_input", key: :job_title, form: form,
-              label: { text: "Job title / description" }
-      .govuk-form-group
-        = form.submit "Save", class: "govuk-button"
+- label_classes = local_assigns[:label_classes].presence
+
+= render "components/govuk_fieldset", legend: legend do
+  = render "form_components/govuk_input", key: :name, form: form,
+          classes: "govuk-!-width-two-thirds",
```
kicferk1: Is the width a property of each field, or of the fieldset itself?

psd-web/app/views/investigations/meetings/context.html.slim
```diff
@@ -5,19 +5,17 @@
...
+              label: { text: "Who was the meeting with?", classes: "govuk-label--m" }
       .govuk-form-group
-        fieldset.govuk-fieldset aria-describedby="correspondence-date-hint" role="group"
-          legend.govuk-fieldset__legend.govuk-fieldset__legend--l
-            | Date of meeting
+        = render "components/govuk_fieldset",
+                describeBy: "correspondence-date-hint",
+                attributes: { role: "group" },
+                legend: { text: "Date of meeting", classes: "govuk-fieldset__legend--m" } do
         = render "date_input", form: form
```
kicferk1: `date_input` wraps its inputs in a `govuk-form-group`, so maybe we don't need form group just above it?

psd-web/app/views/investigations/tests/_form.html.slim
```diff
@@ -1,36 +1,34 @@
...
-  = render "form_components/autocomplete_select", items: relevant_legislation, key: :legislation, form: form,
-          show_all_values: true, hint: { text: "Select the relevant legislation from the list." }
+= render "form_components/autocomplete_select", key: :legislation, form: form,
+        items: relevant_legislation,
+        show_all_values: true,
+        hint: { text: "Select the relevant legislation from the list." },
+        label: { text: "Against which legislation?", classes: "govuk-label--m" }
 
-  legend.govuk-fieldset__legend.govuk-fieldset__legend--l
+fieldset.govuk-fieldset
```
kicferk1: Is there a reason we are not using the fieldset component here?

psd-web/app/views/investigations/ts_investigations/_file_page.html.slim
```diff
@@ -11,15 +11,10 @@ fieldset.govuk-fieldset
...
-        .govuk-grid-column-two-thirds
-          fieldset.govuk-fieldset
-            = render "form_components/govuk_input", key: :title, form: subform, value: @file_title,
-                    classes: "govuk-!-width-two-thirds",
-                    label: { text: "Title" }
-            = render "form_components/govuk_textarea", key: :description, form: subform, value: @file_description,
-                    attributes: { maxlength: 10000 }, label: { text: "Description" },
-                    hint: { text: "For example, 'Orange, 4 socket extension lead'" }
+      hr.govuk-section-break.govuk-section-break--m.govuk-section-break--visible
+      = render "form_components/govuk_input", key: :title, form: subform, value: @file_title,
```
kicferk1: I think the fields here will get wider after cutting `.govuk-grid-column-two-thirds`(as in, I think they stack up). Maybe worth checking with Ed if it's fine.

psd-web/app/views/locations/_address_form.html.slim
```diff
@@ -0,0 +1,20 @@
+= render "components/govuk_fieldset",
+        legend: { text: "Address", classes: "govuk-fieldset__legend--m" } do
+  = render "components/govuk_fieldset", legend: { text: "Building and street" } do
+    = render "form_components/govuk_input", key: :address_line_1, form: form,
+            label: { text: "Address line two", classes: "govuk-visually-hidden" }
```
kicferk1: `Address line one`?

psd-web/app/views/contacts/_form.html.slim
jasiekmiko: Field, and in particular just field and not the form group. We only want to be affecting the input box, but not the legend, say.

psd-web/app/views/investigations/meetings/context.html.slim
jasiekmiko: https://design-system.service.gov.uk/components/date-input/#date-input-example-html - there's a form group outside of the fieldset, too.
Quite probably, I may have missed it in other places the other way 'round - all that should hopefully get cleanedup somewhat once the we have a date input component :)

psd-web/app/views/investigations/ts_investigations/_file_page.html.slim
jasiekmiko: I've countered that by moving the `yield` in repeatable_info.html.slim into the 2/3s column :)

psd-web/app/views/locations/_address_form.html.slim
jasiekmiko: Keen 👀 !

psd-web/app/views/investigations/tests/_form.html.slim
jasiekmiko: Good spot :)

PR #882, opened 2019-04-03
PSD-976: Editing sensitive attachment

kicferk1:
Previous migration caused an error, likely because we can in fact have documents that are not attached to any activity. Namely whatever we attach to case in case creation flow(either ours or ts flow). 

This should add creators to those documents as well, and now it really shouldn't crash(I hope)


psd-web/db/migrate/20190329153321_add_created_by_to_file_blob.rb
```diff
@@ -4,9 +4,13 @@ def change
       dir.up do
         ActiveStorage::Blob.all.each do |blob|
           metadata = blob.metadata
-          first_activity = blob.attachments.where(record_type: "Activity").order(:created_at).first.record
-          blob.metadata.update(metadata.merge(created_by: first_activity&.source&.user_id))
-          blob.save
+          activity = blob.attachments.where(record_type: "Activity").order(:created_at).first
+          investigation = blob.attachments.where(record_type: "Investigation").order(:created_at).first
```
broder: Is the type definitely `Investigation` rather than `Investigation::Allegation` or similar?

kicferk1: Yes, I was pretty sure cause I checked that for activity, but I just double checked for investigation and it is parent class rather than full type.

kicferk1: Which probably makes sense since all attachment needs to know is table and id of the record.

PR #883, opened 2019-04-03
COSBETA-248: Display nanomaterials for both pre- and post-EU Exit notifications

DWRendell:
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/views/notifications/_substances_details.html.slim
+++ b/cosmetics-web/app/views/notifications/_substances_details.html.slim
@@ -6,11 +6,12 @@ h2.govuk-heading-l.poison-centre__table-heading
     - if component.name.present?
       caption.govuk-table__caption.govuk-heading-m
         = component.name
-    tr.govuk-table__row
-      th.govuk-table__header[class="govuk-!-width-one-half" scope="row"]
-        | CMR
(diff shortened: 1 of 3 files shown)


cosmetics-web/app/views/notifications/_substances_details.html.slim
```diff
@@ -6,11 +6,12 @@ h2.govuk-heading-l.poison-centre__table-heading
     - if component.name.present?
       caption.govuk-table__caption.govuk-heading-m
         = component.name
-    tr.govuk-table__row
-      th.govuk-table__header[class="govuk-!-width-one-half" scope="row"]
-        | CMR
-      td.govuk-table__cell
-        = render "none_or_bullet_list", entities_list: component.cmrs.map(&:display_name)
+    -if @notification.notified_post_eu_exit?
```
tom-bell-softwire: I think we should use the local `notification` here, since the `@notification` instance variable might not always be available when rendering this partial.

DWRendell: Good point, fixed

PR #886, opened 2019-04-03
PSD-989: Update date component

kicferk1:
This updates date component to mirror implementation of govuk macro. It adds a glue that handles some of error styling and setting values for fields. 
Date concern got updated to handle multiple date inputs per page easily. 

I also removed an unused and uncollected date field on product model, `date_placed_on_market`

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -23,84 +19,125 @@ def set_date(new_date)
...
-      if date.present? && date_components.all?(&:blank?)
-        self.day = date.day
-        self.month = date.month
-        self.year = date.year
-      else
-        update_from_components
+      keys = date_keys
+      keys.each do |key|
+        self.class_eval do
+          attr_accessor "#{key.to_s}_day".to_sym, :integer
```
jasiekmiko: worth extracting `key.to_s` to a variable?

psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -23,84 +19,125 @@ def set_date(new_date)
...
+          attr_accessor "#{key.to_s}_month".to_sym, :integer
+          attr_accessor ("#{key.to_s}_year".to_sym), :integer
+        end
+
+        date = self[key]
+        if date.present? && get_date_components(key).all?(&:blank?)
+          self.send("#{key.to_s}_day=", date.day)
```
jasiekmiko: cab we use `set_day` etc here?

psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -23,84 +19,125 @@ def set_date(new_date)
...
-      end
-    when 0
-      if self[@date_key].blank?
-        errors.add(@date_key, :invalid)
-        errors.add(:day, "")
-        errors.add(:month, "")
-        errors.add(:year, "")
+    date_keys.each do |key|
+      missing_date_components = {}
+      missing_date_components["#{key.to_s}_day".to_sym] = get_day(key)
```
jasiekmiko: Definitely feels like extracting some sort of `day_key(key)` method will be worth it with the amount of usage it's getting

psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -23,84 +19,125 @@ def set_date(new_date)
...
-        errors.add(missing_component, "")
-      end
-    when 0
-      if self[@date_key].blank?
-        errors.add(@date_key, :invalid)
-        errors.add(:day, "")
-        errors.add(:month, "")
-        errors.add(:year, "")
+    date_keys.each do |key|
+      missing_date_components = {}
```
jasiekmiko: can we call this `date_components` since it holds all of them? Then we're also avoiding overriding the variable.

psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -23,84 +19,125 @@ def set_date(new_date)
...
+      missing_date_components["#{key.to_s}_day".to_sym] = get_day(key)
+      missing_date_components["#{key.to_s}_month".to_sym] = get_month(key)
+      missing_date_components["#{key.to_s}_year".to_sym] = get_year(key)
+      missing_date_components = missing_date_components.select { |_, value| value.blank? }
+
+      case missing_date_components.length
+      when 3
+        errors.add(key, :invalid)
```
jasiekmiko: Don't think that's correct - we shouldn't dictate that every field is required! If we did though, we'd probably want to use `:blank` instead of `:invalid`, too :)

psd-web/app/controllers/investigations/correspondence_controller.rb
```diff
@@ -54,6 +54,7 @@ def set_investigation
 
   def set_correspondence
     @correspondence = model_class.new correspondence_params
+    @correspondence.update_dates_from_params(params[correspondence_params_key])
```
jasiekmiko: `set_dates_from_params`?
just since `update` usually means updating the db in rails' lingo

psd-web/app/controllers/concerns/date_concern.rb
```diff
@@ -1,37 +1,44 @@
...
+      set_year(key, new_date.year)
+      self[key] = new_date
     end
 
-    def set_date(new_date)
-      self.day = new_date.day
-      self.month = new_date.month
-      self.year = new_date.year
-      self[@date_key] = new_date
+    def get_date(key: date_keys.first)
```
jasiekmiko: I need to be convinced about this - why do we choose to cosntruct the date from components here, instead of returning `self[key]`?

psd-web/app/controllers/investigations/emails_controller.rb
```diff
@@ -38,9 +38,6 @@ def request_params
     params.require(correspondence_params_key).permit(
       :correspondent_name,
       :email_address,
-      :day,
```
jasiekmiko: we removed params date components from here, but don't seem to setting dates?

psd-web/app/controllers/investigations/phone_calls_controller.rb
```diff
@@ -26,9 +26,6 @@ def request_params
     params.require(correspondence_params_key).permit(
       :correspondent_name,
       :phone_number,
-      :day,
```
jasiekmiko: we removed params date components from here, but don't seem to setting dates?

psd-web/app/models/product.rb
```diff
@@ -7,7 +7,7 @@ class Product < ApplicationRecord
   include SanitizationHelper
 
   def date_keys
-    [:date_placed_on_market]
+    []
   end
```
jasiekmiko: Remove the whole method and concern inclusion?

psd-web/config/locales/en.yml
```diff
@@ -64,6 +64,7 @@ en:
       even: "%{attribute} must be even"
 
       # Our own
+      date_required: "%{attribute} can't be empty"
```
jasiekmiko: Nice to have a default, but I think there's a few improvements to this:
- The default should follow https://design-system.service.gov.uk/components/date-input/, so be something like "Enter the %{attribute}" (
image)
- I'm not convinced we want to be re-inventing the rails wheel - I'd be much keener to use `blank` for this. See https://github.com/UKGovernmentBEIS/beis-mspsds/pull/886/commits/a2a0530d3f4aa0ee456a58a8d77ebd73184920d6#diff-07308eeb711dc6bb0c2fa4d380cfb013L69 for a way to set it specifically to dates, and not other fields that happen to be `blank`
- If we really need to differentiate it, then we should at least stick to the terminology and make it `date_blank`

psd-web/test/system/investigation_test_result_test.rb
```diff
@@ -88,7 +88,7 @@ class InvestigationTestResultTest < ApplicationSystemTestCase
     fill_in "Year", with: "1984"
     click_on "Continue"
 
-    assert_text("Enter date of the test and include a day, month and year")
+    assert_text("Enter a date of the test and include a day, month and year")
```
jasiekmiko: Double checked with Amy?

psd-web/app/views/investigations/ts_investigations/test_results.html.slim
```diff
@@ -1,12 +1,14 @@
...
-        other_text: "Are there other test results to report?",
-        hide_break: false,
-        further_key: :further_test_results do |form|
-  = render "investigations/tests/form", form: form,
-          test: @test,
-          investigation: @investigation,
-          file_blob: @file_blob,
-          allow_product_linking: false
+.govuk-grid-row
+  .govuk-grid-column-two-thirds-from-desktop
```
jasiekmiko: Not sure this is how we want to do this - we're already setting 2/3s in the [repeatable_info file](https://github.com/UKGovernmentBEIS/beis-mspsds/blob/master/psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim#L6). If this needs to be further shortened, then we are probably looking for overrides on individual fields. Let's chat in person if this isn't quite describing it well enough

jasiekmiko:
Oh, also, on the things we talked about:
- did you manage to get the individual error boxes working?
- did you have a go at implementing the model interface as a `date_attribute :key` instead of having to override that date?

kicferk1:
I managed individual error boxes handling, and it works with validating different inputs with multiple date components :O 
I didn't figure date_attribute thing, but I tried it before I got everything working, so I can probably give it another go.

jasiekmiko:
Awesome! 👏

It might make it feel much more rails-y, which would be neat!

psd-web/app/controllers/investigations/emails_controller.rb
kicferk1: We do it in its parent, because it applies to all correspondence. `correspondence_controller`, in `set_correspondence`, we call `@corresponence.update_dates_from_params`

psd-web/config/locales/en.yml
kicferk1: This might well be touching on a more general issue of 'what should our error messages be', so for this ticket it's probably enough to use generic `blank`.

shared-web/app/views/form_components/_govuk_date_input.html.slim
```diff
@@ -0,0 +1,34 @@
...
+          },
+          {
+                  label: "year",
+                  classes: "govuk-input--width-4 #{'govuk-input--error' if errors&.full_messages_for(:year)&.any?}",
+                  value: form.object.send(:get_year, key),
+                  name: "#{get_attribute_name(form, key)}[year]"
+          }
+  ]
+  # local_assigns[:value] ||= form.object.send(key) if form.object.respond_to?(key)
+  if errors
```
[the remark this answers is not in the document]
kicferk1: I agree rails version of date input is unlikely to be used without a model.

shared-web/app/models/shared/web/concerns/date_concern.rb
```diff
@@ -1,73 +1,60 @@
...
-          after_initialize do
-            keys = date_keys
-            keys.each do |key|
-              self.class_eval do
-                attr_accessor "#{key}_day".to_sym, :integer
-                attr_accessor "#{key}_month".to_sym, :integer
-                attr_accessor "#{key}_year".to_sym, :integer
-              end
+          def self.date_attribute(key, required: true)
+            @@all_date_keys = @@all_date_keys.push([key, required, self.name]) unless @@all_date_keys.include?([key, required, self.name])
```
jasiekmiko: We definitely should not have a central repository of all of the date fields in the system here. We want to be storing that data on the class which includes `DateConcern` - not on the concern itself, which seems to be the current approach. 
Can't say I can see the right way off the top of my head, but we'll probably want to get to the stage where we're using an instance variable on the class, created inside of the `included` scope.
[This](https://stackoverflow.com/a/35272211/2363767) might be of some inspiration value?

shared-web/app/views/components/_govuk_error_message.html.slim
```diff
@@ -6,4 +6,4 @@ span.govuk-error-message *(local_assigns[:attributes] || {}) id=local_assigns[:i
   - visually_hidden_text = local_assigns[:visuallyHiddenText].nil? ? "Error" : local_assigns[:visuallyHiddenText]
   - if visually_hidden_text.present?
     span.govuk-visually-hidden = "#{visually_hidden_text}: "
-  = local_assigns[:html] || local_assigns[:text]
+  = local_assigns[:html] || local_assigns[:text].downcase.capitalize
```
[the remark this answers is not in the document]
kicferk1: Attribute passed to locales file seems to always be upper-case, so we get stuff like "Enter Date decided". Or "Enter Title", "Enter Name", etc. 

It becomes troublesome when we want to change from format "Name can't be blank" to "Enter name". I can push it back to rails glue components at the cost of more duplication :/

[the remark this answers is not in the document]
jasiekmiko: let's hold back on those changes then and attack them properly in a separate story :)

PR #893, opened 2019-04-04
PSD-1004: Create select component

kicferk1:
Create select component and restructure autocomplete select component.

Adds govuk select component and a glue for it. 
Also adds autocomplete select component and a glue for it, because it allows us tighter control over its elements(like x to delete entries). 

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only
Autocomplete js magic generates new input with the same id, and changes old inputs id to "#{id}-select". And even though old input gets display:none, Wave seems to expect a label for it anyway :/


shared-web/app/views/form_components/_govuk_select.html.slim
```diff
@@ -5,7 +5,10 @@ ruby:
   # Infer name and id_prefix from attribute key
   local_assigns[:name] ||= get_attribute_name(form, key)
   local_assigns[:id] ||= get_attribute_id_prefix(form, key)
-  local_assigns[:items] = items.map { |item| { value: item[1], text: item[0] } }
+  local_assigns[:items] = items.map { |item| {
```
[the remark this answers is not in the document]
kicferk1: I wanted to mimic rails `form.select`, which seems to handle both lists of two-element objects, and lists of strings as items with the same key. Not sure whether it should be done here, or whether whatever is using the component should make sure data is in nunjucks format.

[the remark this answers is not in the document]
jasiekmiko: I think it's a nice to have, but not at the expense of breaking nunjucks behaviour. Moving it to a different param (e.g. following [select](https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select) using `choices`) would ensure they can work side by side!
Then it just remains to weigh up the usefulness (e.g. do we currently use it anywhere in the codebase?) vs the extra complexity in the glue :)

shared-web/app/views/form_components/_govuk_select.html.slim
```diff
@@ -3,8 +3,8 @@
 ruby:
   # Infer name and id_prefix from attribute key
   local_assigns[:name] ||= get_attribute_name(form, key)
-  local_assigns[:id] ||= local_assigns[:is_autocomplete] ? "picker-#{key}" : get_attribute_id_prefix(form, key)
-  local_assigns[:items] = items.map do |item|
+  local_assigns[:id] ||= get_attribute_id_prefix(form, key)
+  local_assigns[:items] ||= local_assigns[:choices].map do |item|
```
jasiekmiko: As discussed in person:
Let's only allow the singletons list here, as we're not really gaining much by allowing list of arrays.
Let's also document what params we're making available in the component, even just with a comment :)

shared-web/app/views/components/_govuk_select.html.slim
```diff
@@ -21,14 +21,28 @@ ruby:
...
-                aria-describedby=described_by.presence]
-    - items.each do |item|
-      option[*(item[:attributes] || {})
-                value=item[:value]
-                selected=item[:selected]
-                disabled=item[:disabled]]
-        = item[:text]
+  / Both this div, the button below, and javascript diverge from govuk macro
+  / We do it to facilitate our changes to accessible autocomplete
+  div class=("autocomplete-select-with-clear" if local_assigns[:is_autocomplete] && local_assigns[:show_all_values])
```
jasiekmiko: Let's try and keep the difference we've created very separate (chatted in person to figure out a direction)
We talked about potentially adding all of it dynamically via js, but it did not seem worth it.
Also, can you make a note of deviations like this in the README for this?

cosmetics-web/app/views/component_build/select_frame_formulation.html.slim
```diff
@@ -6,9 +6,10 @@
...
       | Choose frame formulation
     = form_for(@component, url: wizard_path, method: :put) do |form|
-      = render "form_components/autocomplete_select", key: :frame_formulation, form: form, show_all_values: true,
+      = render "form_components/govuk_select", key: :frame_formulation, form: form, show_all_values: true,
+              is_autocomplete: true,
               items: Component.frame_formulations.keys.map { |formulation_key| \
-                [get_frame_formulation_name(formulation_key.to_sym), formulation_key] \
-              }
+                { text: get_frame_formulation_name(formulation_key.to_sym), value: formulation_key } \
+              }, label: { text: "Category", classes: "govuk-visually-hidden" }
```
jasiekmiko: Better way of integrating the label is to place the heading in it: https://design-system.service.gov.uk/get-started/labels-legends-headings/#labels-as-page-headings

cosmetics-web/app/views/component_build/select_category.html.slim
```diff
@@ -10,10 +10,11 @@
...
       | What category of cosmetic is the product?
       = form_for(@component, url: wizard_path, method: :put) do |form|
-        = render "form_components/autocomplete_select", key: :sub_sub_category, form: form, show_all_values: true,
+        = render "form_components/govuk_select", key: :sub_sub_category, form: form, show_all_values: true,
+                is_autocomplete: true,
                 items: get_sub_sub_categories.keys.map { |category_key| \
-                  [get_full_category_name(category_key), category_key] \
-                }
+                  { text: get_full_category_name(category_key), value: category_key } \
+                }, label: { text: "Category", classes: "govuk-visually-hidden" }
```
jasiekmiko: Better way of integrating the label is to place the heading in it: https://design-system.service.gov.uk/get-started/labels-legends-headings/#labels-as-page-headings

PR #895, opened 2019-04-04
PSD-1044 Fix regression in repeat_steps controls not retaining values

jasiekmiko:
Added to the test to guard against it in the future


psd-web/app/views/investigations/ts_investigations/_repeatable_info.html.slim
```diff
@@ -17,8 +17,8 @@
       = render "form_components/govuk_radios",
               form: form, key: further_key,
               fieldset: { legend: { text: other_text, classes: "govuk-fieldset__legend--m" } },
-              items: [{ text: "Yes", value: "Yes", checked: @repeat_step == "Yes" },
-                      { text: "No", value: "No", checked: @repeat_step == "No" }]
+              items: [{ text: "Yes", value: "Yes", checked: @repeat_step },
+                      { text: "No", value: "No", checked: (!@repeat_step.nil? && !@repeat_step) }]
```
jasiekmiko: This is a bit crap now, so looking for input on best way to deal with it. some thoughts on options
- leaving as is
- `@repeat_step == false` works, but makes RubyMine prompt about "fixing" it to `!@repeat_step` (which doesn't work). Can silence it with something like `- # noinspection RubySimplifyBooleanInspection`, but it does look a little unnatural and I'm sure someone will try "fixing" it down the line, too
- `@repeat === step` also happens to work and resembles js's "strong" comparison, though it's actually not that. It's got the updside of not triggering RM's inspection, too, though.
- reverting back to text-based booleans

broder: I'd go for leaving as is and put a comment explaining that rubocop complains about it OR "fixing" it and using the rubocop disabling comment.

PR #902, opened 2019-04-05
COSBETA-158 Prevent users from editing submitted notifications

DWRendell:
Prevents users from updating notifications once they've been submitted, either via replacing IDs in manual journey routes or the product for formulation upload pages.


cosmetics-web/spec/controllers/product_image_upload_controller_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.describe ProductImageUploadController, type: :controller do
```
tom-bell-softwire: Awesome! Thanks for adding this.

PR #906, opened 2019-04-08
PSD-1125: Use single name field on User model

tom-bell-softwire:
The user's full name is stored on Keycloak as the "firstName" field. The "lastName" field is no longer used.

### DIFF
--- a/cosmetics-web/spec/login_helpers.rb
+++ b/cosmetics-web/spec/login_helpers.rb
@@ -26,6 +26,6 @@ module LoginHelpers
   end
 
   def format_user_for_get_userinfo(user, groups: [])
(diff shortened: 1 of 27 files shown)


cosmetics-web/spec/login_helpers.rb
```diff
@@ -26,6 +26,6 @@ def sign_out
   end
 
   def format_user_for_get_userinfo(user, groups: [])
-    { sub: user[:id], email: user[:email], groups: groups, given_name: user[:first_name], family_name: user[:last_name] }.to_json
+    { sub: user[:id], email: user[:email], groups: groups, given_name: user[:name] }.to_json
```
broder: Should we continue to pass in `family_name` but give it a value of "n/a"?

tom-bell-softwire: Yes, good idea.

PR #909, opened 2019-04-08
COSBETA-256 Do not show links to upload product images or formulation…

DWRendell:
… files if unable to edit notification

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/responsible_persons/notifications/show.html.slim
```diff
@@ -11,4 +11,5 @@
   .govuk-grid-column-full
     = render "notifications/responsible_person_details", responsible_person: @notification.responsible_person
     = render "notifications/contact_person_details", contact_person: @responsible_person.contact_persons.first
-    = render "notifications/product_details", notification: @notification
+    = render "notifications/product_details", notification: @notification, \
+      allow_edits: !@notification.notification_complete?
```
tom-bell-softwire: I think `allow_edits` should always be `false` on this page.

cosmetics-web/app/views/responsible_persons/notifications/edit.html.slim
```diff
@@ -14,7 +14,8 @@
   .govuk-grid-column-full
     = render "notifications/responsible_person_details", responsible_person: @notification.responsible_person
     = render "notifications/contact_person_details", contact_person: @responsible_person.contact_persons.first
-    = render "notifications/product_details", notification: @notification
+    = render "notifications/product_details", notification: @notification, \
+      allow_edits: !@notification.notification_complete?
```
tom-bell-softwire: Should `allow_edits` always be `true` on this page?

cosmetics-web/app/views/responsible_persons/notifications/show.html.slim
DWRendell: Ah good point, didn't think of that

cosmetics-web/app/views/responsible_persons/notifications/edit.html.slim
DWRendell: Done

PR #911, opened 2019-04-08
COSBETA-237: Remove the need to be logged in to confirm contact person verification

aidayounesi:
Remove the need to be logged in to confirm contact person verification

## Description
Once the contact person has clicked a valid verification link in the email, take them to a new page:
```/.../contact-person-confirmation```

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### CODE REVIEW

File: cosmetics-web/config/routes.rb [hunk 1]
```diff
@@ -56,6 +56,13 @@
 
   resource :dashboard, controller: :dashboard, only: %i[show]
 
+  resources :confirmation, controller: :confirmation, param: :key, only: %i[] do
```
tom-bell-softwire: I wonder if there's a neater structure for these routes. Would it make sense to have `contact_persons` as a resource, with a nested `confirm` resource under that. The URLs would then look something like:
```
/contact_persons/{{contact_person_id}}/confirm/{{key}}
/contact_persons/{{contact_person_id}}/confirm/link-expired
```

Something like the following should be able to achieve that (though I haven't tested it):
```
resources :contact_persons do
  resources :confirmation, path: "confirm", param: :key, only: %i[show] do
    collection do
      get :link_expired, path: "link-expired"
    end
  end
end
```

cosmetics-web/app/views/confirmation/contact_person.html.slim
```diff
@@ -0,0 +1,9 @@
+- content_for :page_title, "Check your email"
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      ' You are confirmed as the contact person
```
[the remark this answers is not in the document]
aidayounesi: Sure thanks

cosmetics-web/config/routes.rb
tom-bell-softwire: Actually, perhaps we don't need to expose the `contact_person_id` in the above paths, in which case:
```ruby
resources :contact_persons do
  collection do
    resources :confirmation, path: "confirm", param: :key, only: %i[show] do
      collection do
        get :link_expired, path: "link-expired"
      end
    end
  end
end
```
would work better.

cosmetics-web/app/controllers/confirmation_controller.rb
```diff
@@ -0,0 +1,13 @@
+class ConfirmationController < ApplicationController
+  skip_before_action :authenticate_user!
+  skip_before_action :authorize_user!
+  skip_before_action :create_or_join_responsible_person
+
+  def contact_person
+    email_verification_key = EmailVerificationKey.find_by(key: params[:key])
+    return redirect_to "/404" if email_verification_key.blank?
```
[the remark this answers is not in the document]
aidayounesi: I commented the line for redirecting to 404 and got this error ```undefined method `is_expired?' for nil:NilClass```. I think it doesn't automatically render 404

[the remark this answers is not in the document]
aidayounesi: Found a solution
```You can specify an exclamation point (!) on the end of the dynamic finders to get them to raise an ActiveRecord::RecordNotFound error if they do not return any records```

File: cosmetics-web/app/controllers/confirmation_controller.rb [hunk 4]
```diff
@@ -0,0 +1,13 @@
...
+    return redirect_to link_expired_confirmation_path if email_verification_key.is_expired?
```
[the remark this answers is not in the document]
aidayounesi: Sure, thanks

cosmetics-web/config/routes.rb
```diff
@@ -56,6 +56,22 @@
...
+  resources :contact_persons, only: %i[] do
+    collection do
+      resources :confirmation, path: "confirm", param: :key, only: %i[show] do
+        collection do
+          get :link_expired, path: "link-expired"
+        end
+      end
+    end
+  end
+  # resources :confirmation, controller: :confirmation, param: :key, only: %i[] do
```
tom-bell-softwire: I think we should probably remove these commented lines.

cosmetics-web/spec/system/create_responsible_person_spec.rb
```diff
@@ -72,14 +72,13 @@
...
+  it "redirects to confirmation page on email validation" do
     responsible_person = create(:business_responsible_person)
     email_verification_key = responsible_person.email_verification_keys.create
 
-    sign_in_as_member_of_responsible_person(responsible_person)
-    visit responsible_person_email_verification_key_path(responsible_person, email_verification_key.key)
+    visit contact_person_confirmation_path(email_verification_key.key)
 
-    assert_current_path(/responsible_persons\/\d+/)
+    assert_current_path(/confirmation\/[a-zA-Z0-9_\-]+\/contact-person-confirmation/)
```
tom-bell-softwire: I think this assertion might need updating now that the routes have changed.

cosmetics-web/app/views/confirmation/contact_person.html.slim
```diff
@@ -0,0 +1,9 @@
+- content_for :page_title, "Check your email"
```
tom-bell-softwire: I think this file needs renaming to `show.html.slim`

Also, just spotted that the title needs updating, too.

PR #916, opened 2019-04-09
PSD-1176: Fix new teams breaking cases page

jasiekmiko:
The main fix is to load organisations when loading teams - otherwise we can end up with an orphaned team, which causes the breakage.

Some improvements to caching around this area.
@broder, I've tried to implement the changes we talked about (moving parsing-code from KC client to models), but it made for a much more confusing tests setup. I think the current implementation achieves the same goal of improve caching though!

Also fix to a auxillary issue where the same page breaks when we load in a user without a name (this was a regression bug due to changes of how we handle the name I believe).

Going commit-by-commit probably worthwhile.


shared-web/app/clients/shared/web/keycloak_client.rb
```diff
@@ -70,64 +70,65 @@ def initialize
...
+          end
         end
       end
 
-      def all_organisations
-        groups = all_groups
-        organisations = groups.find { |group| group["name"] == "Organisations" }
+      def all_organisations(force: false)
+        Rails.cache.delete(:keycloak_organisations) if force
+        Rails.cache.fetch(:keycloak_organisations, expires_in: 5.minutes) do
```
broder: What would happen if this cache and the team cache weren't in sync?

I think it wouldn't happen currently because we call them both at roughly the same time but it feels a little brittle that it implicitly relies on teams and orgs to be loaded at the same time.

jasiekmiko: That's true. I think in other places we guard against it by rejecting any info that doesn't match what we've fetched elsewhere, e.g. [here](https://github.com/UKGovernmentBEIS/beis-opss/pull/916/files/1d6eb55767e2fa44426928874aa6c143d847aae7#diff-eb638bc6d1bfc552d7c5966134f91f8cR124) we only add team_user info for users we already know about.

A similar approach here I think would be to `Organisation.load`, then to fetch groups data and only load team info for children of existing Orgs.

jasiekmiko: I've implemented it by requiring the dependencies to be passed in via arguments - let me know what you think! :)

shared-web/app/clients/shared/web/keycloak_client.rb
```diff
@@ -70,64 +70,59 @@ def initialize
...
-            id += 1
-          end
+      # @param org_ids specifies teams for which organisations should be returned. This allows us to avoid creating
+      # orphaned team entities
+      def all_teams(org_ids, force: false)
+        Rails.cache.delete(:keycloak_teams) if force
+        Rails.cache.fetch(:keycloak_teams, expires_in: 5.minutes) do
+          all_groups.find { |group| group["name"] == "Organisations" }["subGroups"]
+              .reject(&:blank?)
+              .select { |organisation| org_ids.include? organisation["id"] }
```
broder: Should we make `org_ids` (and some of the other `*_ids`) a set or are we not too concerned about performance?

PR #926, opened 2019-04-11
COSBETA-276 Prevent empty nanomaterial and CMR attributes showing up …

DWRendell:
…on component details pages

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/models/cmr.rb
```diff
@@ -2,6 +2,6 @@ class Cmr < ApplicationRecord
   belongs_to :component
 
   def display_name
-    [name, ec_number, cas_number].join(', ')
+    [name, ec_number, cas_number].reject(&:blank?).join(', ')
```
raamSoftwire: That is a neat solution! Cheers Dave

PR #928, opened 2019-04-11
PSD-1005: Create tabs component

kicferk1:
Adds tabs component, remove old tabs component. Very similar in use. 

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


cosmetics-web/app/views/responsible_persons/notifications/tabs/_errors.html.slim
```diff
@@ -1,3 +1,4 @@
+h2.govuk-heading-m = "Upload errors"
```
broder:
```suggestion
h2.govuk-heading-m Upload errors
```

and in other places too.

psd-web/db/migrate/20190411094416_remove_business_attachments.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveBusinessAttachments < ActiveRecord::Migration[5.2]
+  def change
+    ActiveStorage::Attachment.where(record_type: "Business").delete_all
```
broder: Should we be worried about deleting these attachments? Are we going to delete any live data?

kicferk1: We haven't been showing attachment tab on businesses for a long time, so ideally this will find nothing to delete. Even if it will, we are not surfacing business attachments anywhere, so I doubt anyone actually uses them for anything. And if anything is attached to a business and an investigation, it won't get deleted because delete_all doesn't trigger callbacks. 

As for whether we should bother to do this, I don't know. I don't think anything will crash if we don't.

cosmetics-web/app/views/responsible_persons/notifications/index.html.slim
```diff
@@ -42,11 +42,17 @@
...
-= render "components/tabs", tabs: [\
-  Tab.new("errors", "Errors (#{@erroneous_notification_files.total_entries})", "Upload errors",
-        lambda { render "responsible_persons/notifications/tabs/errors" }),
-  Tab.new("incomplete", "Incomplete (#{@unfinished_notifications.total_entries})", "Incomplete notifications",
-        lambda { render "responsible_persons/notifications/tabs/incomplete" }),
-  Tab.new("notified", "Notified (#{@registered_notifications.total_entries})", "Notified cosmetic products",
-        lambda { render "responsible_persons/notifications/tabs/notified" })\
-]
+- errors_html = capture do
+  = render "responsible_persons/notifications/tabs/errors"
```
jasiekmiko: I *think* `capture` is not needed when we're just spitting more ruby into it - try just doing `errors_html = render "..."`!

(which could potentially mean no need for variables altogether - depending on what looks best)

shared-web/app/views/components/_govuk_tabs.html.slim
```diff
@@ -0,0 +1,24 @@
+ruby:
+  id_prefix = local_assigns[:idPrefix]
+
+.govuk-tabs[*(local_assigns[:attributes] || {})
+                class=local_assigns[:classes]
+                id=local_assigns[:id]
+                data-module="tabs"]
+  h2.govuk-tabs__title
+    = local_assigns[:title] || "Contents"
+  - if local_assigns[:items]
```
jasiekmiko: In a deviation from the template.. `items` [are requried](https://design-system.service.gov.uk/components/tabs/#options-tabs-example), so feel free to use `items` instead of `local_assigns[:items]` everywhere, which would also render this `if` condition senseless :)

psd-web/db/migrate/20190411094416_remove_business_attachments.rb
kicferk1: Just checked, it seems nothing breaks without running this migration, even if there was something attached to a business. Question is, do we want to clean any potential business attachments or not?

jasiekmiko: Can you check what date we stopped allowing adding new business types? if it's before launch, yes we should remove it i think

kicferk1: Sooo, it turns out removing attachments tab from display was probably accidental, so instead of deleting everything that allows businesses to have documents attached I added it back. Which means we don't need this migration.

shared-web/app/views/components/_govuk_tabs.html.slim
```diff
@@ -0,0 +1,19 @@
...
+
+.govuk-tabs *attributes class=local_assigns[:classes] id=local_assigns[:id] data-module="tabs"
+  h2.govuk-tabs__title
+    = local_assigns[:title] || "Contents"
+  ul.govuk-tabs__list
+    - items.each_with_index do |item, index|
+      - id = item[:id] || id_prefix + '-' + index
+      - item_attributes = item[:attributes] || {}
+      li.govuk-tabs__list-item
+        a.govuk-tabs__tab *item_attributes class=("govuk-tabs__tab--selected" if index.zero?) href="##{id}"
```
tom-bell-softwire: Just out of interest, if you specify a different index for the selected tab, is that respected? It would be good to make this a parameter (something like `if item[:selected]`), rather than always defaulting to the first tab.

jasiekmiko: You have control over which tab you're navigating to via url anchors, which should cover the case you're thinking of I think?
See e.g. going to https://design-system.service.gov.uk/components/tabs/default/index.html#past-week

kicferk1: Plus, if you do make it a field on item, you get a problem of how to deal with 0 or more than 1 items being 'selected'.

tom-bell-softwire: @jasiekmiko The use case we have in mind is dynamically selecting a tab on page render depending on which tab(s) have content.

Happy to leave this for now, but it's likely something that we'll be adding shortly anyway!

PR #929, opened 2019-04-11
PSD-6: More account renaming

broder:
And other tech debt.

## Description
Probably worth going commit by commit

## Checklist:


psd-web/Gemfile.lock
```diff
@@ -455,7 +455,7 @@ DEPENDENCIES
   wicked
 
 RUBY VERSION
-   ruby 2.5.3p105
+   ruby 2.5.5p157
 
 BUNDLED WITH
-   1.17.2
+   1.17.3
```
tom-bell-softwire: Could we use v2.0.1 across the board?

broder: Ah, I thought I had kept it as 1.17.3 everywhere but it looks like I missed one.

The `ruby:2.5.5` docker image comes with 1.17.3 by default so it's probably easiest to just use that. The 2.0.1 crept in when I was using my system ruby to update the `Gemfile.lock`s

broder: Ah, I thought I had kept it as 1.17.3 everywhere but it looks like I missed one.

The `ruby:2.5.5` docker image comes with 1.17.3 by default so it's probably easiest to just use that. The 2.0.1 crept in when I was using my system ruby to update the `Gemfile.lock`s

cosmetics-worker/manifest.yml
```diff
@@ -1,7 +1,7 @@
 ---
 applications:
 - name: cosmetics-worker
-  memory: 512M
+  memory: 1G
```
jasiekmiko: was there a particular issue this is trying to solve?

broder: It was a speculative fix for https://sentry.io/organizations/beis/issues/978206675/?project=1329381&query=is%3Aunresolved&statsPeriod=14d

PR #930, opened 2019-04-11
COSBETA-316: Zip files get into in-complete and error tabs

aidayounesi:
## Description
Solution:  excluding invalid files and files from MACOSX zip files from analyzing

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -95,6 +95,7 @@ def get_product_xml_file
 
         file_found = false
         files.each do |file|
+          next if !file_is_valid?(file)
```
raamSoftwire: This is cool, I haven't seen a `next` break.

cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -112,14 +113,18 @@ def get_product_xml_file
...
+      file_is_valid?(file) && static_file?(file) && file_contents_differs?(file)
     end
   end
 
   def file_is_product_xml?(file)
     file.name&.match?(product_xml_file_name_regex)
   end
 
+  def file_is_valid?(file)
+    file.file? && file.name !~ /__MACOSX/ && file.name !~ /\.DS_Store/
```
raamSoftwire: I also haven't seen the `!~`, but it makes this check very readable.

aidayounesi: Yes, I was surprised that !~ exists instead of using ! xxx ~=

PR #932, opened 2019-04-11
COSBETA-179: Show cosmetic products page after signing in

aidayounesi:
## Description
Currently, the user is directed to the landing page after login.
They should be taken to the cosmetic products page instead.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/controllers/landing_page_controller.rb
```diff
@@ -16,4 +17,8 @@ def redirect_poison_centre_users
   def set_responsible_person
     @responsible_person = User.current.responsible_persons.first if user_signed_in?
   end
+
+  def redirect_users
+    redirect_to responsible_person_notifications_path(@responsible_person) if user_signed_in? && !poison_centre_or_msa_user?
```
tom-bell-softwire: I think this will redirect the user every time that they try to visit the landing page, whereas we only want to redirect them to the cosmetic products page once when they log in.

Probably best to have a quick chat about how we might be able to do that.

cosmetics-web/config/routes.rb
```diff
@@ -53,6 +53,9 @@
     post :accept
   end
 
+  resource :dashboard, controller: :dashboard, only: %i[show] do
+  end
```
tom-bell-softwire: I don't think you need the `do ... end` block here.

cosmetics-web/spec/controllers/dashboard_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+RSpec.describe DashboardController, type: :controller do
+  describe "When signed in as a Responsible Person user" do
+    let(:user) { build(:user) }
+    let(:responsible_person_1) { create(:responsible_person, email_address: "one@example.com") }
```
tom-bell-softwire: I think you can simplify this by removing the `let(:user)...` declaration and using:
```ruby
```

cosmetics-web/spec/controllers/dashboard_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    before do
+      responsible_person_1.add_user(user)
+      sign_in_as_member_of_responsible_person(create(:responsible_person), user)
```
tom-bell-softwire: These two lines can also be simplified to just:
```ruby
```
Unless you need to explicitly test properties of the User model, then this is simpler, and will create a `user` instance and associate it with the RP model for you.

cosmetics-web/app/controllers/dashboard_controller.rb
```diff
@@ -0,0 +1,9 @@
+class DashboardController < ApplicationController
+  skip_before_action :authorize_user!
+
+  def show
+    return redirect_to responsible_person_notifications_path(User.current.responsible_persons.first) unless poison_centre_or_msa_user?
```
tom-bell-softwire: Might be slightly more idiomatic to write this as an if-else block:
```ruby
if poison_centre_or_msa_user?
  redirect_to poison_centre_notifications_path and return
else
  redirect_to responsible_person_notifications_path(User.current.responsible_persons.first) and return
end
```

cosmetics-web/config/routes.rb
aidayounesi: Sure, thanks

cosmetics-web/spec/controllers/dashboard_controller_spec.rb
aidayounesi: Sure, thanks

cosmetics-web/spec/controllers/dashboard_controller_spec.rb
aidayounesi: Sure, thanks

cosmetics-web/app/controllers/dashboard_controller.rb
aidayounesi: Sure, thanks

PR #942, opened 2019-04-15
COSBETA-307: Fix race condition when fetching Keycloak token from request cookie

tom-bell-softwire:


shared-web/app/controllers/shared/web/concerns/authentication_concern.rb
```diff
@@ -9,10 +9,7 @@ module AuthenticationConcern
         include Shared::Web::LoginHelper
 
         def initialize
-          Keycloak.proc_cookie_token = lambda do
-            cookies.permanent[cookie_name]
-          end
-
+          Keycloak.proc_cookie_token = -> { nil }
```
jasiekmiko: Might be worth a comment why we need this at all now

tom-bell-softwire: Good call – now added.

PR #945, opened 2019-04-15
COSBETA-NONE: Add ElastAlert files

raamSoftwire:
Add ElastAlert Files


infrastructure/elastalert/cosmetics_web_4xx_frequency.yml
```diff
@@ -0,0 +1,47 @@
...
+# (Required)
+# The alert is use when a match is found
+alert:
+- "email"
+- "slack"
+
+# (required, email specific)
+# a list of email addresses to send alerts to
+email:
+- "user@example.com"
```
jasiekmiko: Are these placeholders to be added when applying this file on logit?

raamSoftwire: Yes, they're placeholders. I'll add a comment in the Readme that explains this.

raamSoftwire:
I'm not sure if we want metrics for 5xx errors. When @broder set up the original metrics they only included 2xx, 3xx and 4xx. If we do want them, I'm happy to add some in!

PR #949, opened 2019-04-16
PSD-6: Setup antivirus and maintenance continuous deployment

broder:
## Description
Add a healthcheck endpoint to the AV API and add CD for the antivirus and maintenance page.

## Checklist:


maintenance/deploy.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -ex
+
+# This is the CI server script to deploy the antivirus API
```
tom-bell-softwire: "antivirus API" -> "maintenance page"

maintenance/deploy.sh
```diff
@@ -0,0 +1,8 @@
...
+#
+# The working directory should be the git root
+
+cf push
```
tom-bell-softwire: Does this work from the git root directory?

tom-bell-softwire: Fixed by specifying the manifest file path.

PR #950, opened 2019-04-16
PSD-1128: Send welcome email

samlionheart:
Welcome email is sent to a user when they have set up their account and accepted the terms and conditions (declaration).
## Description
Welcome email sent after declaration is accepted. Boolean has_been_sent_welcome_email set to true to prevent them being sent a welcome email again, if the declaration changes and has to be accepted again.
## Checklist:


psd-web/db/migrate/20190416094522_add_has_been_sent_welcome_email_to_user_attributes.rb
```diff
@@ -0,0 +1,5 @@
+class AddHasBeenSentWelcomeEmailToUserAttributes < ActiveRecord::Migration[5.2]
+  def change
+    add_column :user_attributes, :has_been_sent_welcome_email, :boolean
```
jasiekmiko: Do we want this to be non-nullable? Or do we treat null as false?

PR #955, opened 2019-04-16
COSBETA-303: Add "Is product mixed" question to manual journey

TiagoSusanoPinto:
COSBETA-303: Add "Is product mixed" question

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/notification_build/is_mixed.html.slim
```diff
@@ -0,0 +1,15 @@
...
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+
+= form_with model: @notification, url: wizard_path, method: :put do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      = render "form_components/govuk_error_summary", form: form
+      - heading = capture do
+        h1.govuk-heading-xl Does the kit need to be mixed?
+      = render "form_components/govuk_radios", form: form, key: :components_are_mixed,
+              fieldset: { legend: { html: heading } },
```
jasiekmiko: Do double check that this would actually have the same visual effect, but I *think* you can avoid the need for capturing html here by doing something like:
`fieldset: { legend: { text: "Does the kit need to be mixed?", classes="govuk-fieldset__legend--xl" } }

TiagoSusanoPinto: Ohhh, that's really helpful!
And it works! :D

cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -71,28 +74,43 @@ def set_countries
     @countries = all_countries
   end
 
+  def component_params
+    return {} if params[:notification].blank?
+
+    params.require(:notification).permit(:components_are_mixed)
```
tom-bell-softwire: I think you can combine both lines into:
```ruby
params.fetch(:notification, {}).permit(...)
```

Also, could this just be combined with the existing `notification_params` method?

TiagoSusanoPinto: Right! Didn't notice that there was already a `notification_params` :)

cosmetics-web/app/views/notification_build/is_mixed.html.slim
```diff
@@ -0,0 +1,13 @@
+- content_for :page_title, "Does the kit need to be mixed?"
+- content_for :after_header do
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+
+= form_with model: @notification, url: wizard_path, method: :put do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      = render "form_components/govuk_error_summary", form: form
```
jasiekmiko: I know Ralph has a preference for the error summaries to exist outside of the grid / to occupy the full width of the page, but I'm not sure what the current convention is on cosmetics?

TiagoSusanoPinto: On the other files in cosmetics-web we normally use it inside the .govuk-grid-column-two-thirds.
But based on the conversation that we had with Ralph, we might want to change this (and the other cases) to outside of the grid, but add a grid-column-two-thirds inside the error banner

PR #962, opened 2019-04-17
COSBETA-NONE: Add logging for metrics from DB

raamSoftwire:
Add a regular job to log metrics from DB. We will probably add more metrics to this in future.


cosmetics-web/app/jobs/log_db_metrics_job.rb
```diff
@@ -0,0 +1,11 @@
+class LogDbMetricsJob < ApplicationJob
+  def perform
+    Sidekiq.logger.info "{
+      completeNotifications: #{Notification.where(state: 'notification_complete').count},
+      responsiblePersons: #{ResponsiblePerson.count},
+      usersPerResponsiblePerson: #{ResponsiblePersonUser.group(:responsible_person_id).count.values.sort.inject(
+        Hash.new(0)
+) { |h, e| h[e] += 1; h }}
```
raamSoftwire: This SQL I wanted for this last one is 

```SELECT *, COUNT(*) from (SELECT COUNT(*) as Users_Per_RP FROM public.responsible_person_users GROUP BY responsible_person_id)  as indiv_count
GROUP BY Users_Per_RP
ORDER BY Users_Per_RP;```

But I was unable to write the equivalent ActiveRecord Query.

File: cosmetics-web/config/initializers/sidekiq.rb [hunk 2]
```diff
@@ -1,4 +1,4 @@
-def remove_expired_email_verification_keys_job
+def create_remove_expired_email_verification_keys_job
```
raamSoftwire: I feel like this name helps to make it a bit clearer that it is not the job itself.

cosmetics-web/config/initializers/sidekiq.rb
```diff
@@ -11,9 +11,23 @@ def remove_expired_email_verification_keys_job
...
 
+def create_log_db_metrics_job
+  log_db_metrics_job = Sidekiq::Cron::Job.new(
+    name: 'log db metrics, every hour',
+    cron: '0 * * * *',
+    class: 'LogDbMetricsJob',
+    queue: 'cosmetics'
+  )
+  unless log_db_metrics_job.save
+    Rails.logger.error "***** WARNING - Log DB metrics job was not saved! *****"
```
raamSoftwire: I've copied these lines from the method above, but I think these should actually be `Sidekiq` logs instead of `Rails` logs. What do you think?

raamSoftwire:
This is really the first half of this work. The second half is picking up these logs from Kibana and creating visualizations from these numbers. This might involve more changes, but we won't be able to tell until we get these initial changes on to INT.

cosmetics-web/config/initializers/sidekiq.rb
```diff
@@ -11,9 +11,23 @@ def remove_expired_email_verification_keys_job
   end
 end
 
+def create_log_db_metrics_job
+  log_db_metrics_job = Sidekiq::Cron::Job.new(
+    name: 'log db metrics, every hour',
+    cron: '0 * * * *',
```
raamSoftwire: I changed my mind to run these every hour instead of every day. I feel like these logs are very cheap so hopefully not too much of an issue?

cosmetics-web/app/jobs/log_db_metrics_job.rb
raamSoftwire: Cool, thanks

cosmetics-web/config/initializers/sidekiq.rb
raamSoftwire: Thanks for clarifying this Jan

cosmetics-web/app/jobs/log_db_metrics_job.rb
```diff
@@ -0,0 +1,11 @@
+class LogDbMetricsJob < ApplicationJob
+  def perform
+    Sidekiq.logger.info "{
+      completeNotifications: #{Notification.where(state: 'notification_complete').count},
```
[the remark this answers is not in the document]
raamSoftwire: Very good point, I've updated that now

cosmetics-web/config/initializers/sidekiq.rb
tom-bell-softwire: Yep, good call!

PR #963, opened 2019-04-17
PSD-1039: Search by title even if it is auto-generated

kicferk1:
After chats with Nasir and Ed, we decided to allow searching by case title even if it is auto-generated. 

## Checklist:


psd-web/test/controllers/investigations_controller_test.rb
```diff
@@ -291,7 +291,7 @@ class InvestigationsControllerTest < ActionDispatch::IntegrationTest
 
   def create_new_private_case
     description = "new_investigation_description"
-    Investigation::Allegation.create(user_title: "title", description: description)
+    Investigation::Allegation.create(description: description)
```
jasiekmiko: How come we're removing this? If it's to test that searching by auto-generated title works then it should be much more explicit - probably with separate cases!

kicferk1: We are removing this because it shouldn't have been here in the first place - no creation flow sets user_title for allegation.

PR #972, opened 2019-04-18
COSBETA-NONE: Update frequency of db logs job

raamSoftwire:
Update frequency of db logs job from every hour to every day at 1am.


raamSoftwire:
On second thought, let's not merge this just yet. There's some extra stuff that I need to think through on the Kibana side.

PR #978, opened 2019-04-22
COSBETA-329: Add tests for authorisation policies

aidayounesi:
## Description
Fixing a bug in responsible person policy
Writing tests for all of the available policies in the cosmetics app

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/spec/login_helpers.rb
```diff
@@ -11,6 +11,9 @@ def sign_in(as_user: build(:user), with_roles: [])
     allow(Keycloak::Client).to receive(:has_role?).and_return(false)
     allow(Keycloak::Client).to receive(:has_role?).with(any_of(with_roles), anything).and_return(true)
 
+    allow(Keycloak::Internal).to receive(:has_role?).and_return(false)
```
tom-bell-softwire: In theory, we shouldn't need to stub these methods in Cosmetics, because we don't check the roles of any user except the current user. It might be that we're failing to set up the current user's state correctly, which is why these methods are being called instead.

cosmetics-web/spec/policy_helpers.rb
```diff
@@ -0,0 +1,15 @@
+module PolicyHelpers
```
tom-bell-softwire: This is really great! Having a custom matcher makes these tests much more readable – we should look for other scenarios where adding our own matchers would make tests easier to follow.

PR #990, opened 2019-04-24
PSD-1119 Cookie banner

samlionheart:
Adding a cookie banner to the top of the page.

## Description
The cookie banner persists until it is dismissed, which will save a cookie for 1 year.
https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js

The `id`, `path` (link to policy), and `message` can be set when using the partial.

The banner cannot be hidden if the user does not have JavaScript enabled, which is currently the same as how it works on GOV.UK.

Updated our policy, copying over cookie policy from gov.uk:
https://www.gov.uk/help/cookies#our-introductory-message

## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


psd-web/app/views/layouts/application.html.slim
```diff
@@ -11,6 +11,7 @@ doctype html
   = link_to html, "https://www.surveymonkey.co.uk/r/TWLYX9W", target: :_blank
   |  will help us to improve it.
   - # rubocop:enable Rails/LinkToBlank
+= render "components/govuk_cookie_banner", path: help_privacy_notice_path(anchor: "cookies-policy")
```
jasiekmiko: Nice api!

shared-web/app/assets/application/javascripts/cookie_banner.js
```diff
@@ -0,0 +1,19 @@
...
+function cookieBanner() {
+  const myButton = document.getElementById('hideLink');
+  function setCookie() {
+    const d = new Date();
+    d.setTime(d.getTime() + (365 * 24 * 60 * 60 * 1000));
+    const expires = `expires=${d.toUTCString()}`;
+    document.cookie = `seen_cookie_message = true;${expires};path=/`;
+  }
+  function hideCookieBanner() {
+    setCookie(365);
```
jasiekmiko: `setCookie` doesn't take any args?

shared-web/app/assets/application/javascripts/cookie_banner.js
```diff
@@ -0,0 +1,19 @@
+import $ from 'jquery';
+
+function cookieBanner() {
+  const myButton = document.getElementById('hideLink');
+  function setCookie() {
+    const d = new Date();
+    d.setTime(d.getTime() + (365 * 24 * 60 * 60 * 1000));
```
jasiekmiko: extracting/naming variables could make this a tad more obvious to follow.

PR #991, opened 2019-04-24
COSBETA-304: Add questions for the "pH range" journey

TiagoSusanoPinto:
Add 3 pages:
- “Is the cosmetic product a hair dye?” question page:
If “Yes” goes to “What is the pH range of the cosmetic product?”
If “No” goes to “Is the pH lower than 3 or higher than 10?”

- “Is the pH lower than 3 or higher than 10?” question page:
If “Yes” goes to “What is the pH range of the cosmetic product?”
If “No” continues the manual journey

- “What is the pH range of the cosmetic product?” question page:
Update the notification pH range (ph_min_value and ph_max_value) value

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/notification_build/is_hair_dye.html.slim
```diff
@@ -0,0 +1,14 @@
+- content_for :page_title, "Is the cosmetic product a hair dye?"
+- content_for :after_header do
+  = link_to "Back", previous_wizard_path, class: "govuk-back-link"
+
+= form_with model: @notification, url: wizard_path, method: :put do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      = render "form_components/govuk_error_summary", form: form
+      = render "form_components/govuk_radios", form: form, key: :is_hair_dye,
+              fieldset: { legend: { text: "Is the cosmetic product a hair dye?",
```
jasiekmiko: I believe all of these new pages are missing an `h1` element - you should be abel to add one by specifying `isPageHeading: true` option on the fieldset legend.

cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -96,8 +135,46 @@ def render_single_or_multi_component_step
...
+    when "false"
+      render_wizard @notification
+    else
+      @notification.errors.add :is_hair_dye, "Must select an option"
+      render step
+    end
+  end
+
+  def render_is_ph_between_3_and_10_step
+    case params[:notification] && params[:notification][:is_between_3_and_10]
```
tom-bell-softwire: It would be slightly more consistent to call the parameter `is_ph_between_3_and_10` (same applies below and in the view).

cosmetics-web/app/controllers/notification_build_controller.rb
```diff
@@ -165,4 +242,8 @@ def render_add_internal_reference
     end
     render_wizard @notification
   end
+
+  def clear_notification_ph_range
```
tom-bell-softwire: I'd be tempted to just call this `clear_ph_range`

cosmetics-web/app/models/notification.rb
```diff
@@ -187,6 +191,12 @@ def mandatory_attributes(state)
       mandatory_attributes('empty')
     end
   end
+
+  def max_ph_is_greater_than_min_ph
```
tom-bell-softwire: Nice!

cosmetics-web/app/views/notification_build/is_mixed.html.slim
```diff
@@ -10,4 +10,4 @@
               fieldset: { legend: { text: "Does the kit need to be mixed?", classes: "govuk-fieldset__legend--xl" } },
               items: [{ text: "Yes, the items need to be mixed", value: "true" },
                       { text: "No, the items are used sequentially", value: "false" }]
-      = form.submit "Continue", class: "govuk-button"
+  = form.submit "Continue", class: "govuk-button"
```
tom-bell-softwire: As with my comment on other reviews, is it intentional to move the submit button outside of the grid elements?

TiagoSusanoPinto: It was somewhat intentional, since it fell the right way to do it, and I had an idea that's how we used to add it.
Just noticed now that we don't do that anywhere else, and I probably just got confused with a different element :)

cosmetics-web/app/views/notification_build/is_hair_dye.html.slim
TiagoSusanoPinto: Didn't knew about it, thanks!

TiagoSusanoPinto: Didn't knew about it, thanks! :)

PR #994, opened 2019-04-24
COSBETA-159: Verify contact person email address

aidayounesi:
## Description
After entering contact person details, show the existing email verification page

Mark the contact person as verified if their email address matches the current user’s email address

When signing in – if you have no RPs with verified contact persons, then you are redirected to the “verify your email address” page

Optional row showing “unverified” status on contact person list with resend email link (Look at relevant pages on PSD for related designs)

NOTE: Behaviour for multiple RPs TBD – covered by COSBETA‌-43

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/models/contact_person.rb
```diff
@@ -4,4 +4,12 @@ class ContactPerson < ApplicationRecord
   validates :name, presence: true
   validates :email_address, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
   validates :phone_number, presence: true
+
+  before_create :current_user_is_verified
+
+private
+
+  def current_user_is_verified
+    self.is_email_verified = true if email_address == User.current&.email
```
TiagoSusanoPinto: I've might have missed it, but should we add a test to cover this?

aidayounesi: You are right, thanks. I am going to write two tests for it:
1. Model test to test this unit
2. A system test to make sure that if the current user email is entered as the contact person's email then we would be redirected to dashboard (instead of confirm  your email address)

cosmetics-web/db/migrate/20190424132227_add_is_email_verified_to_contact_persons.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsEmailVerifiedToContactPersons < ActiveRecord::Migration[5.2]
+  def change
+    add_column :contact_persons, :is_email_verified, :boolean
```
tom-bell-softwire: Should this default to false?

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
```diff
@@ -23,4 +23,4 @@
     = render "notifications/frame_formulation_details", notification: @notification
     = render "notifications/substances_details", notification: @notification
     = render "/notifications/responsible_person_details", responsible_person: @notification.responsible_person
-    = render "/notifications/contact_person_details", contact_person: @notification.responsible_person
+    = render "/notifications/contact_person_details", contact_person: @notification.responsible_person.contact_persons.first
```
tom-bell-softwire: Thanks for fixing this!

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
```diff
@@ -10,7 +10,7 @@ def show
 )
 
     unless email_verification_key.is_expired?
-      email_verification_key.responsible_person.update(is_email_verified: true)
+      email_verification_key.responsible_person.contact_persons.first.update(is_email_verified: true)
```
tom-bell-softwire: I think it would make more sense for the EmailVerificationKey to belong to the ContactPerson directly, rather than to the ResponsiblePerson.

tom-bell-softwire: Relatedly, I wonder if we should move the route to be something like:
`/responsible_persons/rp_id/contact_persons/cp_id/verify/`

This would make it easier to handle multiple contact persons in the future; however, I'm happy to leave that change until we actually implement multiple contact persons, if that's easier.

cosmetics-web/db/migrate/20190424132227_add_is_email_verified_to_contact_persons.rb
tom-bell-softwire: It would be good to name this column something like `email_verified` – that way, we can use `contact_person.email_verified?` in the code, since ActiveRecord gives you the `?` method for free on boolean properties.

aidayounesi: Thanks

aidayounesi: Thanks, done

cosmetics-web/app/controllers/responsible_persons/verification_controller.rb
aidayounesi: It was fixed as part of another ticket. Reminding: contact person email verification does not need to be authorized

cosmetics-web/app/models/email_verification_key.rb
```diff
@@ -12,10 +12,10 @@ def is_expired?
     expires_at < DateTime.current
   end
 
-  def self.verify_key_for_responsible_person(responsible_person, key)
+  def self.verify_key_for_contact_person(contact_person, key)
```
tom-bell-softwire: Is this method used any more?

cosmetics-web/app/models/contact_person.rb
```diff
@@ -1,7 +1,17 @@
 class ContactPerson < ApplicationRecord
   belongs_to :responsible_person
 
+  has_one :email_verification_keys, dependent: :destroy
```
tom-bell-softwire: Previously this was a `has_many` association – is that not longer the case?

cosmetics-web/app/views/poison_centres/notifications/show_poison_centre.html.slim
```diff
@@ -23,4 +23,5 @@
     = render "notifications/frame_formulation_details", notification: @notification
     = render "notifications/substances_details", notification: @notification
     = render "/notifications/responsible_person_details", responsible_person: @notification.responsible_person
-    = render "/notifications/contact_person_details", contact_person: @notification.responsible_person
+    - @notification.responsible_person.contact_persons.each do |contact_person|
```
tom-bell-softwire: Should we also update the Responsible Person `show` page to:
* show each contact person
* indicate the verified status of their email address
* include a link to resend the verification email, if not verified?

cosmetics-web/app/models/email_verification_key.rb
aidayounesi: Not anymore

cosmetics-web/app/views/notifications/_contact_person_details.html.slim
```diff
@@ -1,19 +1,31 @@
...
       td.govuk-table__cell
         = contact_person.name
+      td.govuk-table__cell
     tr.govuk-table__row
       th.govuk-table__header[scope="row"]
         | Email address
       td.govuk-table__cell
         = contact_person.email_address
+      td.govuk-table__cell
+        -if contact_person.email_verified? || User.current&.poison_centre_user? || User.current&.msa_user?
```
tom-bell-softwire: It would be good to avoid having to check the current user type here.

In order for a notification to be submitted, I believe the contact person must have been verified first, so this should always be true for notifications viewed by Poison Centre or MSA users.

Alternatively, you could pass in a parameter to this partial, similarly to the `allow_edits` parameter that is passed in to the `product_details.html.slim` partial.

cosmetics-web/app/views/notifications/_contact_persons_list.html.slim
```diff
@@ -0,0 +1,4 @@
+h1.govuk-heading-m
```
tom-bell-softwire: Is there no way to use a table caption here for consistency with the rest of the page sections?

If not, this should probably be an `h2`, since it is semantically a subheading under the main page heading.

cosmetics-web/app/mailers/notify_mailer.rb
```diff
@@ -1,20 +1,20 @@
 class NotifyMailer < GovukNotifyRails::Mailer
-  def send_responsible_person_verification_email(responsible_person_id, email_address, contact_person_name, responsible_person_name, user_name)
+  def send_contact_person_verification_email(contact_person, responsible_person_name, user_name)
```
tom-bell-softwire: It's generally a bad idea to pass full ActiveRecord models in as parameters to mailer methods.

Remember that these method calls will be saved as worker jobs to be executed later, so the arguments in the method call need to be serialised and saved as part of the job on the Redis queue. Serialising an entire ActiveRecord model is complicated, and it frequently fails to be deserialised correctly (which is the cause of the "failed to deserialise..." errors we sometimes see in the worker logs).

It's therefore better to pass in the necessary properties of the model instead (`id`, `email_address`, `name`).

cosmetics-web/config/routes.rb
```diff
@@ -50,6 +50,8 @@
     end
   end
 
+  resources :email_verification_keys, path: "verify", controller: "verification", param: :key, only: %i[show index]
```
tom-bell-softwire: Is this different from the `confirmation` resource added below as part of the previous PR?
Could these be combined with those existing routes under `/contact_persons/confirm` ?

cosmetics-web/app/views/notifications/_contact_person_details.html.slim
aidayounesi: You are right maybe I overthought about the possibility of changing a contact person and causing email_verified to be false.
I am going to use your alternative solution to pass ```allow_edit```.
Thanks.

cosmetics-web/app/mailers/notify_mailer.rb
aidayounesi: Thank you for fully describing this matter.

cosmetics-web/app/views/notifications/_contact_persons_list.html.slim
aidayounesi: As we discussed, we are going to have 2 different partial for contact person details 
1) contact_person_list (h2) that includes many contact_person_details (description list) -> used in RP dashboard page
2) contact_person_overview (table) -> used in poison centre and notification show/edit page

cosmetics-web/config/routes.rb
aidayounesi: It's same, thanks for mentioning. Removed

PR #1006, opened 2019-04-25
PSD-1132: Remove phase banner and add badge

samlionheart:
Removing the beta phase banner underneath the header, and adding a BETA badge next to the service name
## Description

Removing beta phase banner in application.html - There is no need for the feedback link to be a capture anymore as it is only used once in the footer.
Adding a "phase banner" option to the psd_header component.
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen


psd-web/app/views/layouts/application.html.slim
```diff
@@ -87,7 +69,13 @@ doctype html
              = link_to html, "https://www.gov.uk/government/organisations/office-for-product-safety-and-standards", target: :_blank
            .opss-footer__body
              br
-             = feedback
+              ' This is a new service – your
+              - html = capture do
+                | feedback
+                span.govuk-visually-hidden
+                  | (opens new window)
+              = link_to html, "https://www.surveymonkey.co.uk/r/TWLYX9W", target: :_blank
```
tom-bell-softwire: I think this needs `rel: 'noopener'` since we are linking to a site outside of the service.

psd-web/app/views/layouts/application.html.slim
```diff
@@ -87,7 +69,13 @@ doctype html
              = link_to html, "https://www.gov.uk/government/organisations/office-for-product-safety-and-standards", target: :_blank
            .opss-footer__body
              br
-             = feedback
+              ' This is a new service – your
```
tom-bell-softwire: I think the indentation is off here (and on the following lines below) – should it be in line with the `br` element above?

PR #1009, opened 2019-04-26
COSBETA-205: Add question for the product's physical form

TiagoSusanoPinto:
Add question for the product's physical form

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/views/component_build/add_physical_form.html.slim
+++ b/cosmetics-web/app/views/component_build/add_physical_form.html.slim
@@ -1,10 +1,6 @@
 - content_for :page_title, "Single or multi component"
 - content_for :after_header do
-  - previous_step = @component.shades.nil? ? :number_of_shades : :add_shades
-  = link_to "Back",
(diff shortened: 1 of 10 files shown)


cosmetics-web/app/views/component_build/add_physical_form.html.slim
```diff
@@ -0,0 +1,19 @@
+- content_for :page_title, "Single or multi component"
+- content_for :after_header do
+  - previous_step = @component.shades.nil? ? :number_of_shades : :add_shades
+  = link_to "Back",
```
tom-bell-softwire: Would it be worth overriding `previous_wizard_path` on the controller, in the same way that you've done for the other build controllers?

TiagoSusanoPinto: Good call, done! :)

PR #1012, opened 2019-04-29
PSD-1145: Product search

samlionheart:
Re-adding product search
## Description

Re-adding product search and removing case count on business and product lists.
## Checklist:

### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [x] Works keyboard only


tom-bell-softwire:
Thanks!

PR #1023, opened 2019-05-02
COSBETA-245 Improvements for large drop down select questions

TiagoSusanoPinto:
* Import js file on comestics shared file in order to include autocomplete function
* Add validation to select frame formulation page
* Show per category level on select category page
* Add error summaries to select_category page and select_frame_formulation

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/component_build_controller.rb
@@ -228,8 +228,11 @@ private
   end
 
   def render_select_frame_formulation
-    @component.update(component_params)
-    redirect_to finish_wizard_path
+    if @component.update_with_context(component_params, step)
+      redirect_to finish_wizard_path
+    else
+      render step
+    end
   end
 
   def render_upload_formulation
--- a/cosmetics-web/app/controllers/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/component_build_controller.rb
@@ -16,6 +16,7 @@ class ComponentBuildController < ApplicationController
         :upload_formulation
 
   before_action :set_component
+  before_action :set_category, if: -> { step == :select_category }
 
   def show
     @component.shades = ['', ''] if step == :add_shades && @component.shades.nil?
@@ -32,12 +33,14 @@ class ComponentBuildController < ApplicationController
       render_number_of_shades
     when :add_shades
       render_add_shades
+    when :add_cmrs
+      render_add_cmrs
     when :contains_nanomaterials
       render_contains_nanomaterials
     when :add_nanomaterial
       render_add_nanomaterial
-    when :add_cmrs
-      render_add_cmrs
+    when :select_category
+      render_select_category_step
     when :select_formulation_type
       render_select_formulation_type
     when :select_frame_formulation
@@ -68,6 +71,8 @@ class ComponentBuildController < ApplicationController
       responsible_person_notification_build_path(@component.notification.responsible_person, @component.notification, :add_new_component)
     elsif step == :number_of_shades && !@component.notification.is_multicomponent?
       responsible_person_notification_build_path(@component.notification.responsible_person, @component.notification, :single_or_multi_component)
+    elsif step == :select_category && @category.present?
+      wizard_path(:select_category, category: Component.get_parent_category(@category))
     elsif previous_step.present?
       responsible_person_notification_component_build_path(@component.notification.responsible_person, @component.notification, @component, previous_step)
     else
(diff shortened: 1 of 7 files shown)


cosmetics-web/app/views/component_build/select_category.html.slim
```diff
@@ -1,18 +1,23 @@
-- content_for :page_title, "Select the component category category"
+ruby:
+  question =
+    if @category.present?
+      "What category of #{get_category_name(@category).downcase} is the product?"
```
tom-bell-softwire: Should the title include the component name if it's a multi-component product?

cosmetics-web/app/views/component_build/select_category.html.slim
```diff
@@ -1,18 +1,23 @@
...
+    else
+      "What category of cosmetic product is it?"
+    end
+
+- content_for :page_title, "Select sub category"
```
tom-bell-softwire: Should the page title also change depending on whether it is selecting a category or sub-category?

cosmetics-web/app/views/component_build/select_category.html.slim
TiagoSusanoPinto: This will be covered by COSBETA-243

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -212,6 +229,22 @@ def render_add_cmrs
     render_wizard @component
   end
 
+  def render_select_category_step
+    # @component.update(sub_sub_category: nil)
```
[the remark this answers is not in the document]
TiagoSusanoPinto: good catch! :)

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -100,6 +105,18 @@ def component_params
...
+  def set_category
+    @category = params[:category]
+    if @category.present? && !has_sub_categories(@category)
+      @component.errors.add :sub_category, "Select a valid option"
+      @category = nil
+    end
+    @sub_categories = @category.present? ? get_sub_categories(@category) : get_main_categories
+    @sub_categories.keys.each do |key|
+      @selected_sub_category = key if sub_category_belongs_to_category(@component.sub_sub_category, key)
+    end
```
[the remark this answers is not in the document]
TiagoSusanoPinto: There's something on this method's name (sub_category_belongs_to_category) that sounds wrong/fishy, but not sure how to improve it

cosmetics-web/app/views/component_build/select_category.html.slim
TiagoSusanoPinto: done! :)

cosmetics-web/app/controllers/component_build_controller.rb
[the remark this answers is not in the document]
tom-bell-softwire: Yeah, I think you're right, though I'm not sure how to rename it...

It would be nice if we could somehow convert these enums into objects and then have something like: ```@component.sub_sub_category.belongs_to?(category)``` but I don't know if Ruby/Rails supports anything like that.

Alternatively, would the following code be more readable?
```ruby
@selected_sub_category = @sub_categories.find { |category| category == @component.sub_category || category == @component.sub_sub_category }
```

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -100,6 +105,22 @@ def component_params
...
+    @category = params[:category]
+    if @category.present? && !has_sub_categories(@category)
+      @component.errors.add :sub_category, "Select a valid option"
+      @category = nil
+    end
+    @sub_categories = @category.present? ? get_sub_categories(@category) : get_main_categories
+    p @component.root_category
+    p @component.sub_category
+    @selected_sub_category = @sub_categories.keys.find do |category|
+      category.to_sym == @component.root_category.to_sym ||
```
tom-bell-softwire: Are the `.to_sym` calls needed here ? I think both the `@sub_category` keys and `@component.root_category` should already be symbols.

cosmetics-web/app/helpers/category_helper.rb
```diff
@@ -10,8 +10,23 @@ def get_full_category_name(sub_sub_category)
...
 
-  def get_sub_sub_categories
+  def get_main_categories
     parent_of_categories = Component.get_parent_of_categories
-    Component.categories.reject { |key| parent_of_categories.has_value?(key.to_sym) }
+    Component.categories.reject { |category| parent_of_categories[category.to_sym].present? }
+  end
+
+  def get_sub_categories category
+    Component.get_parent_of_categories.select { |_key, value| value == category.to_sym }
```
tom-bell-softwire: Would it be neater to just return a list of symbols from this method, rather than a hash of key-value pairs? It looks like every place that we use this method, we only care about the keys anyway, so this might make the rest of that code a bit cleaner.

TiagoSusanoPinto: Done, changed both this function and the `get_main_categories` to return only a list of keys (as symbols)

cosmetics-web/app/controllers/component_build_controller.rb
TiagoSusanoPinto: The `sub_categories` weren't always a symbol! But I've already changed that.
Also, as discussed, I've extracted these comparisons into the Component model itself

tom-bell-softwire: Great!

PR #1050, opened 2019-05-14
COSBETA-255: Add validation on pH based on the ph range selected

TiagoSusanoPinto:
BUG FIX:
COSBETA-255: Add validation on pH based on the ph range selected

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/trigger_questions/select_ph_range.html.slim
```diff
@@ -1,15 +1,6 @@
...
-      = render "form_components/govuk_error_summary", form: form
-      = render "form_components/govuk_radios", form: form, key: :range,
-              fieldset: { legend: { text: "What is the pH of #{@component_name}?", \
-                                    classes: "govuk-fieldset__legend--xl", isPageHeading: true } },
-              items: [{ text: "Below 3", value: "below" },
-                      { text: "Between 3 and 10", value: "between" },
-                      { text: "Above 10", value: "above" }]
-      = form.submit "Continue", class: "govuk-button"
+= render "trigger_questions/substance_check", substance: "ph_range",
+        question: "Is the pH between 3 and 10?", yes_option: "Yes"
```
tom-bell-softwire: Shouldn't the question be the inverse of this – selecting "yes" will mark the question as applicable, but actually "no" is the applicable option here.

Perhaps we can just reverse the question ("Is the pH below 3 or above 10?").

Would it make sense to use the `substance_check_with_condition` template, so that the user is immediately prompted for the pH value if they select "yes"? Or is that going to be tricky to implement?

PR #1051, opened 2019-05-14
COSBETA-198: Update contact person email address

aidayounesi:
## Description
On the “check your email” page displayed after a user creates a Responsible Person, add a link to allow the user to edit the contact person details 

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/responsible_persons/contact_persons/_form.slim
```diff
@@ -0,0 +1,22 @@
+= form_with model: @contact_person, url: url, method: method do |form|
+  .govuk-grid-row
+    .govuk-grid-column-two-thirds
+      = render "form_components/govuk_error_summary", form: form
+
+      h1.govuk-heading-xl
```
tom-bell-softwire: I think the heading and paragraphs here may need to be different for the new/edit pages. Probably worth checking with Ralph.

If so, you could pass in the relevant HTML as a block to this partial view, similarly to the way we pass in blocks of HTML to some of the form components.

cosmetics-web/spec/controllers/responsible_persons/contact_persons_controller_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'rails_helper'
+
+RSpec.describe ResponsiblePersons::ContactPersonsController, type: :controller do
+  let(:email_verification_key) { create(:email_verification_key) }
+  let(:responsible_person) { create(:responsible_person) }
+  let(:contact_person) { responsible_person.contact_persons.first }
+  let(:params) do
```
tom-bell-softwire: It's slightly neater to structure this in the same way as in `component_build_controller_spec.rb`

It might also make the tests a bit clearer to have two versions of this called `valid_params` and `invalid_params` instead.

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -34,11 +34,15 @@ def has_accepted_declaration
   def create_or_join_responsible_person
     return unless user_signed_in? && !poison_centre_or_msa_user?
 
+    responsible_person = User.current.responsible_persons.first
+    contact_person = User.current.responsible_persons.first&.contact_persons&.first
+
     if User.current.responsible_persons.empty?
```
TiagoSusanoPinto: What `User.current.responsible_persons.first` returns if `User.current.responsible_persons` is empty?

Could we use the variable `responsible_person` here as well instead?

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -34,11 +34,15 @@ def has_accepted_declaration
   def create_or_join_responsible_person
     return unless user_signed_in? && !poison_centre_or_msa_user?
 
+    responsible_person = User.current.responsible_persons.first
+    contact_person = User.current.responsible_persons.first&.contact_persons&.first
```
TiagoSusanoPinto: It feels like `contact_person` should be defined inside the if case were it's used, and using the variable `responsible_person` instead of doing `User.current.responsible_persons.first` again.

But not too fussed about it 🙂

aidayounesi: Sure, thanks 🙂

cosmetics-web/spec/controllers/responsible_persons/contact_persons_controller_spec.rb
aidayounesi: Sure. Something like this:
```
  let(:valid_params) {
    {
      responsible_person_id: responsible_person.id,
      contact_person:
      {
        name: contact_person.name,
        email_address: contact_person.email_address,
        phone_number: contact_person.phone_number
      }
    }
  }
  let(:invalid_params) {valid_params.merge(contact_person: { email_address: "" })}
```

cosmetics-web/app/controllers/application_controller.rb
aidayounesi: It would be nil, we can use the variable and blank?
```
    responsible_person = User.current.responsible_persons.first

    if responsible_person.blank?

....
```

cosmetics-web/app/views/responsible_persons/contact_persons/show.html.slim
```diff
@@ -8,14 +8,18 @@
       ' We have sent an email to
       strong = @contact_person.email_address
     p
-      ' The contact person will need to click on the link in the email to confirm that they are
+      ' Wrong email address?
+      = link_to "Change the contact person details",
+              edit_responsible_person_contact_person_path(@responsible_person),
```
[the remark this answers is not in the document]
aidayounesi: Sure, thanks.

PR #1059, opened 2019-05-15
COSBETA-190: Content updates for multi-item kit question

tom-bell-softwire:


ralph-hawkins:
Looks like this fell out of another pull request, but it should be added.

Changes:
- fixes a typo on the question
- adds clarity about what a single item product is

PR #1079, opened 2019-05-21
COSBETA-410: Stop creating empty trigger question rules

TiagoSusanoPinto:
COSBETA-410: Stop creating empty trigger question rules
when starting the trigger questions rules on the manual journey

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/trigger_questions_controller.rb
+++ b/cosmetics-web/app/controllers/trigger_questions_controller.rb
@@ -26,7 +26,7 @@ class TriggerQuestionsController < ApplicationController
         :contains_isopropanol
 
   before_action :set_component
-  before_action :set_question
+  before_action :set_question, only: %i[show update]
 
   def show
     initialize_step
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/controllers/trigger_questions_controller.rb
```diff
@@ -159,7 +159,7 @@ def set_component
 
   def set_question
     question = get_question_for_step step
-    @question = TriggerQuestion.find_or_create_by(component: @component, question: question)
+    @question = TriggerQuestion.find_or_create_by(component: @component, question: question) if question.present?
```
TiagoSusanoPinto: Even though the previous change fix the bug (since the empty trigger question was being created when running the new function). I added this for precaution, on the case that we want some step/page without being associated to any trigger question

PR #1080, opened 2019-05-21
COSBETA-198: Ensure existing validation key is deleted when updating contact person email address

aidayounesi:
fixing the bug during sprint review session

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
```diff
@@ -19,14 +19,15 @@ def edit; end
 
   def update
     if @contact_person.update(contact_person_params)
+      remove_contact_person_email_verification_key
```
tom-bell-softwire: Would it be more consistent to remove any existing verification keys at the start of the `send_verification_email` method instead, so that the logic is all in one place?

Or perhaps move that line to the `NotifyMailer.send_contact_person_verification_email` method?

aidayounesi: Sure, I moved it to NotifyMailer.send_contact_person_verification_email because it feels part of that logic.

PR #1088, opened 2019-05-23
COSBETA-365: Update CMR questions

aidayounesi:
## Description
Split the current CMR page into 2 different pages.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/helpers/component_build_helper.rb
```diff
@@ -0,0 +1,11 @@
+module ComponentBuildHelper
+  def cmr_errors(component)
+    result = []
+    component.cmrs.each_with_index do |cmr, index|
+      cmr.errors.each do |attr, msg|
+        result.push(text: msg, href: "#component_cmrs_attributes_#{index}_#{attr}")if cmr.errors.any?
```
TiagoSusanoPinto: do you need the `if cmr.errors.any?` ?
I might be missing something, but if there's no cmr.errors, you will never enter in the for loop :)

TiagoSusanoPinto: do you need the if `cmr.errors.any?` ? :)

TiagoSusanoPinto: Also, this isn't a very immutable approach.
You should be able to get the same result with a flat_map.with_index (for the outer loop) and a map

cosmetics-web/app/models/cmr.rb
```diff
@@ -1,7 +1,22 @@
 class Cmr < ApplicationRecord
   belongs_to :component
 
+  CMR_NUMBER_REGEX = /\A(\d+-)*\d+\z/.freeze
+
+  validates :name, presence: true
+  validates :cas_number, format: { with: CMR_NUMBER_REGEX }, allow_blank: true
+  validates :ec_number, format: { with: CMR_NUMBER_REGEX }, allow_blank: true
```
tom-bell-softwire: As discussed, we should make the validation more strict by imposing min/max number of digits as well.

cosmetics-web/app/views/component_build/add_cmrs.html.slim
```diff
@@ -2,33 +2,32 @@
...
-      | contain substances that are known or presumed to be carcinogenic, mutagenic or toxic for reproduction?
-    p
-      | These substance are commonly referred to as CMR.
-    p
-      = link_to "View carcinogenic, mutagenic or toxic for reproduction substances (opens in a new tab).",
-              cmr_list_url, target: :_blank, rel: 'noopener'
-    = form_for(@component, url: wizard_path, method: :put) do |form|
-      - @component.cmrs.each_with_index do |cmr, index|
-        - cmr_number = index + 1
+- cmr_errors = cmr_errors(@component)
```
tom-bell-softwire: This is neat!

cosmetics-web/app/views/component_build/add_cmrs.html.slim
```diff
@@ -2,33 +2,32 @@
...
-                label: { text: "Substance #{cmr_number} name" }
-        = render "components/govuk_input",
-                name: "cmrs[#{index}][cas_number]",
-                id: "cmrs-cas_number",
-                value: cmr.cas_number,
-                label: { text: "Substance #{cmr_number} Chemical Abstracts Service registry number" }
-      .govuk-form-group
-          = form.submit "Continue", class: "govuk-button"
+= form_with model: @component, url: wizard_path, method: :put do |form|
+    .govuk-grid-row
```
tom-bell-softwire: I think the indentation might be off here (should be 2 instead of 4 spaces)...

cosmetics-web/app/views/component_build/contains_cmrs.html.slim
```diff
@@ -0,0 +1,29 @@
...
+- cmr_list_url = "https://eur-lex.europa.eu/legal-content/en/TXT/PDF/?uri=CELEX:02008R1272-20180301&from=EN"
+- title_text = "Does #{@component_name} contain substances listed as Category 1A or 1B mutagens,
+ carcinogens or reproductive toxicants (CMRs)?"
+- hint_html = capture do
+  p
+    | Category 1A substances are ‘known’ CMRs.
+  p
+    | Category 1B substances are ‘presumed’ CMRs.
+  p
+    =<> link_to "The classification, labelling and packaging Regulation (opens in a new tab)",
```
tom-bell-softwire: Should this be "Regulation" or "regulation" ?

tom-bell-softwire: Since there doesn't need to be whitespace before the link, this should be `=> link_to ...`

cosmetics-web/app/views/component_build/contains_cmrs.html.slim
```diff
@@ -0,0 +1,29 @@
...
+            cmr_list_url, target: :_blank, rel: 'noopener'
+    | contains definitions of category 1A and 1B mutagens (table 3.5.1),
+            carcinogens (table 3.6.1) and reproductive toxicants (table 3.7.1).
```
tom-bell-softwire: Indentation is off here – "carcinogens" should be aligned with "contains"

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -124,7 +126,9 @@ def component_params
         :sub_sub_category,
         :notification_type,
         :frame_formulation,
+        :contains_cmrs,
```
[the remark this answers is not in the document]
aidayounesi: Sure, thanks

cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -298,10 +313,36 @@ def get_previous_step
...
+  def create_required_cmrs
+    if @component.cmrs.size < NUMBER_OF_CMRS
+      cmrs_needed = NUMBER_OF_CMRS - @component.cmrs.size
+      cmrs_needed.times { @component.cmrs.build }
+    end
+  end
+
+  def component_params_without_empty_cmrs
+    filtered_params = component_params
+    component_params[:cmrs_attributes].each do |index, cmr_params|
```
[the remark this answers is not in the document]
aidayounesi: Great, the solution in the link is really neat and also reduces the size of our controller
Adding ``reject_if: proc { |attributes| attributes['name'].blank? }`` to ``accepts_nested_attributes_for :cmrs``

cosmetics-web/app/controllers/concerns/manual_notification_concern.rb
```diff
@@ -0,0 +1,29 @@
...
+  def notification
+    if params[:notification_id]
+      return Notification.find(params[:notification_id])
+    elsif params[:component_id]
+      component = Component.find(params[:component_id])
+      return component.notification
+    end
+  end
+
+  def pre_eu_exit?(step = @step)
```
tom-bell-softwire: Would it be clearer to call this method `skip_step?(step = @step)` ?

For now, the only logic here is based on pre-/post-EU notification state, but I can see this growing to include more conditions in future.

cosmetics-web/app/models/cmr.rb
```diff
@@ -19,4 +19,16 @@ def remove_hyphens
     cas_number&.delete!("-")
     ec_number&.delete!("-")
   end
+
+  def cas_number_length
+    unless cas_number.blank? || cas_number.delete("^0-9").length.between?(5, 10)
```
tom-bell-softwire: Could this just be `delete("-")`? (Any other characters will be rejected by the RegEx format validation above.)

cosmetics-web/app/models/cmr.rb
```diff
@@ -19,4 +19,16 @@ def remove_hyphens
...
+      errors.add(:cas_number, "CAS number must contain between 5 to 10 digits")
+    end
+  end
+
+  def ec_number_length
+    unless ec_number.blank? || ec_number.delete("^0-9").length == 7
```
tom-bell-softwire: Again, could this just be `delete("-")`?

cosmetics-web/spec/models/cmr_spec.rb
```diff
@@ -10,6 +10,20 @@
...
   end
 
+  it "fails if ec number doesn't contain 7 digits" do
+    cmr.ec_number = "123"
+
+    expect(cmr.save).to be false
+    expect(cmr.errors[:ec_number]).to include("EC number must contain 7 digits")
+  end
+
+  it "fails if cas number doesn't contain 5-10 digits" do
```
tom-bell-softwire: Is it worth having separate tests for "CAS number contains less than 5 digits" and "CAS number contains more than 10 digits".

tom-bell-softwire: You might need to add `cas_number` and `ec_number` to the project's `en.yml` languages file to specify the correct capitalisation of "CAS" and "EC" (as an example of how to do this, see the existing entries for `responsible_person`).

cosmetics-web/app/models/cmr.rb
tom-bell-softwire: As discussed, this is still needed even if the above validation catches non-numeric digits, since both validation methods are called and we want to catch the case where the number of digits is incorrect in the presence of other, non-numeric characters (e.g. "123ab-34-6").

cosmetics-web/app/models/cmr.rb
aidayounesi: Same as the comment before.

tom-bell-softwire: Same comment as above.

PR #1089, opened 2019-05-23
COSBETA-NONE: Add consistency when start the manual journey

TiagoSusanoPinto:
At register_on_eu_system define notified_before_eu_exit attribute
when starting a new manual notification journey

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


TiagoSusanoPinto:
I reckon we should have some tests (probably system tests) to track the different paths when adding a new notification (when selecting if using the bulk upload journey, or the manual journey)

PR #1108, opened 2019-05-29
COSBETA-373: Update CMR details

aidayounesi:
Update CMR details in check your answer page.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/models/cmr.rb
```diff
@@ -2,6 +2,14 @@ class Cmr < ApplicationRecord
   belongs_to :component
 
   def display_name
```
tom-bell-softwire: Is is worth adding a couple of model tests for this `display_name` method?

cosmetics-web/app/models/cmr.rb
```diff
@@ -2,6 +2,14 @@ class Cmr < ApplicationRecord
...
-    [name, ec_number, cas_number].reject(&:blank?).join(', ')
+    [name, display_cas_number, display_ec_number].reject(&:blank?).join(', ')
+  end
+
+  def display_cas_number
+    cas_number.dup.insert(-2, "-").insert(-5, "-") if cas_number.present?
+  end
+
+  def display_ec_number
+    ec_number.dup.insert(3, "-").insert(7, "-") if ec_number.present?
```
tom-bell-softwire: For consistency with the above method, could these also be negative indices from the end of the string?

PR #1109, opened 2019-05-29
COSBETA-11: Open link to nanomaterial regulations in a new window

aidayounesi:
This covers the journey where all nanomaterials are used in accordance with the relevant annex for their intended use

(Open Annex link in a new tab)

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


tom-bell-softwire:
Thanks!

PR #1132, opened 2019-06-05
PSD-1168 Search filters case type

samlionheart:
Case type filters, checkboxes for allegation, enquiry and project.
## Description

Checkboxes for case types. When no checkboxes are selected it shows all cases, when all checkboxes are selected it shows all cases.
## Checklist:


CHANGELOG.md
```diff
@@ -3,10 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 <!-- ### Product safety database -->
-- New service navigation
-- Welcome email
-- Product search
-- Cookie banner
```
tom-bell-softwire: I don't think these changes have been released yet, so worth keeping them here.

psd-web/app/helpers/investigations_helper.rb
```diff
@@ -34,6 +35,17 @@ def get_status_filter
     { must: { term: status } }
   end
 
+  def get_type_filter
+    return {} if params[:allegation] == "unchecked" && params[:enquiry] == "unchecked" && params[:project] == "unchecked"
+
+    array_of_types = []
```
tom-bell-softwire: You could probably just call this `types` (the `array_of_` is a bit redundant).

samlionheart:
Thanks for the comments Tom.

Couple of things I missed:

The filters for assignee and case type were "should", they actually need to be "must". I've changed them all to "must" as they should be AND filters. Because they are all "must", when merged they need to add together instead of overwrite each other.

tom-bell-softwire:
Thanks @samlionheart ! How does the AND filter requirement work within a given filter – i.e., if I select both "allegation" and "enquiry", that should be an OR query, right?

PR #1144, opened 2019-06-11
PSD-1167 Created by filter and case type filter

samlionheart:
"Created by" filter, same UI as "Assigned" filter, merged with case type filter

## Description

Uses source.user_id to search for creator in the same way Assignee filter works. A lot of the code for the logic of the checkboxes is copied from assignee and altered to match param names.
## Checklist:


psd-web/app/views/investigations/index.html.slim
```diff
@@ -27,19 +27,41 @@
           = render "form_components/govuk_select", key: :assigned_to_someone_else_id, form: form,
                         items: entities.map { |e| { text: e.display_name, value: e.id } },
                         label: { text: "Name" }, is_autocomplete: true
+        - other_creator = capture do
```
doctorpod: This could be a helper: other_creator

psd-web/app/views/investigations/index.html.slim
```diff
@@ -27,19 +27,41 @@
...
                            conditional: { html: other_assignee } }
+          created_by_items = [{ key: "created_by_me", value: "checked", unchecked_value: "unchecked", text: "Me" }]
+          creator_teams_with_keys.each do |key, team, name|
+              created_by_items << { key: key, value: team.id, unchecked_value: "unchecked", text: name }
+          end
+          created_by_items << { key: "created_by_someone_else",
+                           value: "checked",
+                           unchecked_value: "unchecked",
+                           text: "Other person or team",
+                           conditional: { html: other_creator } }
```
doctorpod: Same here - big blocks of native ruby would indicate a possible helper

PR #1147, opened 2019-06-12
COSBETA-241: Redirect to business or Poison Centre pages based on domain

tom-bell-softwire:
Business users and Poison Centre/MSA users will access the service via different domains. All user groups will use the same service (Keycloak) to log in to their accounts and should be correctly redirected back to the domain from which they started the login journey.


cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -16,12 +16,17 @@ class ApplicationController < ActionController::Base
   before_action :has_accepted_declaration
   before_action :create_or_join_responsible_person
 
+  helper_method :search_domain_url, :submit_domain_url
+
```
aidayounesi: It probably worth to move them to a helper.

keycloak/configuration/initial-setup.json
```diff
@@ -2250,7 +2250,7 @@
     "clientAuthenticatorType" : "client-secret",
     "secret" : "bb1c1dde-7b7a-488a-9061-8090b59857ef",
     "defaultRoles" : [ "user" ],
-    "redirectUris" : [ "http://localhost:3002/*", "http://localhost:3003/*" ],
+    "redirectUris" : [ "http://localhost:3002/*", "http://localhost:3003/*", "http://submit_cosmetics:3002/*", "http://search_cosmetics:3002/*"],
```
aidayounesi: Do you mind updating the readme file?

cosmetics-web/spec/system/create_responsible_person_spec.rb
```diff
@@ -6,6 +6,7 @@
   let(:business_responsible_person) { create(:business_responsible_person) }
 
   before do
+    configure_requests_for_submit_domain
```
aidayounesi: It may worth googling how we can mock something for a specific sort of test

cosmetics-web/app/controllers/application_controller.rb
tom-bell-softwire: Good idea – I've split these out to a separate `DomainHelper` module.

keycloak/configuration/initial-setup.json
tom-bell-softwire: I've added instructions to the cosmetics-web README. Let me know if you think they're unclear.

cosmetics-web/spec/system/create_responsible_person_spec.rb
tom-bell-softwire: Great idea! I've raised a [tech debt ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-463) to look into this, as I think there's potential to pull out a bunch of repeated mock setup calls into global `before` and `after` blocks, which could remove a lot of boilerplate from our tests.

cosmetics-web/app/controllers/application_controller.rb
aidayounesi: Thanks

keycloak/configuration/initial-setup.json
aidayounesi: Thanks

cosmetics-web/spec/system/create_responsible_person_spec.rb
aidayounesi: Great, thanks

PR #1157, opened 2019-06-16
COSBETA-442: Allow multiple purposes for nanomaterials

tom-bell-softwire:
## Description
Currently, the user is only allowed to select one purpose per nanomaterial. This needs to be updated to reflect the behaviour in CPNP, which allows for multiple purposes (“functions”) per nanomaterial.

For example, a nanomaterial could be both a colourant **and** a UV filter.

## Checklist:
### General testing
- [x] Test without javascript
- [x] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/models/nano_element.rb
```diff
@@ -1,8 +1,19 @@
 class NanoElement < ApplicationRecord
   belongs_to :nano_material
 
+  def self.purposes
+    %w(colorant preservative uv_filter other).freeze
```
aidayounesi: For most of the notification properties (e.g. categories, physical form, etc.) we've used having a concern including an enum of values. Do you think we should use the same approach here and extract these bits from the model to a concern?

cosmetics-web/app/controllers/nanomaterial_build_controller.rb
```diff
@@ -5,10 +5,9 @@ class NanomaterialBuildController < ApplicationController
```
aidayounesi: Should we rename the first step to be plural, ``select_purposes``

cosmetics-web/app/models/nano_element.rb
tom-bell-softwire: I think we should generally declare enums and property value lists directly on the model itself. The only exceptions are where an enum or value list are used by multiple models, or where the lists are so long that they would dominate the length of the model class, in which case moving them to a concern helps to keep the business logic in the model clear.

In this case, we can't declare the `purposes` attribute as an [enum](https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html), since we allow multiple values (enums define a model attribute with a single value).

cosmetics-web/app/models/nano_element.rb
```diff
@@ -1,8 +1,19 @@
...
+  end
+
+  validates :purposes, presence: true, on: :select_purpose
+  validates :purposes, array: { presence: true, inclusion: { in: NanoElement.purposes } }
+
```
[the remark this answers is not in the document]
tom-bell-softwire: Good call. I'll combine them into one.

cosmetics-web/app/controllers/nanomaterial_build_controller.rb
tom-bell-softwire: Yes, good catch!

cosmetics-web/app/models/nano_element.rb
aidayounesi: Great, thanks for describing.

cosmetics-web/app/models/nano_element.rb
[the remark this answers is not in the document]
tom-bell-softwire: Actually, it turns out that these need to be treated as separate validations, due to how the `array` validator works. :(

PR #1193, opened 2019-07-03
PSD-1181: Disabled fuzzy transpositions for exact fields and changed …

samlionheart:
…'Case id' to 'Case ID'

## Description
Turned out it was a simple change that was hard to find. For some reason by default the term query has something called fuzzy transpositions (ab → ba) enabled by default. Disabling this for exact fields (which only contains case id) solves the problem, and now only brings exact matches to case ID's. For example searching:
1907-0009
brings only 1 case with that ID and no longer:
1907-0090
1907-000
1970-0009

Also changed id to be capitalised in the highlighted field.
## Checklist:

### DIFF
--- a/psd-web/app/views/investigations/_highlight_card.html.slim
+++ b/psd-web/app/views/investigations/_highlight_card.html.slim
@@ -7,7 +7,5 @@
     span = link_to investigation.title, investigation, class: "govuk-!-font-weight-bold"
   .govuk-grid-column-one-half
     - displayable_highlights.each do |highlight|
-      - label = capture do
-        = highlight[:label] == "Case id" ? "Case ID" : highlight[:label]
-      span.govuk-caption-m[class="govuk-!-font-size-16"] = label
+      span.govuk-caption-m[class="govuk-!-font-size-16"] = highlight[:label]
       span = highlight[:content]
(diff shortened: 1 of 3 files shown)


psd-web/app/views/investigations/_highlight_card.html.slim
```diff
@@ -7,5 +7,7 @@
     span = link_to investigation.title, investigation, class: "govuk-!-font-weight-bold"
   .govuk-grid-column-one-half
     - displayable_highlights.each do |highlight|
-      span.govuk-caption-m[class="govuk-!-font-size-16"] = highlight[:label]
+      - label = capture do
+        = highlight[:label] == "Case id" ? "Case ID" : highlight[:label]
```
matthewford: Feels like this isn't the right place for this, ideally, this would be handled in a translation file but I see there is already this code which looks like it's doing something similar. https://github.com/UKGovernmentBEIS/beis-opss/blob/59903042ded680cbaf955479080645fd594289b6/psd-web/app/helpers/investigations/display_text_helper.rb#L33-L39

samlionheart: Yep, good point. Simpler as well. 👍

matthewford:
Looks like we have a spec failure, not sure if that test should hit ES, if so need to setup the index, if not we could mock it.

PR #1261, opened 2019-08-05
Merge master with staging.

baphled:
## Description
Merge:
* COSBETA-534
* COSBETA-535

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/views/nanomaterial_build/non_standard_nanomaterial_notified.html.slim
+++ b/cosmetics-web/app/views/nanomaterial_build/non_standard_nanomaterial_notified.html.slim
@@ -1,0 +1,9 @@
+- title = "Select nanomaterial purpose"
+
+- content_for :page_title, title
+- content_for :after_header do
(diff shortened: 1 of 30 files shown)


cosmetics-web/app/views/nanomaterial_build/non_standard_nanomaterial_notified.html.slim
```diff
@@ -0,0 +1,9 @@
+- title = "Select nanomaterial purpose"
```
doctorpod: Comprehensive I18n usage would be a good goal - something for the future ?

baphled: I'd prefer it, though starting it now would be sketchy.

Once we're a bit more settled, I'm more than happy to soak it up as TD or enhancements :)

PR #1282, opened 2019-08-14
COSBETA-538 - Check if RP has notified non-standard nanomaterial

baphled:
## Description
Add content for non_standard_nanomaterial_notified and create dummy pages for the next steps in the journey

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


baphled:
Awesome, yeah, the wizard isn't too bad. It's better than using a state machine 😆

PR #1288, opened 2019-08-16
PSD: New enquiry page

samlionheart:
New attributes in database: Enquiry date_received and received_type

Two fields, date input using the date component, with javascript providing a button to set to today or yesterday.
Received type has Email, phone and 'other', if other is selected and left blank then an error is added.
Also chnaged title sizing on the enquiry flow to match.
## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


psd-web/app/controllers/investigations/creation_flow_controller.rb
```diff
@@ -92,9 +93,16 @@ def store_investigation
   end
 
   def investigation_valid?
-    @complainant.validate(step)
-    @investigation.validate(step)
-    validate_blob_size(@file_blob, @investigation.errors, "File")
+    if step == :about_enquiry
+      if params[:enquiry][:received_type].nil?
```
baphled: This feels like it should live in a AR model 🤔 

I don't know the difference between an `enquiry` and an `investigation`, in this context, though.

PR #1313, opened 2019-09-04
Re-try bulk assignment of self.data on failure

doctorpod:
## Description
Ticket: https://trello.com/c/tzbTEwby

In various models based on ActiveHash, data from Keycloak was assigned in bulk by means of `self.data =` - under heavy loads this has randomly lead to `Duplicate key` errors from ActiveHash.

Here we add a re-try mechanism if this occurs.

In local stress tests using JMeter, this approach prevented all status 500 errors that were happening before the changes. The logs revealed that only one retry was every necessary to successfully assign the Keycloak data.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/models/team.rb
+++ b/psd-web/app/models/team.rb
@@ -1,5 +1,6 @@
 class Team < ActiveHash::Base
   include ActiveHash::Associations
+  include ActiveHashSafeLoadable
 
   field :id
   field :name
@@ -30,7 +31,8 @@ class Team < ActiveHash::Base
   def self.load(force: false)
     Organisation.load(force: force)
     begin
-      self.data = Shared::Web::KeycloakClient.instance.all_teams(Organisation.all.map(&:id), force: force)
+      teams = Shared::Web::KeycloakClient.instance.all_teams(Organisation.all.map(&:id), force: force)
+      self.safe_load(teams, data_name: 'teams')
       self.ensure_names_up_to_date
     rescue StandardError => e
       Rails.logger.error "Failed to fetch teams from Keycloak: #{e.message}"
--- a/psd-web/app/models/team_user.rb
+++ b/psd-web/app/models/team_user.rb
@@ -1,14 +1,17 @@
 class TeamUser < ActiveHash::Base
   include ActiveHash::Associations
+  include ActiveHashSafeLoadable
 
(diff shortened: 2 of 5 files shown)


psd-web/app/models/team.rb
```diff
@@ -30,8 +30,9 @@ def add_user(user_id)
   def self.load(force: false)
     Organisation.load(force: force)
     begin
-      self.data = Shared::Web::KeycloakClient.instance.all_teams(Organisation.all.map(&:id), force: force)
-      self.ensure_names_up_to_date
+      teams = Shared::Web::KeycloakClient.instance.all_teams(Organisation.all.map(&:id), force: force)
+      teams.each { |team| self.add(team) } # self.data = teams sometimes causes duplicate key error in ActiveHash
```
baphled: Could we extract this out into a method that is accessed here and `psd-web/app/models/team_user.rb` 🤔

psd-web/app/models/team.rb
```diff
@@ -62,13 +63,13 @@ def assignee_short_name
...
 
-  def self.ensure_names_up_to_date
+  def self.ensure_names_up_to_date(teams)
     return if Rails.env.test?
 
     Rails.cache.fetch(:up_to_date, expires_in: 30.minutes) do
       Rails.application.config.team_names["organisations"]["opss"].each do |name|
         found = false
-        self.data.each { |team_data| found = found || team_data[:name] == name }
+        teams.each { |team_data| found = found || team_data[:name] == name }
```
baphled: I'm not sure why we're assigning `found` to itself 🤔

psd-web/app/models/team_user.rb
```diff
@@ -1,14 +1,17 @@
...
   include ActiveHash::Associations
+  include ActiveHashSafeLoadable
 
   belongs_to :team
   belongs_to :user
 
   def self.load(force: false)
     begin
-      self.data = Shared::Web::KeycloakClient.instance.all_team_users(
+      team_users = Shared::Web::KeycloakClient.instance.all_team_users(
```
baphled: Maybe. Change team_users to team_members 🤔

PR #1319, opened 2019-09-12
Better logging around creating correspondence

doctorpod:
Ticket: https://trello.com/c/XQGnlDyl

## Description
This problem was caused by the `date_received` being blank in old Enquiries (it's now mandatory). But it was difficult to diagnose. Better log messages have been put in so that this kind of problem will be easier to diagnose in the future.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
Closing as moved to https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/14.

PR #1320, opened 2019-09-16
Define an incomplete nanomaterial

baphled:
## Description

We need to allow a user to be able to update missing information for standard and non-standard nano-material.

This also includes saving the users standard and non-standard nano-materials data.

To do so we need to extend the functionality introduced by https://regulatorydelivery.atlassian.net/browse/COSBETA-560.

The service must check that a standard nanomaterials should be declared as `incomplete` if neither the nanomaterial has a purpose defined, it is listed on Annex 4, 5 or 6 and is used in accordance with 4, 5, or 6.

https://regulatorydelivery.atlassian.net/browse/COSBETA-612

Similarly the service needs to check that non-standard nanomaterial should be declared as `incomplete` if the user hasn't submitted a notification before brexit.

https://regulatorydelivery.atlassian.net/browse/COSBETA-616

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/views/nanomaterial_build/non_standard_nanomaterial_notified.html.slim
```diff
@@ -1,6 +1,6 @@
 - title = "Have you submitted a notification about #{@nano_element.inci_name} in the UK since Brexit?"
 - question = title
-- items = [{ text: "Yes", value: "yes" }, { text: "No", value: "no" }, { text: "Not sure", value: "not_sure" }]
+- items = [{ text: "Yes", value: "yes" }, { text: "No", value: "no" }, { text: "Not sure", value: "not sure" }]
```
frankieroberto: Did you mean to revert 827886dbd395b9bc718a0bec54af111842049a0c and remove the underscore?

cosmetics-web/app/models/nano_element.rb
```diff
@@ -1,8 +1,6 @@
 class NanoElement < ApplicationRecord
   belongs_to :nano_material
 
-  enum confirm_toxicology_notified: { yes: 'yes', no: 'no', not_sure: 'not sure' }
```
frankieroberto: This reverts 22785051cf612328741d5fd583663f40d727ff95?

cosmetics-web/db/schema.rb
```diff
@@ -115,6 +115,8 @@
     t.bigint "nano_material_id"
     t.string "purposes", array: true
     t.string "confirm_toxicology_notified"
+    t.string "confirm_usage"
+    t.string "confirm_restrictions"
```
frankieroberto: I find it a bit weird that these are strings, rather than booleans. Is that because there's a third state, as well as `yes` and `no` (`not_sure`)?

cosmetics-web/app/models/nano_element.rb
```diff
@@ -16,7 +14,19 @@ def display_name
...
-    purposes.nil?
+    purposes.blank? ||
+      (non_standard? && confirm_toxicology_notified.nil?) ||
+      (
+        standard? && (
+          confirm_restrictions.nil? ||
+          (confirm_restrictions == 'no' && confirm_toxicology_notified.nil?) ||
+          (confirm_restrictions == 'yes' && confirm_usage.nil?)
+        )
+      )
```
frankieroberto: Four levels of nested ORs and ANDs makes this quite hard to understand. Could it be refactored with named private methods?

cosmetics-web/app/models/nano_element.rb
baphled: Nah, just removed the need for the enum. There wasn't really a need for it.

cosmetics-web/app/views/nanomaterial_build/non_standard_nanomaterial_notified.html.slim
baphled: It's doesn't have to be `not_sure` if we get rid of the enum.

cosmetics-web/db/schema.rb
baphled: I agree, but in keeping with how things were done previously, I decided to use string here also. We should try to keep the code consistent until a better approach or understanding of the consequences are known.

cosmetics-web/app/models/nano_element.rb
baphled: IMHO ... This stuff shouldn't be in this model but I'm not sure what the alternative is yet.

I did play with extracting each check out but it feels less expressive and we end up testing indirectly.

Which tells me there's an object or two we're missing but I don't know what those look like yet.

Something like `NonStandardNanoMaterial#incomplete?` and `StandardNanoMaterial#incomplete?` could work but it feels over engineered atm.

I think things like this will become clearer once we review the architecture the service.

frankieroberto: Something like this might be a bit clearer (and shouldn’t affect your tests):

```ruby
def incomplete?
  purposes.blank? ||
  (non_standard? && non_standard_purpose_questions_incomplete?) ||
  (standard? && standard_purpose_questions_incomplete? )
end

private

def non_standard_purpose_questions_incomplete?
   confirm_toxicology_notified.nil?
end

def standard_purpose_questions_incomplete?
  confirm_restrictions.nil? ||
  (confirm_restrictions == 'no' && confirm_toxicology_notified.nil?) ||
  (confirm_restrictions == 'yes' && confirm_usage.nil?)
end
```

Not a show-stopper though.

cosmetics-web/db/schema.rb
frankieroberto: I wonder if `confirm_toxicology_notified` was a string because a `"not sure"` option was anticipated? Personally I’d rather incrementally improve the db structure rather than be consistent with previous decisions. But I can see the argument both ways.

cosmetics-web/app/models/nano_element.rb
frankieroberto: ok cool.

cosmetics-web/db/schema.rb
baphled: I'm more than happy to do so incrementally but we should agree on an approach as a team :)

cosmetics-web/app/models/nano_element.rb
baphled: Yeah, we had that before but it's not particularly helpful IMHO.

It feels to me that we may be missing one, or two objects. Possibly one of these objects would be responsible for determining whether a nanomaterial is complete or not.

That would be a good home for this logic, but it feels premature right now.

Sometimes it's okay to have a bit of verbose code, it can inform our design.

cosmetics-web/db/schema.rb
frankieroberto: Agreed, let's have that conversation. This can remain as a string for the time being.

PR #1322, opened 2019-09-16
Use webpacker-compatible paths for assets

slorek:
We aren't using the sprockets asset pipeline, so the way in which assets were loaded wasn't ideal, as it created a file system location dependency (i.e. they have to exist in `vendor`, which they don't by default unless built via Docker which is very slow).

Also, despite the assets in `shared-web` being specified (correctly) as a dependency in `package.json` the applications were loading them from `vendor`, which is 'managed' completely independently (in the Docker build scripts).

We want to split `shared-web` out into a separate repository, so this change will enable it to work regardless of how the assets are loaded via `yarn`. It also means the assets can be loaded correctly when the applications are running directly instead of via Docker.

I think I want to instead specify a Git dependency in `package.json` once `shared-web` is split out. This will put the assets in `node_modules` instead.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/psd-web/app/views/homepage/non_opss.html.slim
+++ b/psd-web/app/views/homepage/non_opss.html.slim
@@ -1,5 +1,10 @@
 - content_for :page_title, "Home Page"
-
+- banner_content = capture do
+  p
+    strong = "We’re aware of intermittent problems with the service."
+  p = "Thank you for your patience while we investigate."
+  p = "(Last updated 24 September 2019)"
+= render "components/hmcts_banner", html: banner_content
 .govuk-grid-row
   .govuk-grid-column-two-thirds-from-desktop
     h1.govuk-heading-l


frankieroberto:
This looks ok to me as a short-term fix, but I wonder whether we should fairly quickly move all the CSS and javascript from `shared-web` to the 2 individual apps? This would mean duplicating it initially – but then the two apps could remove any unneeded assets and maintain their own forks?

slorek:
This is a good point, but there is an ambition to create re-usable Ruby ports of the Design System components. Theoretically, the components should be re-usable and standardised. I know this would be valuable across government, not in its current form perhaps but the way I'd like to explore going forward is to strip everything else out of `shared-web` and make a Ruby port of the GOV.UK Design System.

frankieroberto:
@slorek I'm up for making a Ruby port of the Design System macros, but it looks to me like most of the shared CSS and JS assets within `shared-web` aren't from Design System (as those are installed via npm), but are a bunch of other random bits of code (eg the `cookie_banner.js`, `search-box.scss` and so on)?

frankieroberto:
@slorek I can't get this to work locally. Am getting the following error:

```
ERROR in ./app/assets/application/stylesheets/main.scss (./node_modules/css-loader/dist/cjs.js??ref--7-1!./node_modules/postcss-loader/src??ref--7-2!./node_modules/resolve-url-loader!./node_modules/sass-loader/dist/cjs.js??ref--7-4!./app/assets/application/stylesheets/main.scss)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):

@import "shared-web/assets/stylesheets/main";
^
      File to import not found or unreadable: shared-web/assets/stylesheets/main.
```

Any ideas? This is when running `./bin/webpack`

slorek:
@frankieroberto haven’t sent that myself, did you run `yarn install` first?

frankieroberto:
@slorek yep. I think it's because Sass is resolving the `shared-web/assets/stylesheets/main` path relative to the `cosmetics-web` folder rather than its parent. Where are you running `./bin/webpack` from?

slorek:
> @slorek yep. I think it's because Sass is resolving the `shared-web/assets/stylesheets/main` path relative to the `cosmetics-web` folder rather than its parent. Where are you running `./bin/webpack` from?

I don't think that's the problem. `yarn install` should copy `shared-web` into `node_modules` and that's where it is looking. Do you see the `shared-web` folder in there? Webpack should be using the node load path, so if it's there it should be working. If it's not there, there's an issue with the node module not being installed.

Not sure I tested it yet with `/bin/webpack`, only '/bin/webpack-dev-server`. Does that work for you?

frankieroberto:
Ok, I've got this working now by changing `package.json` from

```
"shared-web": "./vendor/shared-web"
```

to

```
"shared-web": "../shared-web"
```

PR #1324, opened 2019-09-18
COSBETA-615 - Storing nano material data

baphled:
## Description

Resolves the issue where a user can select "not sure" but the value is not actually saved.

https://regulatorydelivery.atlassian.net/browse/COSBETA-615

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


baphled:
We'll need to wait for the changes about to go into production, before we can merge this.

PR #1329, opened 2019-09-18
Remove simplecov

frankieroberto:
This stops `simplecov`, a code coverage reporter, from running as part of the `rspec` tests for the Cosmetics application.

I'm not sure this provides much value, and it makes it harder to quickly spot the results of the tests.


frankieroberto:
We could remove this from the `shared-web` gemspec too, if it’s also not useful for the PSD team?

PR #1330, opened 2019-09-19
COSBETA 614: Create guidance for creating a list of ingredients

baphled:
## Description
Add guidance for creating a list of ingredients

WIP: Waiting on content

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


frankieroberto:
@baphled @ralph-hawkins is it intended for these guidance pages to be public, or to require an account?

baphled:
> Would be good to get content designer approval on this PR as well.

Agreed, think Steve isn't on here, so I'll add to Jira, for now.

ralph-hawkins:
> @baphled @ralph-hawkins is it intended for these guidance pages to be public, or to require an account?

Public

ralph-hawkins:
This content is still in progress – can we hold work until it's finished?

Should be ready on Monday 23

cosmetics-web/app/views/guidance/how_to_create_your_list_of_ingredients.html.slim
```diff
@@ -0,0 +1,186 @@
+- title = "How to create your list of ingredients"
+
+- content_for :page_title, title
+- hint = capture do
```
frankieroberto: Maybe rename this variable to `ingredients_text` as it's not really a hint?

cosmetics-web/app/views/guidance/how_to_create_your_list_of_ingredients.html.slim
```diff
@@ -0,0 +1,186 @@
...
+    li Xanthine derivatives
+
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    h1.govuk-heading-xl
+      = title
+
+    p class="govuk-!-margin-bottom-1"
+      | If you notify using exact concentrations or concentration ranges,
+      ' you need to create a document which lists the cosmetic product’s ingredients:
```
frankieroberto: This line seems to render immediately after the previous one, with no whitespace, so that it reads `ranges,you`

cosmetics-web/app/views/layouts/_footer.html.slim
```diff
@@ -1,5 +1,22 @@
 footer.govuk-footer[role="contentinfo"]
   .govuk-width-container
+    .govuk-footer__navigation
+      .govuk-footer__section
+        h2.govuk-footer__heading.govuk-heading-m
+          | Guidance
+        ul.govuk-footer__list.govuk-footer__list
+          li.govuk-footer__list-item
+            = link_to "How to notify nanomaterials", how_to_notify_nanomaterials_path,
```
frankieroberto: These first two are unrelated to this pull request?

cosmetics-web/app/views/guidance/how_to_create_your_list_of_ingredients.html.slim
baphled: It's a common pattern within the codebase :)

More than happy to come up with ways of standardising going forwards.

cosmetics-web/app/views/guidance/how_to_create_your_list_of_ingredients.html.slim
baphled: Good catch

cosmetics-web/app/views/layouts/_footer.html.slim
baphled: It's because it's linked to a previous ticket, once we merge that. These will go away.

cosmetics-web/app/views/guidance/how_to_create_your_list_of_ingredients.html.slim
frankieroberto: I guess the Details component is mostly used for hint text elsewhere? But here it’s guidance, so calling it `hint` is a bit confusing (confused me at least).

baphled: I agree it could be called something else, not sure `ingredients_text` is that helpful though, and `text` is even less so 🤔 🤷‍♂

baphled: As the copy needs sign off, we can mull over it, for now. Hopefully we'll come up with a better name then.

baphled: Writing a test, to check that this is accessible to the public, without auth, seems like overkill imho.

This is the equivalent of a static page, we should only check auth if it's required. Otherwise we'll end up with boiler plate tests that have little value.

frankieroberto: @baphled well, it would catch the current issue which is that the page _does_ require logging in. :D

I’ve also found that having basic tests to check that every view at least renders is useful for catching inadvertent syntax errors in the view templates. Especially for pages which you might not otherwise look at that often (like privacy policy pages and so on).

baphled: If these pages are public, there’s no need to test auth :)

As for the last point we’ve got linters and we review copy. A test to render a plain view seems overkill imho. I’d definitely consider implementing one or two if there was actual logic in the controller. That’s starting to test Rails, which holds little value.

frankieroberto: @baphled the test would be to make sure the page renders without auth (which it currently doesn’t). The main value I’ve found in a basic "does it render" test is that it can catch any errors in your view template.

baphled: I’m not sure I follow, the page relies on application controller. If the page is behind with, which it shouldn’t, that needs resolving.

A render test isn’t the way to go. If the above is true, we’re likely missing a check that needs to be introduced at the application controller level. A spec or two should accompany that.

frankieroberto: I think you’ll need to add

```ruby
  skip_before_action :authenticate_user!
  skip_before_action :authorize_user!
```

to the controller to turn off the authentication and authorisation. See [example from the HelpController](https://github.com/UKGovernmentBEIS/beis-opss/blob/master/cosmetics-web/app/controllers/help_controller.rb#L2)

baphled: We'd just need to turn off authentication but we'll need to disable a couple more things. Just added a parent controller (`PubliclyAccessibleController`) and it's accompanying specs to handle all this.

PR #1331, opened 2019-09-19
Add poison information question page

frankieroberto:
This adds a new question page for users adding products with a frame formulation, which lists poisonous ingredients and asks them with their product contains any of these:

does-your-product-contain-poisonous-ingredients

After answering this, they are taken to the existing PH level question, skipping the "Does the cosmetic product contain anti-dandruff agents?" question.

An error is displayed if the user doesn't select either Yes or No.

The Yes or No answer is included within the Check your answers summary page at the end.

See https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-632


cosmetics-web/app/views/component_build/contains_poisonous_ingredients.html.slim
```diff
@@ -0,0 +1,146 @@
...
+
+    = form_with model: @component, url: wizard_path, method: :put do |form|
+
+      = render "form_components/govuk_error_summary", form: form
+
+      h1.govuk-heading-xl Poisonous ingredients
+
+      p.govuk-body
+        | To treat anyone who may be poisoned by the product,
+          National&nbsp;Poison&nbsp;Centres need the exact quantities (% w/w)
```
ralph-hawkins: National&nbsp;Poison&nbsp;Information&nbsp;Services

cosmetics-web/app/views/component_build/contains_poisonous_ingredients.html.slim
```diff
@@ -0,0 +1,146 @@
...
+
+      p class="govuk-body govuk-body-s govuk-body-s govuk-!-margin-bottom-1" More than 1% w/w
+
+      ul.govuk-list.govuk-list--bullet.govuk-body-s
+        li Benzyl Benzoate
+        li Butyl Acetate
+        li Ethyl Acetate
+        li Hydroxylamine HCL, Hydroxylamine Sulfate
+
+      h2.govuk-heading-s Any ingredients classified as:
```
ralph-hawkins: Remove ":" for consistency

cosmetics-web/db/migrate/20190919114626_add_contains_poisonous_ingredients_to_components.rb
```diff
@@ -0,0 +1,5 @@
+class AddContainsPoisonousIngredientsToComponents < ActiveRecord::Migration[5.2]
+  def change
+    add_column :components, :contains_poisonous_ingredients, :boolean
```
baphled: Could we add a model spec to describe this attribute 🤔

Also, it’s worth checking what we’ve done with yes and no type questions in the past.

I prefer Boolean but I have a feeling we’ve made these enums or strings in the past.

If so we should follow the convention or come up with a more standardised way of doing this.

cosmetics-web/app/controllers/trigger_questions_controller.rb
```diff
@@ -94,6 +94,9 @@ def finish_wizard_path
   def previous_wizard_path
     if step == steps.first
       return responsible_person_notification_component_build_path(@component.notification.responsible_person, @component.notification, @component, :select_frame_formulation)
+
+    elsif step == :select_ph_range && @component.predefined?
```
baphled: Would be useful to capture this change with a spec or two :)

frankieroberto: I just looked into this, but it’s kinda tricky as `previous_wizard_path` isn't assigned as an instance variable, but instead seems to be a public method on the controller which is somehow exposed as a helper to the view.

Possibly the best way to test this would be as an integration test, by actually rendering the view and then asserting that the view contains a back link to the right page?

cosmetics-web/db/migrate/20190919114626_add_contains_poisonous_ingredients_to_components.rb
frankieroberto: I think we should definitely move towards using boolean columns for yes and no questions.

We don't seem to have any model specs for plain Rails attributes so far, and I'm not sure how useful that would be?

slorek: I've seen a couple of instances where yes/no/not_sure were valid values so this may have stemmed from that.

I don't think we need to test ActiveRecord behaviour, unless there's something special about this attribute.

cosmetics-web/app/controllers/trigger_questions_controller.rb
baphled: We should be able to check it renders or redirects to where we exit it to go 🤔

cosmetics-web/db/migrate/20190919114626_add_contains_poisonous_ingredients_to_components.rb
baphled: We should test behaviour but we should at least describe attributes 🤷‍♂️

cosmetics-web/app/controllers/trigger_questions_controller.rb
frankieroberto: This isn't used for redirecting though - just for setting the `href` attribute of the back link?

slorek: there is some code smell here. I want to avoid adding more integration tests as it's all very slow already. Probably `step` (and others) should be injected as an argument to `previous_wizard_path` so we can unit test it easily.

cosmetics-web/db/migrate/20190919114626_add_contains_poisonous_ingredients_to_components.rb
baphled: @frankieroberto agreed, though it’d be nice to agree on that and tackle it as TD, I’m assuming we’d need a presentation layer introduced. Especially for “check your details”

cosmetics-web/app/controllers/trigger_questions_controller.rb
frankieroberto: It’s definitely all a bit weird and hard to test. I haven't yet gotten my head around how WickedWizard actually works.

Could we flag this a task for a future refactoring? This logic (and the whole controller) will soon change when some future cards are implemented.

cosmetics-web/db/migrate/20190919114626_add_contains_poisonous_ingredients_to_components.rb
frankieroberto: I've added a test for the attribute in 09715fdf095225a9f039b86b9ad91cc3201e29de but personally I'm still not convinced this is useful.

cosmetics-web/app/controllers/trigger_questions_controller.rb
baphled: I agree @slorek, although this is a new addition to an already unnecessarily complex flow. I'm not fussed how it's tested but we should at least make sure we're not unexpectedly breaking anything here.

We could easily use a PORO that takes the steps and returns x. Depending on that return value, we send the user to the appropriate step.

frankieroberto: Agreed with @baphled to catch this as part of our QA / acceptance tests.

cosmetics-web/spec/models/component_spec.rb
```diff
@@ -6,8 +6,11 @@
   let(:exact_component) { create(:exact_component) }
   let(:text_file) { fixture_file_upload('/testText.txt', 'application/text') }
 
-  it "has a contains_poisonous_ingredients attribute" do
-    expect(described_class.new).to have_attributes(contains_poisonous_ingredients: nil)
+  describe "attributes" do
+    subject(:component) { described_class.new }
+    it "has a contains_poisonous_ingredients attribute" do
```
baphled: I'd remove "attribute" from the end of the it block, but we're good.

cosmetics-web/spec/models/component_spec.rb
```diff
@@ -6,8 +6,11 @@
...
+      is_expected.to have_attributes(contains_poisonous_ingredients: nil)
```
baphled: We can use `expect(component)to have_attributes(contains_poisonous_ingredients: nil) if we're using named subjects

frankieroberto: I originally had that, but then switched to `is_expected`. I think the named subject reads better though (and [Rubocop agrees](https://github.com/rubocop-hq/rspec-style-guide#use-subject)) so I'll switch back.

PR #1332, opened 2019-09-19
Allow 'lograge' to be disabled from an env variable

frankieroberto:
This alters the 'lograge' configuration to enable it to be turned off by setting a `LOGRAGE_ENABLED` environment variable to `false`. It remains enabled by default if no environment variable is present.

[Lograge](https://github.com/roidrage/lograge) is a Ruby gem which changes the default Rails logging format to make it less verbose, instead having a single line of information. Whilst this may be useful for production environments, I've found it to be frustrating in development, as the default Rails logging contains useful information such as the parameters, and each view and partial being rendered.

With lograge:

Screenshot 2019-09-19 at 17 07 16

Without lograge (Rails default):

Screenshot 2019-09-19 at 17 06 44

### DIFF
--- a/shared-web/config/initializers/lograge.rb
+++ b/shared-web/config/initializers/lograge.rb
@@ -1,7 +1,7 @@
 # Be sure to restart your server when you modify this file.
 
 Rails.application.configure do
-  config.lograge.enabled = true
+  config.lograge.enabled = ENV.fetch('LOGRAGE_ENABLED', 'true') == 'true'
 
   config.lograge.custom_payload do |_controller|
     extra_payload = {}


frankieroberto:
@slorek I was guessing that the logs were being parsed in this format by some service or another.

I was tempted to default Lograge to be disabled in development for everyone, but didn't want to make this decision without consulting all the other developers first. Probably makes sense though - I can't see any reason why anyone would prefer the lograge format in development vs the more verbose Rails default format.

slorek:
👌 As we split `shared-web` apart I want to copy development dependencies into each app so we can avoid this cross-dependency.

PR #1333, opened 2019-09-20
Allow Redis URL to be configured via environment variable

frankieroberto:
The existing configuration assumes that Redis is running on a host called `redis` in development and test environments.  This change allows this default to be overriden by setting a  `REDIS_URL` environment variable.

This is useful, I think, as running Redis locally on the `redis` host means altering your `/etc/hosts` file, which is a minor inconvenience and not necessarily obvious.


frankieroberto:
@baphled yep, but I'll do those as separate pull requests. 😄

PR #1337, opened 2019-09-23
Change ph question to a range

frankieroberto:
This changes the "What is the PH of the product?" from asking for a single PH number to asking for a range – ie the maximum and minimum PH.

https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-600

**Before**

what-is-the-ph

**After**

Screenshot 2019-10-02 at 15 02 35

As is currently the case, this question is only asked if the user previously indicates that the PH is outside of the range of between 3 and 10.

**Check your answers page when a range is entered**

Screenshot 2019-10-02 at 15 02 18

**Check your answers page when minimum and maximum pH are the same**

Screenshot 2019-09-25 at 16 07 13

The way this data is saved to the database also changes, from being stored against a `TriggerQuestionElement` as a string to being stored as two decimals (`minimum_ph` and `maximum_ph` on the `Component` itself. This is to simplify the data model, make it easier to query, and in anticipation of dropping the `TriggerQuestion` and `TriggerQuestionElement` models soon.

**TODO**
- [x] Data migration for existing values
- [x] Update the Check your answers screen
- [x] Any additional validations
- [x] Check impact on bulk upload flow
- [x] Consider whether text input boxes should appear side-by-side


ralph-hawkins:
Looks good, can you change: 
- the input labels from "PH" to "pH"
- the "-" in your check your answers page between the pH range values to "to"

frankieroberto:
> Looks good, can you change:
> 
> * the input labels from "PH" to "pH"
> * the "-" in your check your answers page between the pH range values to "to"

Thanks - this is done.

ralph-hawkins:
LGTM 👍 

We can fix the alignment of the fields once we've got a few more important things sorted.

cosmetics-web/spec/db/migrate/20190923133032_add_minimum_and_maximum_ph_to_components_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        expect(trigger_questions.first.applicable).to be true
+      end
+
+      it "creates a trigger question element for the component" do
+        ActiveRecord::Migrator.new(:down, migrations, previous_version).migrate
+
+        trigger_question = TriggerQuestion.find_by(component_id: component.id)
+        trigger_question_element = TriggerQuestionElement.find_by(trigger_question_id: trigger_question.id)
+
+        expect(trigger_question_element.answer).to eq '3.2'
```
baphled: A bit pedantic but could we keep expectations down to 2 per it block.

The issue with having multiple expectations in 1 block is that they can hide broken tests, which makes it harder to get to the bottom of failing tests.

Ideally 1 would be sweet, but I'm aware we have a few specs that use 2 already.

frankieroberto: Good point! I've refactored the migration spec to have one expectation per test. Let me know what you think.

cosmetics-web/app/models/cpnp_export.rb
```diff
@@ -63,7 +63,7 @@ def shades
 
   def components
     current_version_component_lists_node.xpath(".//component").collect do |component_node|
-      Component.create(name: component_name(component_node),
+      Component.create!(name: component_name(component_node),
```
baphled: Is there a reason why we’d prefer to throw an exception if the object can’t be saved?

cosmetics-web/spec/analyzers/read_data_analyzer_spec.rb
```diff
@@ -212,5 +212,29 @@
...
+
+    it "imports the PH range of a component" do
+      notification = create(:notification)
+      notification_file = create(:notification_file, uploaded_file: create_file_blob("testExportWithComponentWithPHRange.zip"))
+
+      analyzer_instance = ReadDataAnalyzer.new(notification_file.uploaded_file)
+      analyzer_instance.metadata
+      notification = Notification.order(created_at: :asc).last
+
+      expect(notification.components.first.minimum_ph).to eq(13.0)
```
baphled: An expectation per it block would be awesome :)

cosmetics-web/app/models/cpnp_export.rb
frankieroberto: Yeah, I think we’d want to see an exception if this fails to save, so that we can see where any errors are in Sentry? Better than failing silently I think...?

cosmetics-web/.rubocop.yml
```diff
@@ -4,7 +4,7 @@ inherit_gem:
   shared-web: .rubocop.yml
 
 RSpec/MultipleExpectations:
```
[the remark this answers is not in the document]
frankieroberto: 👍

frankieroberto:
Closing this in favour of https://github.com/UKGovernmentBEIS/beis-opss/pull/1372 which is against prod.

cosmetics-web/app/models/cpnp_export.rb
baphled: Yeah, though the exception being thrown won't be handled cleanly and we'll still have to dig into which one of these components failed, and why, as this is looping 🤔 

What kind of exception are we expecting?

Maybe it's okay to have return nil here or some kind of sentinel, so that we have a rough idea of what actually went wrong.

frankieroberto: Ok I've reverted for now, so it'll return a `nil`. This then causes the overall notification to fail to save when called from `ReadDataAnalyzer`, but it looks like there’s some existing error handling there.

The reason I added it was because we have an existing bug with parsing ZIP files containing components with PH ranges, but it was hard to debug because only a generic exception was being raised.

Let's revisit this at some point further down the line.

cosmetics-web/spec/analyzers/read_data_analyzer_spec.rb
```diff
@@ -212,5 +212,44 @@
       expect(notification.components.first.shades.first).to eq("blue, green")
       expect(notification.components.second.shades.first).to eq("pink, lazuli")
     end
+
+
+    context "when the file contains a component with a PH range" do
+      before do
```
baphled: Could use `let` when instantiating, so we're not setting these values again, on each tests.

PR #1339, opened 2019-09-24
PSD Date received bug fix

samlionheart:
…Added date_received blank validation to the enquiry model.

Trello: https://trello.com/c/10kGi6Ox/44-bug-date-received-now-mandatory-for-enquiry

It now validates if the date is blank in the model rather than the component, as the component forces validation of date_received every time it's updated, so setting the component to required: false, disables empty validation, which is now on the enquiry model.

## Checklist:


psd-web/app/models/investigation/enquiry.rb
```diff
@@ -20,9 +20,15 @@ def case_type
     "enquiry"
   end
 
+  def date_cannot_be_blank
+    if date_received.blank? && errors.messages.blank?
```
slorek: Consider using the built-in `presence` validation (see https://edgeguides.rubyonrails.org/active_record_validations.html#presence). Also I'm not sure why we would want this to only work when there are no other error messages?

samlionheart: Yeah this is true, either way I kinda have to work around the date component to make it work nicely. I've used `validates` for date_received so it can't be blank like you suggested, but have to override the validation if one of the three date input boxes is filled (as the date component adds an error message, so to avoid "Date can't be blank" and "Please enter a month" showing up on top of each other).

Hope that explanation makes sense.

psd-web/test/models/enquiry_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  teardown do
+    reset_keycloak_and_notify_mocks
+  end
+
+  test "enquiry with correct date" do
+    investigation = load_case(:enquiry)
+    assert_equal "2019-01-01", investigation.date_received.to_s
+  end
+
+  test "enquiry received date validated" do
```
slorek: As a reader I'm not clear on what the intent is with these tests, certainly outside of the context of this PR I'm not sure I would guess what is being tested here. I would expect to see something like:

```ruby
user = User.new(name: "Test", date_received: "2009-01-01")
assert(user.valid?)

user = User.new(name: "Test", date_received: "not_valid")
assert(user.invalid?)
```

Does this make sense? Happy to discuss.

PR #1345, opened 2019-09-26
Remove external regression tests as part of build/deploy pipeline.

slorek:
These tests are fundamentally brittle and are managed externally, so we should not have a dependency on them during our automated build/deploy pipeline. These tests may still have value if run nightly or manually.


slorek:
Tests observed passing. Rebasing before merge.

PR #1347, opened 2019-09-26
Allow us to configure production GOV.UK PaaS environment independent …

slorek:
…of other environments.

We should be running more than one instance in production. I haven't bumped up keycloak or the others as I'm not sure we know they can run this way reliably.


slorek:
> I wonder if it is possible to pass instances count to manifest as `ENV` variable, instead copying manifest file?

I think there's a few things we would want to configure independently; we're doing both memory and instance count here. We would want that configured in a convenient way. I don't like the duplication in the rest of the file, though.

PR #1348, opened 2019-09-26
Get scout key from VCAP_SERVICES

doctorpod:
## Description
Ticket: https://trello.com/c/CQ2wukzQ/

The `SCOUT_KEY` env var was not set up correctly on staging and production. They were set ad-hock as user defined but now are set within the `VCAP_SERVICES` env var via the cloud foundry user provided services mechanism.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


slorek:
This may not actually work. See https://github.com/UKGovernmentBEIS/beis-opss/blob/master/infrastructure/env/get-env-from-vcap.sh. You might need to use `ENV` and make sure the key name in the service is named correctly.

slorek:
Merged into [new repo](https://github.com/UKGovernmentBEIS/beis-opss-psd).

PR #1349, opened 2019-09-26
Switch to using port 3000 for the app in both development and CI

frankieroberto:
This updates the keycloak configuration and the `docker-compose` files to use port 3000 for the main application.

The main advantage of this is allowing developers to run the main application outside of Docker without having to remember to specify a different port from the Rails default one.


slorek:
@frankieroberto should be able to close this now as #1375 and #1382 address these issues.

frankieroberto:
@slorek yep, although I think we still need to figure out how to avoid needing the changes in `keycloak/configuration/initial-setup.json` locally?

slorek:
You can add an extra redirect URL to include port 3000, but the admin URL looks like it can only have one. I guess you could just make sure you run it on port 3002...

slorek:
@frankieroberto we can probably make use of your changes to `keycloak/configuration/initial-setup.json` since ports 3000/3001 were reserved for `psd-web` in the Docker context I believe. These ports should now be available for use as we have now removed PSD from this repo. The benefit being you won't need to specify port 3002 anymore. Do you want to rebase this branch so we can try it?

frankieroberto:
@slorek can you spot why CI might be failing the system tests? Must be connected with the port change somehow but I can't see it.

slorek:
@frankieroberto the system tests will spin up a new server listening on port 3001 (was 3003) so in the Keycloak config where you changed the whitelisted URLs from 3003 you should change them to 3001. Also check that 3003 isn’t configured/hard coded anywhere.

frankieroberto:
@slorek I’ve whitelisted 3001 in the keycloak config, but the system tests are still failing. I did think I’d updated the docker-compose files to run the app on port 3000 (line 5 of `cosmetics-web/spec/system_helper.rb`) but that doesn't seem to be helping either... 😞

cosmetics-web/spec/system_helper.rb
```diff
@@ -2,7 +2,7 @@
 require "selenium-webdriver"
 
 ENV["HTTP_HOST"] = "localhost"
-ENV["HTTP_PORT"] = "3003"
+ENV["HTTP_PORT"] = "3000"
```
slorek: I think this should be `3001`

frankieroberto:
@slorek this is still not working... 😞  Any ideas?  It'd be nice to get to a point where we can run `bin/rails` without having any uncommitted changes to the keycloak config.

frankieroberto:
Closing this for now, as we have a reasonable workaround.

PR #1350, opened 2019-09-27
Remove ActiveHashSafeLoadable#safe_load success logging

slorek:
This is producing a lot of noise in the log which isn't very useful.


slorek:
Merged in [new repo](https://github.com/UKGovernmentBEIS/beis-opss).

PR #1351, opened 2019-09-27
Skip poisonous ingredients question for pre-Brexit products

frankieroberto:
This updates the follow of the component questions so that _only_ post-Brexit products get asked whether the product contains harmful ingredients when using a frame formulation.  Pre-Brexit products skip straight to the PH question instead.

Fixes a bug introduced by https://github.com/UKGovernmentBEIS/beis-opss/pull/1331

### DIFF
--- a/cosmetics-web/app/controllers/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/component_build_controller.rb
@@ -275,7 +275,7 @@ private
   end
 
   def update_frame_formulation
-    if @component.update_with_context(component_params, :update_frame_formulation)
+    if @component.update_with_context(component_params, :select_frame_formulation)
 
       if @component.notification.notified_post_eu_exit?
         # Redirect to "Contains poisonous ingredients?" question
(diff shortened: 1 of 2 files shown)


cosmetics-web/app/controllers/component_build_controller.rb
```diff
@@ -272,6 +274,20 @@ def render_select_formulation_type
     render_wizard @component
   end
 
+  def update_frame_formulation
+    if @component.update_with_context(component_params, :select_frame_formulation)
+
+      if @component.notification.notified_post_eu_exit?
+        # Redirect to "Contains poisonous ingredients?" question
```
baphled: Is there a way we could make the redirect more explicit, so we could remove the comment? 🤔

frankieroberto: @baphled good idea, thanks. Done now.

baphled: :shipit:

PR #1371, opened 2019-10-07
Reference shared-web rather than copy it

frankieroberto:
Currently the Dockerfile copies `shared-web` into `vendor`, and both the `Gemfile` and `package.json` specifies it as a dependency within vendor.

This change removes the copy step, and just references the shared-web folder within the parent directory instead. This makes local development easier, as manual copying is no longer required to run the application without using Docker.


slorek:
I suspect this’ll result in shared-web not getting included in the deployment. I think in the first instance it should be separated into a new repo and referenced via Git. Then the things that shouldn’t be in there copied into the main apps. And the Design System components removed into a separate repo from that.

frankieroberto:
@slorek yeah you're right, this is going to break our current Docker-based build process. I'll abandon this for now, and we can re-review.

PR #1372, opened 2019-10-07
Change component ph to a range

frankieroberto:
This changes the "What is the PH of the product?" from asking for a single PH number to asking for a range – ie the maximum and minimum PH.

https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-600

**Before**

what-is-the-ph

**After**

Screenshot 2019-10-02 at 15 02 35

As is currently the case, this question is only asked if the user previously indicates that the PH is outside of the range of between 3 and 10.

**Check your answers page when a range is entered**

Screenshot 2019-10-02 at 15 02 18

**Check your answers page when minimum and maximum pH are the same**

Screenshot 2019-09-25 at 16 07 13

The way this data is saved to the database also changes, from being stored against a `TriggerQuestionElement` as a string to being stored as two decimals (`minimum_ph` and `maximum_ph` on the `Component` itself. This is to simplify the data model, make it easier to query, and in anticipation of dropping the `TriggerQuestion` and `TriggerQuestionElement` models soon.

**TODO**
- [x] Data migration for existing values
- [x] Update the Check your answers screen
- [x] Any additional validations
- [x] Check impact on bulk upload flow
- [x] Consider whether text input boxes should appear side-by-side


frankieroberto:
@slorek @baphled I’ve cherry-picked this from master and rebased it against prod. Please re-review!

This also still needs to be QAed, so would ideally get deployed to staging or int (before merging into prod).

PR #1373, opened 2019-10-07
Move development dependencies to main application's Gemfile

slorek:
Development dependencies should be managed independently by the main applications. This change ensures that, and removes the extra gem dependency.

The issue of the strict versioning of dependencies will be resolved separately.


frankieroberto:
> Gems should be sorted in an alphabetical order within their section of the Gemfile.
🙄

slorek:
> > Gems should be sorted in an alphabetical order within their section of the Gemfile.
> 
> 🙄

Oops, thought I had done but obviously not 😂

slorek:
> Looks good. Could we remove simplecov given that we agreed to stop this from running automatically in #1329?

SimpleCov is still required by `Shared::Web` so for now let's leave it and we can remove it if we want to once I've finished #1374

PR #1375, opened 2019-10-07
Provide meaningful error message when domains aren't set

slorek:
If you try to start the application without providing the `SUBMIT_HOST` and `SEARCH_HOST` env vars, the application exists with an unhelpful error message. This fixes that.

### DIFF
--- a/cosmetics-web/lib/constraints/domain_constraint.rb
+++ b/cosmetics-web/lib/constraints/domain_constraint.rb
@@ -1,5 +1,7 @@
 class DomainConstraint
   def initialize(domains)
+    raise "No domains specified" unless domains
+
     @domains = domains.split(',')
   end
 
--- a/cosmetics-web/config/routes.rb
+++ b/cosmetics-web/config/routes.rb
@@ -26,7 +26,7 @@ Rails.application.routes.draw do
     mount Sidekiq::Web => "/sidekiq"
   end
 
-  constraints DomainConstraint.new(ENV["SEARCH_HOST"] || ENV["COSMETICS_HOST"]) do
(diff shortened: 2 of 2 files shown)


cosmetics-web/lib/constraints/domain_constraint.rb
```diff
@@ -1,5 +1,6 @@
 class DomainConstraint
   def initialize(domains)
+    raise "No domains specified - are SUBMIT_HOST and COSMETICS_HOST environment variables set?" unless domains
```
frankieroberto: Feels like a slightly odd place to be referencing the named environment variables, as those aren't strictly required by this class. That said, raising an error if initialised with `nil` feels sensible.

I think I'd be tempted to change `ENV["SEARCH_HOST"]` to `ENV.fetch("SEARCH_HOST")` in the `routes.rb` as that will throw a fairly descriptive error straight away.

I'm not entirely sure what the intention behind the `COSMETICS_HOST` option is –  the readme and the code seems to suggest that we use `SEARCH_HOST` and `SUBMIT_HOST`?

baphled: This approach seems fine for now, it'll direct us to the problem, which should be a rarity, but also loud enough so we can address this issue of two services in one ASAP.

slorek: I don't really understand the `COSMETICS_HOST` either, except in the `README.md` it states:

>>> There's one final environment variable which is the URL for the website and is used for sending emails:

    cf set-env cosmetics-worker COSMETICS_HOST XXX

frankieroberto:
@slorek yep, although I think we still need to figure out how to avoid needing the changes in `keycloak/configuration/initial-setup.json` locally?

PR #1376, opened 2019-10-07
Add contains poisonous ingredients question

frankieroberto:
This adds a new question page for users adding products with a frame formulation, which lists poisonous ingredients and asks them with their product contains any of these:

does-your-product-contain-poisonous-ingredients

After answering this, they are taken to the existing PH level question, skipping the "Does the cosmetic product contain anti-dandruff agents?" question.

An error is displayed if the user doesn't select either Yes or No.

The Yes or No answer is included within the Check your answers summary page at the end.

See https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-632

**Re-opening #1331 and #1351, now rebased against `prod`.**


frankieroberto:
@slorek @baphled I’ve cherry-picked this from master and rebased it against prod. Please re-review.

PR #1377, opened 2019-10-08
Redirect the user to Additional Information when the user has uploaded a bulk file

baphled:
## Description

Making sure that when a user is completes the nanomaterial section of the bulk journey that they are sent to frame formulation next.

This is done by making sure the user has uploaded a bulk files and then redirecting the to additional information, similarly to how we've done with frame formulation and product image upload.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/spec/controllers/nanomaterial_build_controller_spec.rb
```diff
@@ -164,6 +164,26 @@
 
         expect(response).to redirect_to(responsible_person_notification_component_nanomaterial_build_path(responsible_person, notification, component, nano_element1, :notify_your_nanomaterial))
       end
+
+      context "when completed nanomaterial within bulk journey" do
+        before do
+          # rubocop:disable RSpec/AnyInstance
+          allow_any_instance_of(Notification).to receive(:notification_file_imported?).and_return true
```
frankieroberto: This doesn’t seem ideal. Can you avoid this by setting up a notification in this state, eg something like:

```
let(:notification_with_file_imported) { create(:notification, components: [create(:component)], responsible_person: responsible_person, state: :notification_file_imported)
  }
```

...and then specifying the component associated with this notification in the params?

baphled: I played with that last night but it's a couple steps removed from the actual tests.

IMHO, we'd be hiding a code smell that should be address as soon as possible. It also makes it harder to reason about the actual request and the expectations of it (this tests requires that `Notification#notification_file_imported?` returns `true`.

When you read the request params, you'll need to dig deeper to understand why we've setup the payload in the way we have.

frankieroberto: I’m personally a bit wary of stubbing methods like this, as it ties the unit test to the implementation, which can make the tests brittle (which makes refactoring harder as you then also have to update the tests). In this instance, `notification_file_imported?` simply checks that the `state == 'notification_file_imported'` on the notification object, so I think it should be just as easy to set that the setup rather than stubbing it?

I think it’s also worth a bit of digging into why you're having to stub on any instance of the Notification class, rather than the specific one set up in the `let` statement.

baphled: I agree, that I prefer to setup rather than stub, but we're not working with ideal code. So this seems the most pragmatic and expressive way to express that.

The reason why we're having to do this is because we've got tied dependencies that we should identify and gradually move away from.

The thing about setting up the environment in the way we'd like is that it'll hide the main issue we've got. Decoupled objects.

At least with the current approach it's obvious that we're having to stub because we're checking a notification object that we have little control of.

Dealing with that will remove the need for both approaches. As it stands that's more work than is required for this piece.

PR #1378, opened 2019-10-09
Add extra validation rule for component pH

frankieroberto:
The difference between minimum pH and maximum pH cannot be greater than 1.

Part of https://regulatorydelivery.atlassian.net/browse/COSBETA-600


frankieroberto:
@baphled agreed – I only discovered that towards the end of this work. Doh. One for the future.

PR #1382, opened 2019-10-10
Make the service runnable without Docker.

baphled:
## Description
Add `dotenv-rails` so that we can pass in environment variables to the service, without having to use Docker.

Also add `pry` for debugging.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/Gemfile
```diff
@@ -36,4 +36,7 @@ group :development, :test do
   gem "simplecov-console", "0.5.0"
   gem "slim_lint", "0.17.0"
   gem "solargraph", "0.33.2"
+
+  gem "pry"
```
slorek: Is there a reason to put them at the bottom in a separate list instead of alphabetically in the main list? Only asking because Rubocop expects gems to be listed alphabetically, and I think it can be helpful sometimes to find where things are located.

baphled: Bad habits 🤔🤣

Agreed

baphled:
> Consider adding an `.env.example` file as well.

Good catch, I thought we’d already addressed that.

Will patch 👌

cosmetics-web/Gemfile
baphled: Tbf, I think it’s because I kept it there locally when we were doing the whole stash and pop dance 🤣

frankieroberto:
@baphled @slorek the `env.example` is here: https://github.com/UKGovernmentBEIS/beis-opss/pull/1360 – will merge now.

baphled:
@frankieroberto cool, I'll remove `.env.example` from here. Once #1360 is merged, we can rebase this with those changes.

slawosz:
@baphled please take a look here: https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/3

I would add `.env.development.example` and `.env.test.example` and exclude `.env.development` and `.env.test` in .gitignore.

cosmetics-web/Gemfile
```diff
@@ -36,4 +38,5 @@ group :development, :test do
   gem "simplecov-console", "0.5.0"
   gem "slim_lint", "0.17.0"
   gem "solargraph", "0.33.2"
+  gem "sprockets", "4.0.0.beta8"
```
slorek: 4.0.0 final is out but Brakeman is complaining about it. I decided to keep the sprockets version unchanged in my PR for now. Not sure if this version is stable.

baphled:
> UKGovernmentBEIS/beis-opss-psd#3

@slawosz what's the benefit of having multiple `.env.*.example`?

I can understand the reasoning, if we've got slight changes dependant on environment but not so much when the environment variables are all the same regardless of environment.

We could quite easy do `cp .env.example .env.development` or `cp .env.example .env.test`

cosmetics-web/Gemfile
baphled: Yeah, that stumped me, it's pretty new.

I've pinned the version to 3.7.2 for now, as it seems to be recommended.

cosmetics-web/Gemfile
```diff
@@ -55,4 +57,5 @@ group :development, :test do
   gem "simplecov-console", "0.5.0"
   gem "slim_lint", "0.17.0"
   gem "solargraph", "0.33.2"
+  gem "sprockets", "3.7.2"
```
slorek: Would be worth adding a comment to explain why this is explicitly pegged to 3.7.2.

baphled: hmm, maybe.

Although I would ask myself why do we have a comment here but not for the other pinned gems, if I were looking at this from another readers perspective 🤔 🤷‍♂

If it's just to highlight that we're doing so to resolve a security issue. Maybe getting dependabot or hakiri back, so that we're updating dependancies when external issues are resolved automatically.

slorek: Normally you would not need to specify `sprockets` at all as it’s a dependency of `rails` so it is odd to do so here. I do agree we need to fix some of the dependency versioning. I just discovered today the reason why ElasticSearch is pegged to 6.x, but we have to resolve security vulnerabilities.

baphled: Yeah, I'd agree. We only really need to pin sprockets, or define it, when something like this happens.

There's loads of nice little treats in here, eh 😆

PR #1388, opened 2019-10-14
Update pH question for components

frankieroberto:
This updates the pH question for components from a Yes / No question:

Screenshot 2019-10-14 at 14 53 15

...to a 4 answer question:

Screenshot 2019-10-14 at 14 52 41

This is to make the question easier to understand, and to enable users to specify some components as not having a pH at all.

Where files are uploaded from CPNP, the answers to the various component-level pH questions are mapped to one of these 4 answers, or else specified as "not given" if no pH question was asked.

See https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-602

Review app: https://cosmetics-602-update-ph-question-submit-web.london.cloudapps.digital


frankieroberto:
This is ready for peer review now. Comments, suggestions and questions welcome.

I'm going to work with @slawosz on Wednesday to try and get a review app working so that the change can be reviewed in-browser. 😄

cosmetics-web/app/models/component.rb
```diff
@@ -16,6 +16,14 @@ class Component < ApplicationRecord
   has_one :nano_material, dependent: :destroy
   has_one_attached :formulation_file
 
+  enum ph: {
```
baphled: I'm not sure we need to `enum` here.

I think I went down this rabbit how a week or so ago.

baphled:
Apart from the `enum` I think we're pretty much good to go.

cosmetics-web/app/controllers/trigger_questions_controller.rb
```diff
@@ -195,6 +196,19 @@ def initialize_step
     end
   end
 
+  def update_component_ph_range
```
slorek: We could consider refactoring this to reduce the nested logic. For example:

```ruby
def update_component_ph_range
  return re_render_step unless @component.update_with_context({ ph: params.fetch(:component, {})[:ph] }, :ph)
  return skip_question unless @component.ph_lower_than_3? || @component.ph_above_10?
  redirect_to wizard_path(:ph)
end
```

slorek: To me it is easier to understand because you can follow the order of events more easily. But happy to go with the majority here.

slorek: You could also write the above suggestion using `if` instead of `unless` if that makes it easier.

cosmetics-web/app/controllers/trigger_questions_controller.rb
```diff
@@ -195,6 +196,19 @@ def initialize_step
...
+    if @component.update_with_context({ ph: params.fetch(:component, {})[:ph] }, :ph)
+
+      if @component.ph_lower_than_3? || @component.ph_above_10?
```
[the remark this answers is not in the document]
frankieroberto: Good idea. I’ve done this, which I think this best captures the business logic:

```ruby
  def ph_range_not_required?
    ph_between_3_and_10? || ph_not_applicable?
  end
```

cosmetics-web/app/controllers/trigger_questions_controller.rb
frankieroberto: Good debate!

Here’s my suggested compromise:

```ruby
  def update_component_ph_range
    return re_render_step unless @component.update_with_context(ph_param, :ph)

    if @component.ph_range_not_required?
      skip_question
    else
      redirect_to wizard_path(:ph)
    end
  end
```

This avoids the nested conditions, keeps the condition for the branch based on the answer given to the question, but returns early from the validation error which happens if no answer given. A bit inconsistent perhaps, but in my head that makes sense as the validation error is less likely and outside of the normal flow.

I did try:
```ruby
@component.update_with_context(ph_param, :ph) or return re_render_step
```
...which I think is more readable than the `unless`, but it's banned by our Rubocop settings. 😞

frankieroberto:
@baphled you didn't say what your objection to the `enum` is? I think it’s the best way to represent a 4-option question. We could just store this as a string, but declaring the enum in Rails gives us a few extra boolean methods (eg `ph_not_applicable?`) and stops us from accidentally saving a different value to the database.

I did try also creating an enumerated type within Postgres using [this type of migration](https://edgeguides.rubyonrails.org/active_record_postgresql.html#enumerated-types) but sadly our version of Rails can't save that to the `schema.rb` file (we'd have to switch to the `schema.sql` option I think).

PR #1389, opened 2019-10-14
Set the SUBMIT_HOST and SEARCH_HOST env vars

frankieroberto:
This updates the worker deploy script to set the `SUBMIT_HOST` and `SEARCH_HOST` environment variables, which are now required by the application (eg in the `routes.rb` file).

### DIFF
--- a/cosmetics-worker/deploy.sh
+++ b/cosmetics-worker/deploy.sh
@@ -9,6 +9,19 @@ set -ex
 #
 # SPACE: the space to which you want to deploy
 
+DOMAIN=cosmetic-product-notifications.service.gov.uk
+if [[ $SPACE == "prod" ]]; then
+    SUBMIT_HOSTNAME=submit
+    SEARCH_HOSTNAME=search
+elif [[ $SPACE == "research" ]]; then
+    DOMAIN=london.cloudapps.digital
+    SUBMIT_HOSTNAME=cosmetics-research
+    SEARCH_HOSTNAME=cosmetics-research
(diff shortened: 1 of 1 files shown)


slorek:
I think this can be more easily accomplished using CloudFoundry services, There is a script which will automatically pull them in as environment variables which persist across instances and deployments:

https://docs.cloudfoundry.org/devguide/services/user-provided.html#deliver-service-credentials-to-an-app

See also: https://github.com/UKGovernmentBEIS/beis-opss/blob/master/infrastructure/env/get-env-from-vcap.sh

It should just be a case of running a couple of `cf` commands rather than a code change.

frankieroberto:
@slorek doe you mean we'd just set them as user-provided environment variables (via something like `cf set-env cosmetics-work SEARCH_HOST=submit.cosmetic-product-notifications.service.gov.uk`)?

I think that would work too. I mainly did it this way to be consistent with the `deploy.sh` for  `cosmetics-web`, but it does seem a bit weird to me to be specifying this within the codebase.

slorek:
`set-env` is equivalent to what you did on your code change, i.e. it's isolated to one instance and will be lost when the instance is destroyed. User-provided services are created slightly differently using `cf cups`. See https://docs.cloudfoundry.org/devguide/services/user-provided.html#deliver-service-credentials-to-an-app. Basically the script will take the name you specify as the name of the environment variable. You then need to bind the service to the application and it should then persist across deployments.

PR #1390, opened 2019-10-14
Run the static analysis outside of Docker

frankieroberto:
Hoping this will speed things up.


baphled:
We're assuming that these commands will successfully run, which will not always be the case 🤔

frankieroberto:
@baphled if they don't run, they should exit with a non-zero exit code, which will get flagged as a fail, I think?

infrastructure/ci/static-analysis-if-changed.sh
```diff
@@ -1,17 +1,15 @@
...
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT yarn eslint app config
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT yarn sass-lint -vq 'app/**/*.scss'
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/brakeman --no-pager
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
-else
-    echo 'Static analysis not required.'
-fi
+cd cosmetics-web
+
+bundle install
```
baphled: What happens when this fails? 🤔

infrastructure/ci/static-analysis-if-changed.sh
```diff
@@ -1,17 +1,15 @@
...
+bundle exec rubocop
+bundle exec slim-lint app
+
+yarn install
```
baphled: Not sure we need to do a `yarn install` 🤷‍♂

baphled: If so, don't we need to make sure it's been successfully executed before the below commands have been run?

infrastructure/ci/static-analysis-if-changed.sh
```diff
@@ -1,17 +1,15 @@
...
+yarn eslint app config
+yarn sass-lint -c .sasslint.yml -vq 'app/**/*.scss'
+
+bundle exec brakeman --no-pager
+echo "Static analysis complete - no issues found"
```
baphled: Won't this hide the failing commands preceding this?

infrastructure/ci/static-analysis-if-changed.sh
```diff
@@ -1,17 +1,15 @@
 #!/usr/bin/env bash
 set -ex
```
baphled: Okay, `set -ex` will help fail fast. I should have spotted that.

I'm still not sure we need to run `bundle install` or `yarn install` though.

All deps _should_ be installed and in `Gemfile.lock` and `packages.json` when we push changes.

infrastructure/ci/static-analysis-if-changed.sh
```diff
@@ -1,17 +1,15 @@
...
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/rubocop
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/slim-lint app
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT yarn eslint app config
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT yarn sass-lint -vq 'app/**/*.scss'
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml exec $COMPONENT bin/brakeman --no-pager
-    docker-compose -f docker-compose.yml -f docker-compose.ci.yml down
-else
-    echo 'Static analysis not required.'
-fi
+cd cosmetics-web
```
slorek: Use `$COMPONENT` here? Unless we want to rip this out.

infrastructure/ci/static-analysis-if-changed.sh
slorek: Maybe `bundle check` would be preferable?

infrastructure/ci/static-analysis-if-changed.sh
slorek: Probably not necessary to run `yarn install`.

infrastructure/ci/static-analysis-if-changed.sh
slorek: Should exit before this if any preceding tests fail.

infrastructure/ci/static-analysis-if-changed.sh
frankieroberto: Good point. I forgot that that is part of the build process which runs before this script.

infrastructure/ci/static-analysis-if-changed.sh
frankieroberto: I've removed this now.

frankieroberto:
@slorek @baphled thanks for your comments. I've made a few more changes.

`bundle install` and `yarn install` are required in order to install the gems rubocop, brakeman, slim-lint and the node modules eslint and sasslint. Previously these were installed within the docker container, but because the static analysis tests are no longer using that, they have to be installed separately.

We could install these directly (eg via `gem install rubocop` and `npm install eslint --global`) which would be faster, as there would be a lot less to install, but might be harder to maintain. I think using bundler and yarn is probably simpler.  I have removed these install commands from the `static-analysis.sh` and into a separate install step within the Travis config. This has the advantage of minimising the install steps within the report on Travis, and also makes it quicker to run the `static-analysis.sh` script locally.

So far, these improvements have resulted in the static analysis job on Travis taking around ~4 mins instead of ~10 mins.

PR #1392, opened 2019-10-15
Allow users to complete questions for multi-item notifications.

baphled:
## Description
When a user uploads a bulk file, they are only able to fill in the notification data for one of those items instead of all of them.

This PR address this, so that the user is successfully looped through
each multi-item. until they've all been completed.

As this functionality was introduced, for free, by COSBETA-560, this PR introduces a few specs to make sure that the behaviour is captured in test.

It also makes sure that we're getting the components in the order we correctly expect (which is in reverse at present).

## Checklist:


nasirkhanbeis:
tested in review app, looking good as expected!

PR #1395, opened 2019-10-17
Remove trigger rule questions

frankieroberto:
This removes a long list of questions about ingredients ("trigger rule questions") as these are now covered by a single "Upload list of poisonous ingredients" step.

Review app: https://cosmetics-remove-trigger-rules-submit-web.london.cloudapps.digital

https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-651
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-650


slorek:
I'm wondering why there is a separate `trigger_questions_controller` when it simply follows on from the questions in `component_build_controller`. I guess we don't need to fix this now but it becomes even more silly when there's only two questions.

We need to ensure the test changes I made in `trigger_questions_controller_spec` in #1397 are not lost in the deletion after merging.

frankieroberto:
@slorek yeah, it no longer makes any sense to have the two controllers - especially as the `TriggerQuestionsController` now no longer has any interaction with the `TriggerQuestion` models... 😂

slorek:
#1397 is now merged so please rebase and bring in my changes to the trigger questions controller ;)

cosmetics-web/spec/requests/trigger_questions_requests_spec.rb
```diff
@@ -0,0 +1,183 @@
...
+      end
+
+      it "includes the question" do
+        expect(response.body).to include('What is the pH range of the product?')
+      end
+
+      context "with a frame formulation component" do
+        let(:component) { create(:predefined_component, notification: notification) }
+
+        it "includes a back link to the 'Poisonous ingredients' question" do
```
slorek: This should be the upload_formulation page now.

frankieroberto: Good spot. Think this is done now.

ralph-hawkins:
👍 Looks good to me

slorek:
Sorry, I just spotted on the product details page that this template https://github.com/UKGovernmentBEIS/beis-opss/blob/master/cosmetics-web/app/views/notifications/_trigger_question_details.html.slim is rendered which will display the trigger question answers when viewing a previously created notification. I think we need to check and test that this page renders correctly and remove the redundant code from there.

slorek:
Ignore the above comment; I think this is desired behaviour still for historical/bulk uploads.

frankieroberto:
@slorek yeah, I think that's still required for the bulk upload notifications (which do still generate trigger question rows in the database).

PR #1397, opened 2019-10-17
Frame formulation ingredients

slorek:
## Description

If the user selects predefined formulation type and containing poisonous ingredients they will be redirected to the upload ingredients page. The title will be changed to "Upload list of poisonous ingredients".

Demo video:

ezgif com-video-to-gif (1)

I didn't create a review app as the above video is sufficient to demonstrate the changes.

I've also fixed some broken back/forward wizard logic and added tests around the immediate pages before/after.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/spec/controllers/component_build_controller_spec.rb
```diff
@@ -74,6 +75,48 @@
...
+      before { get(:show, params: params.merge(id: :upload_formulation)) }
+
+      render_views
+
+      describe "page title" do
+        context "with a component with predefined formulation" do
+          let(:component_type) { "predefined" }
+
+          it "sets the page title to poisonous ingredients" do
+            expect(response.body).to match(/<title>Upload list of poisonous ingredients .+<\/title>/)
```
frankieroberto: Should this be `Upload list of poisonous ingredients\s*`? ie only matching any whitespace characters at the end of the title, rather than any other characters?

baphled: These will become brittle when we have to change the page titles 🤔

cosmetics-web/spec/controllers/trigger_questions_controller_spec.rb
```diff
@@ -69,6 +71,42 @@
...
+              expect(response.body).to match(/\<a class="govuk-back-link" href=".+\/build\/contains_poisonous_ingredients">Back<\/a>/)
+            end
+          end
+        end
+
+        context "when the component is not of predefined formulation" do
+          let(:component_type) { "exact" }
+
+          it "links to the upload formulation page" do
+            expect(response.body).to match(/\<a class="govuk-back-link" href=".+\/build\/upload_formulation">Back<\/a>/)
```
baphled: Could possibly use url_helpers here, as they're less likely to change.

cosmetics-web/spec/controllers/component_build_controller_spec.rb
slorek: Yes, I did check the locale file but it appears we're not using it at the moment, all the text being hard coded in the view. I think we should look at that separately.

cosmetics-web/spec/controllers/component_build_controller_spec.rb
```diff
@@ -74,6 +75,48 @@
...
+          end
+        end
+      end
+
+      describe "back link" do
+        context "with a component with predefined formulation" do
+          let(:component_type) { "predefined" }
+
+          it "links to the poisonous materials page" do
+            expect(response.body).to match(/\<a class="govuk-back-link" href=".+\/build\/contains_poisonous_ingredients">Back<\/a>/)
```
frankieroberto: I’d rather use some sort of html/css selectors for these kind of tests, so they're a little less brittle (eg if we add classes to the html). But let's figure that out separately.

cosmetics-web/app/helpers/component_build_helper.rb
```diff
@@ -3,4 +3,28 @@ def cmr_errors(component)
     component.cmrs.each_with_index.flat_map { |cmr, index| cmr.errors.map { |attr, msg| { text: msg, href: "#component_cmrs_attributes_#{index}_#{attr}" } } } +
       component.errors.reject { |attr, _| attr.to_s.include? "cmrs." }.map { |_, msg| { text: msg, href: "#component_cmrs_attributes_0_name" } }
   end
+
+  def previous_wizard_path
+    previous_step = get_previous_step
+    previous_step = previous_step(previous_step) if skip_step?(previous_step)
+
+    if step == :add_component_name
+      responsible_person_notification_build_path(@component.notification.responsible_person, @component.notification, :add_new_component)
```
frankieroberto: Should we pass in `@component` and `@category` to the helper as method arguments rather than relying on them being present as instance variables?

cosmetics-web/app/views/component_build/upload_formulation.html.slim
```diff
@@ -1,4 +1,4 @@
-- title = "List of product ingredients"
+- title = @component.predefined? ? "Upload list of poisonous ingredients" : "List of product ingredients"
```
ralph-hawkins: I know this is slightly outside the scope of the story but for consistency, it would be better if the exact and ranges versions of this page had an h1 of "Upload list of ingredients".

cosmetics-web/spec/controllers/component_build_controller_spec.rb
frankieroberto: I personally would expect tests to fail if the page titles change. It does make them more brittle, but the tests will fail fast and be easy to update. Plus it makes them more readable, and allows you to refactor or reorganise the locale file without breaking the tests.

cosmetics-web/app/views/component_build/upload_formulation.html.slim
slorek: @ralph-hawkins OK I've changed it.

slorek:
Review app available at https://cosmetics-frame_formulation_ingredients-search-web.london.cloudapps.digital/

cosmetics-web/spec/controllers/component_build_controller_spec.rb
slorek: No it appends the service name at the end.

cosmetics-web/app/helpers/component_build_helper.rb
slorek: I don't think we can reliably. This is a method used by the `wicked` wizard gem, not necessarily called directly.

cosmetics-web/spec/controllers/component_build_controller_spec.rb
frankieroberto: Ah yeah, I forgot...!

PR #1405, opened 2019-10-18
Don't show product ingredients to MSA users

slorek:
https://regulatorydelivery.atlassian.net/browse/COSBETA-631

This ensures that market surveillance authority users can't view product ingredients.

ezgif com-video-to-gif (2)

As demonstrated by the video, the user who created the notification can still see them.

Poison centre users view the product on a different set of templates which haven't been altered here.

The diff on the view templates is a bit over the top; I mostly just added indentation beneath the new conditionals. Slim 🤷‍♂ 

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

screencapture-search-cosmetics-3002-notifications-54252960-2019-10-21-10_31_42

### DIFF
--- a/cosmetics-web/app/views/notifications/_component_details.html.slim
+++ b/cosmetics-web/app/views/notifications/_component_details.html.slim
@@ -1,4 +1,4 @@
-- if current_user.can_view_product_ingredients? && component.notification.notified_post_eu_exit?
+- if component.notification.notified_post_eu_exit?
   tr.govuk-table__row
     th.govuk-table__header[scope="row"]
       | Contains CMR substances
@@ -11,7 +11,6 @@
         td.govuk-table__cell
           = render "none_or_bullet_list", entities_list: component.cmrs.map(&:display_name)
 
-- if current_user.can_view_product_ingredients?
   tr.govuk-table__row
     th.govuk-table__header[scope="row"]
       | Nanomaterials
@@ -85,11 +84,14 @@ tr.govuk-table__row
                   class: "govuk-link--no-visited-state"
         - else
           | N/A
-  tr.govuk-table__row
-    th.govuk-table__header[scope="row"]
-      | Physical form
-    td.govuk-table__cell
-      = get_physical_form_name(component.physical_form)
+
+tr.govuk-table__row
+  th.govuk-table__header[scope="row"]
+    | Physical form
+  td.govuk-table__cell
+    = get_physical_form_name(component.physical_form)
+
+- if current_user.can_view_product_ingredients?
   tr.govuk-table__row
     th.govuk-table__header[scope="row"]
       | Special applicator
--- a/cosmetics-web/app/views/notifications/_component_details.html.slim
+++ b/cosmetics-web/app/views/notifications/_component_details.html.slim
@@ -1,4 +1,4 @@
-- if component.notification.notified_post_eu_exit?
+- if current_user.can_view_product_ingredients? && component.notification.notified_post_eu_exit?
   tr.govuk-table__row
     th.govuk-table__header[scope="row"]
       | Contains CMR substances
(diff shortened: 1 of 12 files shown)


cosmetics-web/app/helpers/notification_helper.rb
```diff
@@ -24,4 +24,8 @@ def product_import_country(notification)
       "EU (before Brexit)"
     end
   end
+
+  def can_view_product_ingredients?
```
frankieroberto: Is there any reason this can’t be a method of the User model?

cosmetics-web/app/views/notifications/_component_details.html.slim
```diff
@@ -1,4 +1,4 @@
-- if component.notification.notified_post_eu_exit?
+- if can_view_product_ingredients? && component.notification.notified_post_eu_exit?
```
frankieroberto: Could this be `current_user.can_view_product_ingredients?` if we expose `current_user` to the views in `ApplicationController` via [`helper_method`](https://api.rubyonrails.org/classes/AbstractController/Helpers/ClassMethods.html#method-i-helper_method)?

slorek: There is no `current_user`, only `User.current` 😂  We could create a helper, but I don't think  it's any better.

cosmetics-web/app/helpers/notification_helper.rb
slorek: Good point.

cosmetics-web/app/views/notifications/_component_details.html.slim
slorek: Actually I see what you mean 👍

cosmetics-web/app/helpers/notification_helper.rb
frankieroberto: 👍

cosmetics-web/app/views/notifications/_component_details.html.slim
frankieroberto: 👍

ralph-hawkins:
I think 'Minimum pH' and 'Maximum pH' also need to be hidden. I also think that 'Still on the market' needs to be hidden? But I don't know if we understand how that works @tobyhughes ?

nasirkhanbeis:
Still on market needs to be removed from check your answers page anyway so will apply default to  MSA product view page.

tobyhughes:
@ralph-hawkins 'still on the market' is a bug appearing everywhere at the moment. There is a ticket to remove it from businesses product view, but not sure if this will automatically remove it from the msa and NPIS product view... Any ideas @slorek ?

Also think we need to add:

i) nanomaterial information
ii) CMR information
iii) physical form
iv) contact person information 
v) the generated reference number (if possible)

to the MSA product view

slorek:
@tobyhughes Business and MSA view uses the same template, so by default any change will affect both users. Poison centres have a different view template, so changes to MSA/business view will not affect them, and vice versa. Does that make sense?

tobyhughes:
@slorek that makes sense, thank you

slorek:
@toby do we need to show Special applicator, Applicator type, or Acute poisoning information? I did hide these as well.

slorek:
Confirmed reference number should be displayed already if present

slorek:
@toby contact person is currently being shown to poison centre users but not MSA users or business users. Which of these three should see it? Should they be able to see email, phone number as well as name?

tobyhughes:
@slorek 
We shouldn't show acute poisoning information to msa users.

RE special applicator - I don't think msa users can currently see this in CPNP (the screenshots we have don't show it, could be because there was no special applicator in the notification), so we should probably mirror this for now and not show it. (however, worth keep this in mind for the future as this sounds like it could be useful @ralph-hawkins ?)

just checked a product notification and looks like it is being surfaced to business users too? 
image

should be shown to all users, I think

slorek:
@tobyhughes @ralph-hawkins I've made the following amendments:

* Added contact person details for MSA users
* Reinstated CMRs and nanomaterials and physical form
* Removed 'still on the market' from both MSA and business users (since it's an easy fix)

See updated screenshot below:

screencapture-search-cosmetics-3002-notifications-54252960-2019-10-21-16_15_35

Please could you check and approve if possible?

tobyhughes:
@slorek looks good to me :)

cosmetics-web/spec/controllers/poison_centres/notifications_controller_spec.rb
```diff
@@ -99,10 +99,68 @@
...
+        let(:notification) { create(:notification, :imported, :registered, :ph_values, components: [component], responsible_person: responsible_person) }
+
+        render_views
+
+        it "renders contact person overview" do
+          expect(response.body).to match(/Contact person/)
+        end
+
+        it "does not render product imported status" do
+          expect(response.body).not_to match(/Imported/)
```
baphled: If any of these copy based expectations change, from a UI perspective, we'll need to update here as well.

We could solve this by using CSS classes, or IDs 🤔 

ie: `expect(response.body).to has_css?('#contact-details')` or `expect(response.body).to has_xpath?('x-path-to-element')`.

slorek: Agreed, but I'm not sure attribute assumptions are any more reliable as they can be changed in the template just as easily as copy.

In the past I have put copy text in the `I18n` locale file, referencing keys in the tests, so that the tests automatically get updated if the copy changes. I think that's the safest option, but I didn't want to expand the scope of this ticket too much.

baphled: Agreed that I18n is ideal. However we'll likely change copy soon and the more we introduce tests like this the more housekeeping we'd need to do.

So pre-emptively using attributes matching, until we move to I18n keys, would be reduce churn in the nearer future.

PR #1408, opened 2019-10-21
Don't pull the antivirus docker image on CI

frankieroberto:
Not sure this is needed, as none of the tests test it?

(Don't merge this yet)


slorek:
According to the [build log](https://travis-ci.org/UKGovernmentBEIS/beis-opss/jobs/600793349) on Travis, instead of pulling the image it is building it from source, so I don't think this is actually solving the problem.

frankieroberto:
@slorek no, not yet...  The build & test process is more complicated than I realised...

PR #1409, opened 2019-10-21
Allow antivirus to be disabled

frankieroberto:
Currently the application requires an antivirus API to be available, which is provided by a separate Docker image.

For the ease of local development and running integration tests, this allows the anti-virus check to be disabled by using the `ANTIVIRUS_ENABLED=false` environment variable.

It remains enabled by default.


cosmetics-web/app/analyzers/read_data_analyzer.rb
```diff
@@ -9,7 +9,11 @@ def initialize(blob)
   end
 
   def self.accept?(given_blob)
-    return false unless given_blob.present? && given_blob.metadata["safe"]
+    return false if given_blob.blank?
+
+    if ENV.fetch('ANTIVIRUS_ENABLED', 'true') == 'true'
+      return false unless given_blob.metadata["safe"]
```
baphled: Are we sure we always have the `safe` key within `given_blob.metadata`.

cosmetics-web/config/initializers/storage.rb
```diff
@@ -2,7 +2,11 @@
 # to contain just a MasterAnalyzer. This Analyzer will decide which of the other analyzers to use
 # document_analyzers is a list of the analyzers that we will use
 Rails.application.config.document_analyzers = Rails.application.config.active_storage.analyzers
-Rails.application.config.document_analyzers.append AntiVirusAnalyzer
+
+if ENV.fetch('ANTIVIRUS_ENABLED', 'true') == 'true'
```
baphled: What happens when we need to use the AV API for our integration tests. It seems like it'd be more prudent to mock out the requests via webmock/VCR and the like.

This may also hide issues we've spotted when it comes to trying to test bulk upload locally, which seems to happen when we append both `AntiVirusAnalyzer` and `ReadDataAnalyzer` to `Rails.application.config.document_analyzers`.

At the very least if seems like it'd be more flexible to mock those requests or have a config that we can change within our tests.

frankieroberto: We already mock the API within the integration tests, and could continue to do so. The aim of this mainly to enable us to turn it off for local development, as running the antivirus Docker image is such an overhead.

cosmetics-web/app/analyzers/read_data_analyzer.rb
frankieroberto: That's what the existing code assumes - the key is added by the antivirus analyzer. This just removes that condition if you've set `ANTIVIRUS_ENABLED=false`.

baphled: I appreciate that, although if the AV changes this, which we seem to have little control over, we'll have to do some digging to work out what's wrong.

May if it's not present we could throw an exception with a helpful error message. Or if it's not vital, just return false 🤷‍♂

frankieroberto: I'm a bit confused by your comment... 😕 

We need to continue to check that "safe" has been added to the metadata (which is done by our code [here](https://github.com/UKGovernmentBEIS/beis-opss/blob/master/cosmetics-web/app/analyzers/anti_virus_analyzer.rb#L13)), in order to make sure the files have been scanned for viruses before being parsed by our XML parser.

This change just no longer applies this check if you've set `ANTIVIRUS_ENABLED=false` so that we can run the application and the Sidekiq queue locally without having the antivirus running.

baphled: Ah, cool, if we have control over it, that's slightly more comforting.

We should still not assume that we're getting the metadata in this format though. We should make sure that it actually exists instead of expecting it to.

frankieroberto: I still don't quite follow. Are you suggestion we should be checking the actual value in the Hash rather than just that there's a `safe` key?

cosmetics-web/config/initializers/storage.rb
baphled: We only really need AV for bulk upload, unless I'm missing something, if so are we sure this is actually worth it. 

Especially via an environment variable, where we'd have to restart the service to get AV running again.

frankieroberto: I _think_ all files uploaded with ActiveStorage are analyzed with the antivirus scanner, including images – as the antivirus analyzer returns `true` to `accept?(_blob)`.

cosmetics-web/app/analyzers/read_data_analyzer.rb
frankieroberto: Perhaps. I've not looked at what values are actually stored under that key. This PR doesn't change the existing condition logic, it just gives developers the opportunity to turn the condition off. We could review how the antivirus scanner does & should work as a separate task?

PR #1412, opened 2019-10-21
Moved view based specs

baphled:
WIP

## Description
To make it clear where our view functionality is coming from. This PR moves the recent text based expectations to their appropriate view specs.

It also refactors those specs so that we rely upon css selectors instead of text.

Ideally we'd likely want to use `I18n` keys to refer to text within our tests but for now this makes sure that we have the expected elements whilst not relying on copy, which is still being refined.

## Checklist:


slorek:
We discussed test approach in the last meeting and settled on an integration test approach. I didn’t convert the existing controller tests to integration tests as part of the changes but it wouldn’t take a lot to do so. I think I’d prefer we did that than move the tests to the view layer, unless we also have integration testing but not instead of.

PR #1414, opened 2019-10-22
Ask users CMR questions regardless of pre or post brexit

baphled:
## Description
Regardless of whether a user is notifying on a product pre or post brexit they should be asked whether their product contains CMRs.

This PR changes the pre-brexit journey to make sure that this now happens.

## Checklist:

### DIFF
--- a/cosmetics-web/app/controllers/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/component_build_controller.rb
@@ -183,11 +183,7 @@ private
 
   def render_add_cmrs
     if @component.update_with_context(component_params, step)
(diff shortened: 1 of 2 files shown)


slorek:
Are there any back button links which need to be amended in the following steps?

slorek:
It seems like `post_eu_exit_steps` is not called anywhere in `component_build_controller` already so I'm not sure this actually changes anything.

baphled:
@slorek FYI: https://github.com/UKGovernmentBEIS/beis-opss/blob/master/cosmetics-web/app/controllers/concerns/manual_notification_concern.rb#L61

PR #1415, opened 2019-10-22
Deploy directly to production

slawosz:
Production will be deployed only when staging will be deployed.

https://docs.travis-ci.com/user/build-stages/#how-do-build-stages-work
```
Stages group jobs that run in parallel and different stages run sequentially.
```


frankieroberto:
@slawosz looks good! Might be worth removing the `Deploy to int` job completely as this would no longer be triggered?

PR #1428, opened 2019-10-24
Add CMRs to the "Check your answers" page for all products

baphled:
## Description
As we are now including CMR questions regardless of whether a product is pre or post brexit. We need to update the "Check your answers" page to reflect this.

This PR removes the check and displays CMR related product information.

## Checklist:


baphled:
We should be pragmatic in our approach to testing.

Integration tests for a view change seems like overkill in this situation, especially as we wish to deploy this and a following ticket that are all related.

slorek:
I'm concerned with having consistency of approach to testing and getting the most value. We don't currently have any view tests, and while they might be a bit faster than a request spec I don't think we derive nearly as much value from them.

baphled:
Replaced view spec with a request spec, although it takes twice the time and can be harder to reason about when it comes to determining whether we have a bug or whether something as simple as copy has changed.

cosmetics-web/spec/requests/check_your_answers_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      it "should have a UK cosmetics product number " do
+        expect(response.body).to include("UK cosmetic product number:")
+      end
+
+      it "displays that the product contains CMRs" do
+        expect(response.body).to include("Contains CMR substances")
+      end
+
+      it "displays whether CMRs are present" do
+        expect(response.body).to include("Yes")
```
frankieroberto: Feels a bit weird to be just asserting that "Yes" is included anywhere on the page... How about creating a custom assertion which checks that "Yes" is in the same check-your-answers 'row' as "Contains CMR substances"?

Something like:

```ruby
expect(response.body).to have_summary_list_item(key: "Contains CMR substances", value: "Yes")
```

Could use Nokogiri to parse the HTML and check that the key and value appear within `<dt>` and `<dd>` elements within the same parent element? (See [Summary list component](https://design-system.service.gov.uk/components/summary-list/)) 

– would be super useful as we come to work more on the check your answers page...

baphled: Agreed, sorry I should have waited to appease rubocop 🤦‍♂ 

I'm not super comfortable with asserting text on a page, in any way tbf.

Ideally we should be using I18n instead of relying on actual copy, but that's a bit much atm. So I was about to change these to XPath/CSS matchers, for now.

I'm all for custom matchers, once we formalise how we're testing our pages, It'd be a useful addition.

PR #1436, opened 2019-10-28
Don't rescue from all errors when parsing CPNP files

frankieroberto:
Currently if there are _any_ errors when parsing the CPNP files, including syntax errors in our code, we rescue from them and just add an "uknown_error", which displays as:

> Try again or manually enter the production notification data

This means that Sentry isn’t notified of the exception.  By removing this, we'll be alerted to these errors earlier, and can then either fix them or add more specific errors that will make more sense to users.

### DIFF
--- a/cosmetics-web/app/analyzers/read_data_analyzer.rb
+++ b/cosmetics-web/app/analyzers/read_data_analyzer.rb
@@ -84,9 +84,6 @@ private
     rescue CpnpFileNotifiedPostBrexitError => e
       Sidekiq.logger.error e.message
       @notification_file.update(upload_error: :post_brexit_date)
-    rescue StandardError => e
-      Sidekiq.logger.error "StandardError: #{e.message}\n #{e.backtrace}"
-      @notification_file.update(upload_error: :unknown_error)
     end
     # rubocop:enable Metrics/BlockLength
   end


slorek:
I think we might prefer instead of surfacing the error to the user in an unfriendly way to send the exception manually, e.g. (from the [docs](https://docs.sentry.io/clients/ruby/)):

```
begin
  1 / 0
rescue ZeroDivisionError => exception
  Raven.capture_exception(exception)
end
```

frankieroberto:
@slorek well by removing the rescue the user wouldn’t see an explicit error at all – they'd just see the "Refresh your browser" message forever.  I guess the ideal might be some kind of "There was a problem processing your file. Please get in touch with the support team" type of message?

One of the annoying things about rescuing from StandardError is that it makes debugging harder when running the site locally or running the tests.

Maybe something like this would be the best of both worlds?

```ruby
rescue StandardError => exception
  if Rails.env.production?
    Raven.capture_exception(exception) 
    @notification_file.update(upload_error: :unknown_error)
  else
    raise exception
  end
end
```

frankieroberto:
Also, worth considering is that if we allow the exception to be raised, then Sidekiq will register an error and will keep trying for around 21 days. This should be enough time for us to fix the bug, and then the file will be either successfully processed (or else we'd add a more specific error message). If we rescue and return true, then the user will have to re-upload the file once we fix the bug.

slorek:
I guess the question is how useful an error after 21 days would be in this context.

frankieroberto:
@slorek hopefully we'd have fixed the bug by then! 😉   Maybe the ideal is to immediately display an error to the user, but to also fail the Sidekiq job allowing it to be retried and to report to Sentry?

slorek:
I think that would be ideal, but in the meantime if that's too time consuming surfacing the exception will at least help us find the problem.

PR #1438, opened 2019-10-28
Remove checkbox from declaration

frankieroberto:
This removes the need to check a checkbox in order to agree to the declaration:

## Before

Screenshot 2019-10-28 at 16 39 48

## After

Screenshot 2019-10-28 at 16 42 02

https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-690


tobyhughes:
@frankieroberto think it should say 'I confirm' for the business declaration page :)

Then 'I accept' for the MSA and poison centre declaration pages

PR #1445, opened 2019-11-01
Use `cf auth` instead of `cf login`

AP-Hunt:
## Description
This PR is raised following a conversation with @frankieroberto 

`cf login` is intended for use interactively, by humans. Whilst it can be used
in a non-interactive way, it requires the credentials to be exposed in
plaintext as part of the command being run.

Instead, `cf auth` [1] should be used when working with CF in an automated
fashion. It avoids the problems of `cf login` by sourcing its credentials from
the `CF_USERNAME` and `CF_PASSWORD` environment variables.

This commit replaces `cf login` with `cf auth` in every place I could find.

[1] https://cli.cloudfoundry.org/en-US/cf/auth.html

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


frankieroberto:
Re-opened as #1448 so that CI can run.

PR #1457, opened 2019-11-08
GitHub CI test

frankieroberto:
This adds a new workflow for GitHub to run the model and request specs.

In order to keep this fast by reducing the dependencies required, the tests are configured to use the 'async' (in-process thread pool) adapter rather than redis, and the requests specs mock Keycloak. Elasticsearch and the antivirus services also aren’t required.


cosmetics-web/spec/login_helpers.rb
```diff
@@ -3,6 +3,16 @@ module LoginHelpers
     match { |actual| items_to_match.include? actual }
   end
 
+  # rubocop:disable RSpec/AnyInstance
+  def mock_keycloak_as_not_signed_in
+    allow(Keycloak::Client).to receive(:user_signed_in?).and_return(false)
```
slorek: 🤔 should this be the default context unless the test explicitly signs the user in?

frankieroberto: Maybe? We'd just then have to skip this step or override it somehow for tests that do sign the user in (which is likely most of them)?

frankieroberto:
This is currently failing because the GitHub `setup-ruby` action doesn’t yet support 2.5.7: https://github.com/actions/setup-ruby/issues/36 😞

PR #1459, opened 2019-11-11
Allow Sidekiq queue name to be specified - especially to segregate review apps

slorek:
## Description
Port of https://github.com/UKGovernmentBEIS/beis-opss-psd/pull/62.

Currently all review app workers are sharing the same queue, so it is not possible to segregate them.

An immediately obvious example would be if you create a new scheduled job another review app worker can pick it up, but immediately fail because it doesn't have that job class available.

Other problems could occur when app logic is modified and it will not be possible to tell which review app processed the job.

This change separates the queues based on the review instance name, e.g. `pr-XX`.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/jobs/application_job.rb
+++ b/cosmetics-web/app/jobs/application_job.rb
@@ -1,3 +1,3 @@
 class ApplicationJob < ActiveJob::Base
-  queue_as :cosmetics
+  queue_as ENV["SIDEKIQ_QUEUE"] || "cosmetics"
 end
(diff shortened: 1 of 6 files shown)


frankieroberto:
@slorek I thought each review app has its own redis instance, see https://github.com/UKGovernmentBEIS/beis-opss/blob/master/infrastructure/ci/deploy-review-app.sh#L15 ?

slorek:
Ah thanks, I missed this. It's not implemented on PSD. I think we need a consistent approach, and provisioning a new Redis might be overkill IMO.

PR #1463, opened 2019-11-11
Fix html structure

frankieroberto:
This fixes the HTML tag nesting, as currently a `</body></html>` is inserted immediately after the header, due to the way that the templates are structured.

To fix this, I've split the `_head.html.slim` into `_head.html.slim` (which just generates the `<head>` tag and its contents) and a new `_header.html.slim` (which generates a `<header>` tag and its contents). The `<html>` and `<body>` tags move to the parent templates `application.html.slim` and `landing_page.html.slim`.


frankieroberto:
@slorek I'm pretty confident about this. The easiest way to see the difference is to run the HTML from production through https://html5.validator.nu and then do the same with the HTML from this PR. Both have errors, but the howler on production of having:

```html
</header></body></html><div class="govuk-main-wrapper">
```

is fixed.

PR #1467, opened 2019-11-12
Make sure bulk upload caters for nano-material that is not notified the UK

baphled:
## Description
This PR addresses making sure that nano-matieral, which are not notified within the UK, are classed as "incomplete".

This means that the only way that a user can complete a notification is when they have confirmed that toxicology has been notified within the UK, for all nano-material present.

As the main change here is to update the definition of "incomplete", for a nano-material, we enhance the 
## Checklist:

### DIFF
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -40,6 +40,6 @@ class NanoElement < ApplicationRecord
   end
 
   def usage_confirmed?
-    (confirm_usage.nil? || confirm_usage == "no")
+    confirm_usage.nil?
   end
 end
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -15,8 +15,8 @@ class NanoElement < ApplicationRecord
 
   def incomplete?
     purposes.blank? ||
-      (non_standard? && notified_toxicology?) ||
-      (standard? && restrictions_confirmed?)
+      (non_standard? && toxicology_incomplete?) ||
+      (standard? && restrictions_confirmed_incomplete?)
   end
 
   def standard?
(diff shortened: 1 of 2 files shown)


nasirkhanbeis:
@baphled  
just had a quick look, happy path appears to be good - however If I say “No” to subsequent annexes confirmation page and return to dash board, then it treats as nano-information is complete and jumps to subsequent frame formulation journey when I click “Add missing info”. Also as we discussed, we want user to to return to the beginning of nano-element rather continuing where they said no.

nasirkhanbeis:
@baphled  patch fix is looking good now

cosmetics-web/app/models/nano_element.rb
```diff
@@ -32,4 +26,25 @@ def standard?
   def non_standard?
     purposes.present? && purposes.include?("other")
   end
+
+protected
```
frankieroberto: These could be private methods rather than protected?

cosmetics-web/app/models/nano_element.rb
```diff
@@ -32,4 +26,25 @@ def standard?
...
+    confirm_toxicology_notified.nil? ||
+      confirm_toxicology_notified == "not sure" ||
+      confirm_toxicology_notified == "no"
+  end
+
+  def restrictions_confirmed_incomplete?
+    confirm_restrictions.nil? ||
+      (confirm_restrictions == "no" && toxicology_incomplete?) ||
+      (
+        confirm_restrictions == "yes" && usage_confirmed_incomplete? ||
```
frankieroberto: Add parentheses around `confirm_restrictions == "yes" && usage_confirmed_incomplete?` for consistency with the next line?

baphled: Yeah, that's valid 👍

These messages should probably be extracted soon, I'm just not sure we understand the logic properly yet.

Think I was mid-thought and gave up 😆

cosmetics-web/app/models/nano_element.rb
baphled: Yeah, I don't see much value in having this a protected either and private is a good way of getting a feel for what doesn't belong in an object 🤔

PR #1471, opened 2019-11-14
Fix back link on check your answers page

frankieroberto:
This updates the back link on the "Check your answers page" so that it takes you back to the right page. This could be one of:

* The upload image (for the product) page
* The pH radio-button question for the last component
* The min and max pH for the last component
* The dashboard (for products notified via ZIP file upload)

Review app: https://cosmetics-pr-1471-submit-web.london.cloudapps.digital


cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -38,6 +38,21 @@ def new
 
   # Check your answers page
   def edit
+    if @notification.notified_post_eu_exit?
```
slorek: Probably OK for now but as more logic is added for the other pages you might consider creating a helper method to determine the previous page to keep things clean.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -38,6 +38,21 @@ def new
...
+
+      # Last question is image upload page
+      @previous_page_path = responsible_person_notification_build_path(@notification.responsible_person, @notification, :add_product_image)
+    else
+
+      # Last component questions were answered for is the one with the highest ID
```
slorek: Is this true even when they go back and edit? Perhaps storing a variable in the session when viewing the edit page might be more reliable.

cosmetics-web/spec/requests/check_your_answers_page_spec.rb
```diff
@@ -62,5 +62,41 @@
         expect(response.body).not_to have_tag("td#cmr-names")
       end
     end
+
+    context "when the notification was post-Brexit" do
```
slorek: I can't see anything in the `params` which determines that the notification was post-Brexit. This might be implicit in the `draft_notification` factory, but I think it's important to set the context clearly within the context block to match the description as that could change. For instance I would explicitly set the date on the draft notification inside this context block, or create a new factory with clear naming and attributes to denote it. This ensures a meaningful test should the factory change and also informs the reader how a post-Brexit notification is created.

frankieroberto: Good idea. I'll add a trait.

cosmetics-web/spec/request_helpers.rb
```diff
@@ -0,0 +1,18 @@
...
+  match do |actual|
+    body = Nokogiri::HTML(actual).at_css("body")
+    back_link = body.at_css("a.govuk-back-link")
+    back_link && back_link["href"] == expected_path
+  end
+  failure_message do |actual|
+    body = Nokogiri::HTML(actual).at_css("body")
+    back_link = body.at_css("a.govuk-back-link")
+    if back_link
+      "expected a back to #{expected_path}, but was #{back_link['href']}"
```
slorek: *back link

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
frankieroberto: @slorek I've refactored slightly in d5d6edd494a853cfb0558b5d7af0e790d482460e to a private method, which hopefully improves readability a bit.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
frankieroberto: Good point. I'm not sure. I'll ask @nasirkhanbeis to check this within testing.

frankieroberto: This logic was wrong - multi-item notifications go back to the List items page.

frankieroberto:
@nasirkhanbeis I've updated this now so that for multi-item kits, the back link goes to the right page.

nasirkhanbeis:
@frankieroberto 
Multi-item journey looks good with the new fix. good stuff. 
Just a minor issue around single item journey - 
Steps to replicate 
1.On check your answer page, click back button, 
2.It will take to pH previous question, suppose if they entered range min max then its okay, 
3. Click back and on pH selection page, select "no it doesn't have pH" and continue
4. Now click back on check your answer page - It is expected to take back to pH selection page but not to range min max page which was in step 2.

frankieroberto:
@nasirkhanbeis good spot! I think this is actually a separate bug, which is that if you change your answer from having an explicit pH range to not needing a pH range (eg selecting between 3 - 10 or "it does not have a pH") then the previous explicit pH range isn't removed. I think I'll fix this as a separate ticket, and then that should make the back link work here too.

frankieroberto:
@nasirkhanbeis bug fixed as part of #1474. Could you review that? Once that's merged, I'll update this branch and the back button link should fix itself.

nasirkhanbeis:
Possible scenario on back button in check your answer page 
1. Single item manual journey pre-bexit - should return to pH question page - Passed( further back button breaks the journey- will be fixed in separate ticket)
2. Single Item manual journey post brexit- should return to Image upload page - Passed
3. Multi-item manual pre/post brexit -  should to component list page - Passed
4. Bulk upload pre brexit- single item  - should return to dashboard page  - Passed
5. Bulk upload  pre brexit - single item frame formulation upload  then check your answer page - should return to frame formulation page - Failed ( returning to pH question page) 
6. Bulk upload pre-brexit - multi-item  frame formulation - should return to frame formulation page of last item - Failed  (returns to component list page) 
7. Bulk upload pre-brexit - single item nano material - should return to nano-material question - Failed( returning to pH question)

nasirkhanbeis:
@frankieroberto  I summarised list of possible scenarios in my comment above, can you pls fix the failures mainly around bulk-upload  pls? thanks.

PR #1473, opened 2019-11-15
Don't render error pages within test specs (except where opted-in)

frankieroberto:
This changes the behaviour for tests so that error pages are no longer rendered, except for system tests.

This has the advantage that for request specs, if you've got a syntax error or typo in your code, you can see the exception within the console, rather than getting an error based on the application rendering a 500 error page.

The downside is that if we want to test errors (eg 404 not found or 403 access denied) we'd either have to assert that an exception was raised, or change this setting just for those tests so that we can assert based on the http response code or error page content.


frankieroberto:
@slorek hmm, so for some reason it doesn’t consistently change the setting back for the system specs, and I can't work out why. When I re-run the failing tests individually they pass, but they sometimes fail when run as part of the full suite. Any ideas?

frankieroberto:
@slorek ok so I figured out why it was behaving differently when running the full test suite vs an individual test – rails caches the config, so updating the config in a `before` step doesn't always have any effect. I’ve committed a workaround based on [this code](https://github.com/rspec/rspec-rails/issues/2024#issuecomment-420646336) which adds a method that makes changes to the Rails cached config directly (and then resets it after).

slorek:
Is it possible to do the config change using a Rspec shared context? I think this would be a cleaner interface. Are you saying that before/after blocks don’t have any effect? These should always be run whether as part of a full suite or individual test so if something isn’t working it points to a problem somewhere.

frankieroberto:
@slorek have refactored to use a shared context. Good idea, thanks. 👍

cosmetics-web/spec/system/error_page_spec.rb
```diff
@@ -3,30 +3,26 @@
 RSpec.describe "Error page", type: :system do
   let(:responsible_person) { create(:responsible_person) }
 
+  include_context "with errors rendered"
```
slorek: Nicer would be `RSpec.describe "Error page", type: :system, with_errors_rendered: true do`

frankieroberto: Ah, I didn't spot what `rspec.include_context` was for. Thanks.

frankieroberto: Okay, I've done that within 17b265c7ddbddea45ea4b85040c3db1a60c5e6ab

I've not added it to the top level of the `sign_in_spec.rb` file though, as only 2 out of 3 of those describe blocks need the errors to be rendered.

cosmetics-web/spec/system/sign_in_spec.rb
```diff
@@ -26,6 +26,7 @@
   end
 
   describe "a user associated with a Responsible Person account" do
+    include_context "with errors rendered"
```
slorek: you can do this on the `describe` block e.g.

```ruby
describe "a user associated with a Responsible Person account", with_errors_rendered: true do
```

cosmetics-web/spec/system/sign_in_spec.rb
```diff
@@ -44,13 +45,14 @@
...
     it "cannot access the Poison Centre notifications page" do
       visit poison_centre_notifications_url
-
       assert_current_path poison_centre_notifications_path
       assert_text "Access denied"
     end
   end
 
   describe "a Poison Centre user" do
+    include_context "with errors rendered"
```
slorek: As above.

PR #1474, opened 2019-11-15
Bugfix: remove pH range if answer changed to not require it

frankieroberto:
This fixes the following bug:

If you select "The minimum pH is lower than 3" for a component, and then give an explicit range (eg 2.3 to 3.0), but then go back and change the answer to the first question to be "It does not have a pH", the 2.3–3.0 pH range was not removed from the component.

This wasn't really visible unless you changed your answer _again_ to require a pH range, in which case your previous answer was still there.

Spotted by @nasirkhanbeis as part of #1471. 👍

### DIFF
--- a/cosmetics-web/app/models/component.rb
+++ b/cosmetics-web/app/models/component.rb
@@ -138,6 +138,16 @@ class Component < ApplicationRecord
     super(reject_non_decimal_strings(value))
   end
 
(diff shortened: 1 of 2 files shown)


nasirkhanbeis:
@frankieroberto  Just had quick look, back button is now returning straight to dashboard instead  pH question page.

PR #1476, opened 2019-11-18
Single review app with multiple processes

frankieroberto:
This changes the review app deploy process on CloudFoundry so that rather than deploying both a "web" and a "worker" app, a single app is deployed with "web" and "worker" processes.

This simplifies and speeds up the deploy process, as the app only needs to be pushed once, not twice.

Once we are happy with how this works, we can do the same for our staging and production environments.

Review app: https://cosmetics-pr-1476-submit-web.london.cloudapps.digital/


slorek:
We should also test that the worker is working on the review app.

frankieroberto:
@slorek I've verified that the worker is working (by scaling it down to 0, seeing that no CPNP files get processed, scaling it back up to 1 and seeing the files in the queue get processed).

PR #1478, opened 2019-11-18
Add nanomaterial notifications

frankieroberto:
This adds a new process for notifying OPSS about non-standard usage of Nanomaterials. Most of the details are submitted via upload a PDF, with the exception of the name and the details of if and when it was previously submitted to the EU via CPNP.

### 1. Nanomaterials tab

nanomaterials-1

### 2. Nanomaterial name

nanomaterials-2

### 3. Was the EU previously notified about it?

nanomaterials-4

### 4. Upload PDF

nanomaterials-5

### 5. Check your answers

nanomaterials-6

### 6. Confirmation

nanomaterails-7


tobyhughes:
Notice a bug when using "<a nano>" as the nanomaterial name

image

tobyhughes:
If I press 'back' on the 'Was the EU notified about blano on CPNP before 1 February 2020?' page, when I press continue on the 'What is the name of the nanomaterial?' page, then I get taken to the 'check your answers' page. Instead I should continue through the journey.

Instead, I should continue through the journey back to the date page.

tobyhughes:
Need to restrict file type to PDF

tobyhughes:
Display name of file here? 
image

tobyhughes:
Need "What happens next" content on this page (conditional on answer to 'Was the EU notified...' page)

image

tobyhughes:
Need some content for here 
image @PabCab

frankieroberto:
@tobyhughes thanks, all good bugs / things not implemented yet!

frankieroberto:
> Noticed a bug when using " < a nano > " as the nanomaterial name

Fixed, thanks.

frankieroberto:
> Display name of file here?

Good idea, done.

frankieroberto:
> Need "What happens next" content on this page (conditional on answer to 'Was the EU notified...' page)

This is done.

frankieroberto:
@tobyhughes @ralph-hawkins @PabCab think I've addressed all the issues - but let me know if you spot any more.

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
```diff
@@ -0,0 +1,141 @@
...
+    @form_url = name_nanomaterial_path(@nanomaterial_notification)
+    @form_method = :patch
+  end
+
+  def update_name
+    @nanomaterial_notification.name = params[:nanomaterial_notification][:name]
+
+    if @nanomaterial_notification.save(context: :add_name)
+
+      if @nanomaterial_notification.submittable?
```
slawosz: I would rather extract redirection logic to some separate controller. There are several reasons for this:
* easier testing, as you don't need to run entire controller upfront
* easier code to follow - you have a redirection in one place
* easier debugging

frankieroberto:
To-do:

[ ] Remove tab bar
[ ] Add back button from first question
[ ] Make sure Brexit date doesn't wrap
[ ] Make "Date EU notified on" bold

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
frankieroberto: I'm not convinced by this, as it would introduce a double-redirect. We’d also still need the condition on whether the model was valid or not, so that we can re-render the page with errors.

I think there could be another way of refactoring the condition on whether to redirect to the check-your-answers page or the next question though, so I'll give some thought to that.

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+    @back_link_path = back_link_path_for(:name)
+    @form_url = responsible_person_nanomaterials_path(@responsible_person)
+    @form_method = :post
+    render "name"
+  end
+
+  def create
+    @nanomaterial_notification = @responsible_person.nanomaterial_notifications.new
+    @nanomaterial_notification.name = params[:nanomaterial_notification][:name]
+    @nanomaterial_notification.user_id = current_user.id
```
baphled: We should be able to set this up when we initialise the object 🤔

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
baphled: I think we're missing a service object, or something here. Although I'm not quite sure what it should look like yet.

More food for thought, than anything 🤔

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  before_action :set_nanomaterial_notification_from_url, only: %i[notified_to_eu update_notified_to_eu upload_file update_file review name update_name submit confirmation_page]
+
+  before_action :redirect_to_confirmation_page_if_submitted, only: %i[notified_to_eu update_notified_to_eu upload_file update_file review name update_name submit]
+
+  def index; end
+
+  def new
+    @nanomaterial_notification = @responsible_person.nanomaterial_notifications.new
+    @back_link_path = back_link_path_for(:name)
```
baphled: I think we used something like `@previous_page`, or similar to keep track of this previous. Ideally we should keep it consistent.

cosmetics-web/app/models/nanomaterial_notification.rb
```diff
@@ -0,0 +1,138 @@
...
+
+private
+
+  def eu_notification_date_must_be_pre_brexit
+    if notified_to_eu_on && Date === notified_to_eu_on && notified_to_eu_on > EU_EXIT_DATE
+      errors.add(:notified_to_eu_on, I18n.t(:post_brexit_date_given, scope: %i[activerecord errors models nanomaterial_notification attributes notified_to_eu_on]))
+    end
+  end
+
+  def eu_notification_date_is_real
```
baphled: This feels like a custom validator 🤔

cosmetics-web/app/views/nanomaterial_notifications/notified_to_eu.html.erb
```diff
@@ -0,0 +1,87 @@
...
+
+      <% if @nanomaterial_notification.errors[:notified_to_eu_on].size > 0 %>
+        <% errorMessage = {
+          text: @nanomaterial_notification.errors.full_messages_for(:notified_to_eu_on).first.presence
+        }  %>
+
+        <% error_class = "govuk-input--error" %>
+      <% end %>
+
+      <% date_notified_to_eu_html = render("components/govuk_date_input",
```
baphled: It's starting to feel like we're in need of a presenter or decorator for things like this 🤔

cosmetics-web/config/initializers/shared_rails_config.rb
```diff
@@ -9,3 +9,9 @@
 ActionView::Base.include GovukDesignSystem::ComponentsHelper
 ActionView::Base.include GovukDesignSystem::DetailsHelper
 ActionView::Base.include GovukDesignSystem::WarningTextHelper
+ActionView::Base.include GovukDesignSystem::BackLinkHelper
```
baphled: Could we not dynamically load these helpers?

cosmetics-web/spec/request_helpers.rb
```diff
@@ -29,6 +29,69 @@ def have_back_link_to(html)
   end
 end
 
+# Adds a custom matcher asserting that a given HTML string contains
+# a GOV.UK 'Back link' component to the URL path given.
+module HaveH1WithTextMatcher
```
baphled: We could put these matchers in the support directory.

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
frankieroberto: Ah yeah, I think we’ve used `@previous_page_path` elsewhere, which I think is also more descriptive. Will change. 👍

cosmetics-web/config/initializers/shared_rails_config.rb
frankieroberto: Yeah I think we should just automatically load them all – will investigate how to do that as a separate PR (need to investigate a bit more about how Rails engines work).

cosmetics-web/spec/request_helpers.rb
frankieroberto: Good idea. See 6887d45adb00912bd45115d42412ba900901be60.

cosmetics-web/app/views/nanomaterial_notifications/notified_to_eu.html.erb
frankieroberto: These components are currently partials in the [govuk-design-system-rails](https://github.com/uKGovernmentBEIS/govuk-design-system-rails), but I'm slowly moving them to be Rails helpers instead.

cosmetics-web/app/models/nanomaterial_notification.rb
frankieroberto: I'm definitely tempted to extract it to be a more generic custom validator (`:date_is_real`?) – feels like it should be helpful wherever we ask for dates. Default Rails behaviour seems to be to throw an exception if you try and select an invalid date, which isn’t super helpful.

cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
frankieroberto: Good call. Fixed in ff744b0eaa4e6a70d4f8384e20c6b9f5f6af1f7d.

cosmetics-web/app/models/nanomaterial_notification.rb
frankieroberto: I think I'll save this for a future change though – whenever the need comes up to share this validator with a different date in another model.

PR #1480, opened 2019-11-21
Drop docker build/push/pull steps

frankieroberto:
Currently we push all builds to DockerHub, and tag all deployed builds. This takes a fair amount of time, and hasn’t proved to be that useful.


frankieroberto:
@slorek so I realised after pushing this commit that currently our tests pull ALL the docker images from `docker-compose.yml` from dockerhub in order to run the tests. 3e4f23a613fb1db44efebe5f39b5acfbac8b1f31 switches this to build them instead, which should work, but is way slower – because DockerHub is acting as a sort-of cache for the keycloak and antivirus docker images which rarely change.

We could switch the tests to use a shared long-running keycloak service, in the short term? Antivirus could be disabled for tests, as I don’t think it’s really needed?

frankieroberto:
Screenshot 2019-11-21 at 23 06 31

17 minutes to build all the docker images from scratch... at least the review app was deployed in parallel though.

slorek:
👍 We're working on running a Keycloak instance for tests separately.

PR #1482, opened 2019-11-22
Make sure users can add additional information when nano material requires more information.

baphled:
## Description
When a notification still requires nano material information, before completing the notification, we were not checking whether this was the case on the Notification Dashboard.

This PR remedies this by refactoring notification's to consider whether there's missing information and allowing the user to add any missing information from the dashboard.

## Checklist:

- [x] Test without javascript
- [x] Updating appropriate views from Slim to Erb


cosmetics-web/app/controllers/additional_information_controller.rb
```diff
@@ -2,9 +2,9 @@ class AdditionalInformationController < ApplicationController
   before_action :set_notification, only: %i[index]
 
   def index
-    if @notification.nano_material_incomplete?
-      component = @notification.components.find(&:nano_material_incomplete?)
-      nano_element = component.nano_material.nano_elements.find(&:incomplete?)
+    if @notification.nano_material_required?
```
slorek: I guess it’s already understood that this means we can only show the first nano material when there may actually be several.

cosmetics-web/app/models/component.rb
```diff
@@ -111,8 +111,8 @@ def display_root_category
     get_category_name(root_category)
   end
 
-  def nano_material_incomplete?
-    nano_material&.nano_elements_incomplete?
+  def nano_material_required?
+    nano_material&.nano_elements_required?
```
slorek: May return `nil` which would be unexpected for a `?` method.

cosmetics-web/app/models/nano_element.rb
```diff
@@ -29,22 +29,22 @@ def non_standard?
 
 private
 
-  def toxicology_incomplete?
+  def toxicology_required?
     confirm_toxicology_notified.nil? ||
```
slorek: It might be more efficient to return the previously calculated result of these methods from instance variables, since they’re used quite a lot. Assuming it doesn’t cause difficulty if the model is mutated.

cosmetics-web/spec/controllers/additional_information_controller_spec.rb
```diff
@@ -78,9 +78,9 @@
         create(:notification, components: [first_component, second_component], responsible_person: responsible_person)
       end
 
-      context "when all component's have incomplete nano materials" do
-        it "redirects to the 2nd component's next incomplete nano element" do
-          nano_element = second_component.nano_material.nano_elements.find(&:incomplete?)
+      context "when all component's have required nano materials" do
```
slorek: Remove apostrophe

cosmetics-web/spec/features/responsible_persons/notifications_dashboard_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  it "is able to view my Notification Dashboard" do
+    visit responsible_person_notifications_path(responsible_person)
+
+    expect(body).to have_css("#main-content")
+  end
+
+  it "has a errors tab" do
+    visit responsible_person_notifications_path(responsible_person)
+
+    expect(body).to have_css(".govuk-tabs section#incomplete")
```
slorek: The element IDs appear to be mixed up between this and the following test.

cosmetics-web/spec/features/responsible_persons/notifications_dashboard_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  let(:responsible_person) { create(:responsible_person_with_user) }
+
+  before do
+    sign_in_as_member_of_responsible_person(responsible_person)
+  end
+
+  it "is able to view my Notification Dashboard" do
+    visit responsible_person_notifications_path(responsible_person)
+
+    expect(body).to have_css("#main-content")
```
slorek: 🤔is there something more specific we can check for here?

baphled: Possibly, it was just to get the feature up and running tbf 🤔

cosmetics-web/spec/features/responsible_persons/notifications_dashboard_spec.rb
baphled: Good catch

cosmetics-web/app/models/component.rb
baphled: Agreed, although this was a refactoring of the message, the existing behaviour hasn't changed.

cosmetics-web/app/controllers/additional_information_controller.rb
baphled: I'm sure there's a spec for that 🤔 :)

cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -38,4 +38,46 @@
...
+  describe "#missing_information?" do
+    let(:notification) { build(:notification) }
+
+    before do
+      allow(notification).to receive(:nano_material_required?).and_return(true)
+      allow(notification).to receive(:formulation_required?).and_return(true)
+      allow(notification).to receive(:images_required?).and_return(true)
+    end
+
+    it "has no missing information" do
```
slorek: Context blocks (where the various flags are set true/false with the corresponding descriptions) would make the tests a lot more readable for me.

For example:

```ruby
context "when nano material is complete" do
  before { allow(notification).to receive(:nano_material_required?).and_return(false) }
  it "returns true" do
    expect(notification).to be_missing_information
  end
end
```

cosmetics-web/app/models/component.rb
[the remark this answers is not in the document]
baphled: This has been addressed in a separate PR, as it wasn't part of this changed here.

https://github.com/UKGovernmentBEIS/beis-opss/pull/1484

PR #1484, opened 2019-12-02
Make sure Component#nano_material_required? acts as expected

baphled:
## Description

We expect a response of `true` or `false` when making a call to `Component#nano_material_required` previously this would come back as `nil` if the component didn't have a nano material present.

This resolves this, along with adding a few specs to describe the behaviour.

Review app: https://cosmetics-pr-1484-submit-web.london.cloudapps.digital/
## Checklist:


cosmetics-web/spec/models/component_spec.rb
```diff
@@ -373,4 +373,34 @@
       end
     end
   end
+
+  describe "#nano_material_required?" do
+    context "when there is no nano material" do
+      let(:component) { build(:component) }
+
+      it "does not require nano material information" do
+        expect(component).not_to be_nano_material_required
```
frankieroberto: This might be slightly clearer as:

```suggestion
        expect(component.nano_material_required?).to be false
```

cosmetics-web/spec/models/component_spec.rb
```diff
@@ -373,4 +373,34 @@
       end
     end
   end
+
+  describe "#nano_material_required?" do
+    context "when there is no nano material" do
```
frankieroberto: minor typo:
```suggestion
    context "when there is no nanomaterial" do
```

cosmetics-web/app/models/component.rb
```diff
@@ -112,6 +112,8 @@ def display_root_category
   end
 
   def nano_material_required?
+    return false if nano_material.nil?
```
slorek: Not sure the naming of this method makes sense, as it seems it can be `true` if `nano_material` is already supplied. Maybe `nano_elements_required?` would be better. Not directly related to this change but since we're refactoring...

Also, could be more succinctly represented as:

```suggestion
    return nano_material && nano_material.nano_elements_required?
```

frankieroberto: I guess it’s really about validity, rather than whether they’re "required" or not?

Rename as `nano_material_valid?`? (And then we can refactor into an actual Rails validation at a later date).

baphled: @slorek it's name isn't great, as this functionality was dealt with reactively. I've mentioned this needs to be ironed out by the ground but that's not happened yet.

I suggest there's a conversation about what a complete/incomplete notification looks like, then refactor these as appropriate.

At the moment nano elements are required if there's a nano material that does not have nano element information, so it seems to make sense in this context.

baphled: > I guess it’s really about validity, rather than whether they’re "required" or not?
> 
> Rename as `nano_material_valid?`? (And then we can refactor into an actual Rails validation at a later date).

`*_valid` would be confusing. This isn't validation, so I don't think that's the appropriate course of action.

cosmetics-web/spec/models/component_spec.rb
baphled: I think this is telling us the unary name isn't great. I'd look to improve that, rather than change a unary expectation to a truthy/falsey one 🤷‍♂

frankieroberto: @baphled `be false` tests by doing `== false`, according to [the docs](https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers). The falsey matcher is `be_falsey`.

But you're right that `nano_material_required?` is a bad name... 😩

cosmetics-web/app/models/component.rb
```diff
@@ -112,6 +112,8 @@ def display_root_category
   end
 
   def nano_material_required?
+    return nano_material && nano_material.nano_elements_required?
+
     nano_material&.nano_elements_required?
```
frankieroberto: This line can be removed now.

baphled: Yup, I haven't updated this yet :)

cosmetics-web/spec/models/component_spec.rb
baphled: Same principle, tbf. It still reads worse imho.

PR #1485, opened 2019-12-02
Make sure users can only submit a notification when there is no more information to add.

baphled:
## Description
Resolves issue where users were allowed to notify a product whilst there was still missing information.

Review app: https://cosmetics-pr-1485-submit-web.london.cloudapps.digital/
## Checklist:

### DIFF
--- a/cosmetics-web/app/models/notification.rb
+++ b/cosmetics-web/app/models/notification.rb
@@ -99,7 +99,7 @@ class Notification < ApplicationRecord
       transitions from: :draft_complete, to: :notification_complete,
                   after: Proc.new { __elasticsearch__.index_document } do
         guard do
-          notified_pre_eu_exit? || images_are_present_and_safe?
+          (notified_pre_eu_exit? || images_are_present_and_safe?) && !missing_information?
         end
       end
     end
--- a/cosmetics-web/spec/factories/image_upload.rb
+++ b/cosmetics-web/spec/factories/image_upload.rb
@@ -1,9 +1,9 @@
 FactoryBot.define do
   factory :image_upload do
+    file { Rack::Test::UploadedFile.new("spec/fixtures/testPdf.pdf", "image/png") }
+
     trait :uploaded_and_virus_scanned do
       after :create do |image_upload|
-        file = fixture_file_upload(Rails.root.join("spec", "fixtures", "testImage.png"), "image/png")
-        image_upload.file.attach(file)
         image_upload.file.metadata["safe"] = "true"
       end
     end
--- a/cosmetics-web/spec/factories/image_upload.rb
+++ b/cosmetics-web/spec/factories/image_upload.rb
@@ -1,0 +1,11 @@
+FactoryBot.define do
+  factory :image_upload do
+    trait :uploaded_and_virus_scanned do
+      after :create do |image_upload|
+        file = fixture_file_upload(Rails.root.join("spec", "fixtures", "testImage.png"), "image/png")
+        image_upload.file.attach(file)
+        image_upload.file.metadata["safe"] = "true"
+      end
+    end
+  end
+end
--- a/cosmetics-web/spec/models/notification_spec.rb
+++ b/cosmetics-web/spec/models/notification_spec.rb
@@ -91,7 +91,7 @@ RSpec.describe Notification, type: :model do
 
       context "when notified pre EU exit" do
         before do
-          allow(notification).to receive(:notified_pre_eu_exit?).and_return(true)
+          notification.was_notified_before_eu_exit = true
         end
 
         it "can submit a notification" do
@@ -100,8 +100,14 @@ RSpec.describe Notification, type: :model do
       end
 
       context "when images are present and safe" do
+        let(:image_upload) { ImageUpload.new }
+        let(:file) { fixture_file_upload("/testImage.png", "image/png") }
+
         before do
-          allow(notification).to receive(:images_are_present_and_safe?).and_return(true)
+          image_upload.file.attach(file)
+          image_upload.file.metadata["safe"] = "true"
+
+          notification.image_uploads = [image_upload]
         end
 
         it "can submit a notification" do
@@ -117,7 +123,7 @@ RSpec.describe Notification, type: :model do
 
       context "when notified pre EU exit" do
         before do
-          allow(notification).to receive(:notified_pre_eu_exit?).and_return(true)
+          notification.was_notified_before_eu_exit = true
         end
 
         it "can not submit a notification" do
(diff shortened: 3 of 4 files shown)


cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -80,4 +80,60 @@
...
+    context "when no missing information" do
+      before do
+        allow(notification).to receive(:missing_information?).and_return(false)
+      end
+
+      context "when notified pre EU exit" do
+        before do
+          allow(notification).to receive(:notified_pre_eu_exit?).and_return(true)
+        end
+
```
frankieroberto: I think it would be better to setup these contexts by setting the relevant attributes on the model rather than mocking a different method on the same model. Otherwise the test is tightly tied to the implementation, which van make refactoring more difficult.

For example, there’s already a trait we could use here:
```ruby
let(:notification) { build(:notification, :pre_brexit) }
```

cosmetics-web/app/models/notification.rb
```diff
@@ -99,7 +99,7 @@ def as_indexed_json(*)
       transitions from: :draft_complete, to: :notification_complete,
                   after: Proc.new { __elasticsearch__.index_document } do
         guard do
-          notified_pre_eu_exit? || images_are_present_and_safe?
+          (notified_pre_eu_exit? || images_are_present_and_safe?) && !missing_information?
```
frankieroberto: Given that `missing_information?` also checks whether `images_required?`, could we refactor this to just:


```ruby
guard { !missing_information? }
```

cosmetics-web/spec/models/notification_spec.rb
slorek: I agree, we should set the state of the object rather than mocking the object under test. Mocking is more useful for other dependencies.

cosmetics-web/app/models/notification.rb
baphled: I think those checks are slightly different.

`missing_information?` checks whether more information is required, where `images_are_present_and_safe?` is check whether the images are present and checked by AV.

I think `notified_pre_eu_exit?` doesn't really live in `missing_information?` either but I could be wrong about that. I have seen any requirements for it yet 🤔

cosmetics-web/spec/models/notification_spec.rb
baphled: I can see your point, although we'll start to get into the realms of testing by indirection and tying our tests to factories instead.

I can see an argument for a `draft_notification` trait, but changing it to the above would lose the intent of the specs.

As we're describing a draft notification, it makes sense to show that in the setup.

We don't care about the internals of the `Notification`, nor what we need to do to get `images_are_present_and_safe?` or `notified_pre_eu_exit?` to return `true` or `false`, as that should be done in other specs. So we mock the responses to control the conditions and make it explicit what conditions are required when we verify our expectations.

This approach has helped to guide the refactoring process, and keep the specs descriptive enough so that they're self documenting.

frankieroberto: You don’t _have_ to use factories or traits - for some tests, `Notification.new(foo: "bar")` within the setup might be more explicit.

I still don’t think we should mock the other methods of the object being tested though. The test for `#may_submit_notification?` doesn’t need to know how this is implemented within the model. If we refactor this to no longer call `images_are_present_and_safe?` but some other internal method, then the tests should still pass.

cosmetics-web/app/models/notification.rb
frankieroberto: You’re probably right, but I'm a bit confused as this calls `images_are_present_and_safe?` which is defined as:

```ruby
  def images_are_present_and_safe?
    !image_uploads.empty? && image_uploads.all?(&:marked_as_safe?)
  end
```

and then calls `!missing_information` which is defined as:

```ruby
  def missing_information?
    nano_material_required? || formulation_required? || images_required?
  end
```

which is defined as:

```ruby
  def images_required?
    notified_post_eu_exit? && image_uploads.empty?
  end
```

...so `images_required?` seems to be checking whether the the notification is Post Brexit and has no images, whereas `(notified_pre_eu_exit? || images_are_present_and_safe?) ` seems to be checking that the notification is either pre-Brexit or that images are present – ie the same logic, but negated?

baphled: Could that be that `images_required?` is missing logic ie.  Checking for `notified_pre_eu_exit?` or something like that.

I have to admit I'm not fully aware of what other conditions `images_required?` may have when the notification is pre brexit 🤔 

If that's the case, then we're likely missing a scenario, or haven't written tests around it yet.

cosmetics-web/spec/models/notification_spec.rb
baphled: There's merit in both approaches tbf but these types of distinctions require a bit more than a few comments in a PR.

As there seems to be consensus for factories over mocks now, I'll refactor accordingly.

cosmetics-web/app/models/notification.rb
baphled: Okay, so I've had a look at this a bit more.

So I think `notified_pre_eu_exit? || images_are_present_and_safe?` is in the wrong place and should likely be part of `images_required?`. I don't think there's a test around this, although it's worth raising and consolidating this asap.

The assumption being, an image is only required if the following is met:

1. is notified before EU exit and there are no uploaded images
2. is notified after EU exit and has uploaded images that have been AV checked.

I'd also suggest we refactor any code that uses the old way of doing things so the code base starts to get a bit more coherent.

baphled: I've created a PR for this: https://github.com/UKGovernmentBEIS/beis-opss/pull/1490

I'll finish it off, once this has been merged.

cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -80,4 +80,48 @@
...
+        it "can submit a notification" do
+          expect(notification).to be_may_submit_notification
+        end
+      end
+    end
+
+    context "when information is missing" do
+      before do
+        allow(notification).to receive(:missing_information?).and_return(true)
```
baphled: I've replaced most of the mocks with factories. There's still this one, which I've spent a bit of time on but ended up leaving for the moment as it takes quite a bit of setup.

cosmetics-web/spec/factories/image_upload.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :image_upload do
+    trait :uploaded_and_virus_scanned do
+      after :create do |image_upload|
+        file = fixture_file_upload(Rails.root.join("spec", "fixtures", "testImage.png"), "image/png")
```
frankieroberto: I found a slightly simpler way of doing this, which doesn't require an `after :create` hook: https://github.com/UKGovernmentBEIS/beis-opss/pull/1478/files#diff-63d5e43f790a367ef6aea7e24f443eabR8

cosmetics-web/spec/rails_helper.rb
```diff
@@ -69,8 +69,11 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 
-  config.include FactoryBot::Syntax::Methods
+  FactoryBot::SyntaxRunner.class_eval do
+    include ActionDispatch::TestProcess
+  end
```
frankieroberto: What is this needed for?

baphled: Good catch, think that's left over from when I was setting up the `ImageUpload` factory 🤔 🤷‍♂

PR #1492, opened 2019-12-09
Fix deploys by restoring docker builds of the app

frankieroberto:
(Short-term fix whilst we work on removing the docker hub push / pull step)


infrastructure/ci/build-docker-images.sh
```diff
@@ -7,7 +7,9 @@ COMPONENTS=(
     'antivirus'
     'maintenance'
     'db'
-    'keycloak'
+    'keycloak',
```
slorek: Remove commas.

PR #1493, opened 2019-12-09
Converting another 5 templates from Slim to ERb

frankieroberto:


cosmetics-web/app/views/responsible_persons/add_notification_wizard/register_on_eu_system.html.erb
```diff
@@ -0,0 +1,30 @@
+<% title = "If you’re likely to notify the EU about the products" %>
```
baphled: Not sure why we're not setting these as instance variables in the controller 🤔

frankieroberto: I think ultimately I’d prefer to be specifying page titles in the localization files. Doesn’t feel like a concern of the controller to me. There’s no logic involved, it’s only a variable as it’s being used in two places (the `<title>` tag and the `<h1>`).

For these Slim to ERb conversions though, I think we should avoid refactoring, and tackle that as future tasks?

baphled: Sure, but this still applies to the other local variables.

PR #1511, opened 2019-12-16
Update behaviour for confirming nano materials

frankieroberto:
Fixes https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-763


nasirkhanbeis:
Had a quick look, seems for notifications with nano only works fine however couple regressions on below scenarios
1.Image should not be mandatory for pre-brexit notifications
2. If notification needs formulation upload then after nano I should be taken to formulation upload page and then check your answer page

nasirkhanbeis:
looking good now, images and the flow with nano only working fine. However there seems a regression issue which is already in prod so will raise another ticket for this.

PR #1512, opened 2019-12-16
Remove db component

frankieroberto:
We never deploy this (as we're using databases provided as a service, rather than building our own), so we don't need to build the docker images.


slorek:
I think this is till required to support running Keycloak in docker-compose context. We should move this out into a separate repo which would solve the main issues with this.

PR #1514, opened 2019-12-18
Fix multicomponent upload

slawosz:
It fixes error on multicomponent upload. Original error was caused by failing to create duplicated component, as at the very beginning of the process, the component is created without notification.
It was fixed by calling new  in component creation in `CpnpParser` instead of `create!`. Also adds validation and db constraints to require `notification_id`. 

Other part of this PR is refactoring. Before, everything related to `Notification` import process was done inside `ReadDataAnalyzer`. Tests were flaky and it wasn't best place for this code.
This PR introduces following refactorings, by spliting `ReadFileAnalyzer` into 2 classes:

### New service class `CpnpNotificationImporter`

It gets the instance of `CnpnParser`, which stores result of parsing cpnp zip file and creates notification.

### New job `NotificationFileProcessorJob`

This job encapsulates the entire process of file parsing.


cosmetics-web/app/models/component.rb
```diff
@@ -37,6 +37,7 @@ class Component < ApplicationRecord
   # a notification when the user indicates that it is a kit/multi-component,
   # so the uniquness validation has to allow non-unique null values.
   validates :name, uniqueness: { scope: :notification_id, allow_nil: true, case_sensitive: false }
+  validates :notification, presence: true
```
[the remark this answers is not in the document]
slawosz: For consistency, it is better to have this validation in place I would say.

cosmetics-web/spec/jobs/notification_file_processor_job_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.describe NotificationFileProcessorJob do
+  let!(:notification_file_basic) { create(:notification_file, uploaded_file: create_file_blob) }
+
+  after do
+    sign_out
```
slorek: I'm not sure we have called `sign_in` so this shouldn't be necessary.

cosmetics-web/spec/jobs/notification_file_processor_job_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      described_class.new.perform(notification_file_basic.id)
+      notification = Notification.order(created_at: :asc).last
+
+      expect(notification.product_name).equal?("CTPA moisture conditioner")
+    end
+
+    context "when the file contains a post-Brexit date" do
+      let(:notification_file) { create(:notification_file, uploaded_file: create_file_blob("testExportFilePostBrexit.zip")) }
+
+      before do
```
slorek: Refactor suggestion: Move this `before` block to the top level and rename the first `let` to `notification_file` and remove the redundant `perform` calls in the previous specs.

cosmetics-web/app/models/component.rb
slorek: I think this is probably not necessary because Rails now automatically validates presence of `belongs_to` associations unless overridden.

cosmetics-web/app/jobs/notification_file_processor_job.rb
```diff
@@ -0,0 +1,110 @@
...
+      static_file?(file) && file_contents_differs?(file)
+    end
+  end
+
+  def file_is_product_xml?(file)
+    file.name&.match?(product_xml_file_name_regex)
+  end
+
+  def file_is_valid?(file)
+    file.file? && file.name !~ /__MACOSX/ && file.name !~ /\.DS_Store/
```
slorek: 👀

PR #1517, opened 2019-12-19
Adding feature specs for notifications

frankieroberto:
This adds rspec feature specs to test the end-to-end journey of notifying cosmetic products via either uploading a ZIP file or "manually" by answering questions, in both pre and post Brexit scenarios, and with different variants of notification.

As ZIP files and product images are processed by jobs in queue, to make these flows more easily testable, the job queuing process within the test environment is changed from `:async` (using a background thread) to `:inline`. As a consequence of this, some of the other background tasks (such as scanning for viruses or sending e-mails) are stubbed to avoid throwing exceptions.

Once this branch has been merged, the intention is to convert and remove the remaining "system" tests (which also test end-to-end, but use a full Chrome browser using Selenium).


nasirkhanbeis:
looks good to me, just need to add further end to end test flow against each zip file
Also I want to add more scenarios

cosmetics-web/.rubocop.yml
```diff
@@ -39,6 +39,12 @@ ExampleLength:
 Capybara/FeatureMethods:
   Enabled: false
 
+# We can disable this as we’re using Capybara with RackTest which
+# doesn’t support javascript, and so we we don’t need to use the
+# asynchronous `have_current_path` matcher.
+Capybara/CurrentPathExpectation:
+  Enabled: false
```
slorek: I've used `have_current_path` on PSD feature specs; is it not analogous to `page.current_path`?

frankieroberto: @slorek it is, but I’d already written dozens of `expect(page.current_path).to end_with("/foo/bar")`.  Could be rewritten as `expect(page).to have_current_path(/\foo\/bar\z/)` but that felt less readable (and I didn't fancy escaping all those forward slashes).

Probably the ideal is to create a `have_current_path_ending_with` expectation, but given that the intent of the rubocop is to work with async javascript drivers (which we’re not using here), it felt ok to disable it for now.

I did consider whether the feature specs should assert on the full URL paths (eg `\/responsible_persons\/\d+\/notifications\/\d+\/components\/\d+\/build\/ph`) but we have such long URLs that it's a bit of a pain, and I figured that just looking at the last part was probably good enough for asserting the flows through the app... (and request specs could assert the full paths).

Thoughts welcome though!

slorek: 👍 It's a pain that the Wicked Wizard gem doesn't create route helpers for these pages! I guess we don't need to explicitly disable this helper, though?

frankieroberto: Route helpers do get generated, although the 'step' is the `:id (eg `responsible_person_notification_component_build_path(responsible_person_id, notification_id, component_id, :ph)` – but they're also super long, and would require the feature specs to somehow figure out the IDs of all the models. I also _kinda_ feel that it's cheating if the BDD/feature tests use methods exposed by the implementation.

> I guess we don't need to explicitly disable this helper, though?
Not quite sure I follow this...?

cosmetics-web/spec/features/manual_post_brexit_notifications_spec.rb
```diff
@@ -0,0 +1,373 @@
...
+    expect_to_see_message "SkinSoft tangerine shampoo notification submitted"
+  end
+
+  scenario "Manual, post-Brexit, ingredient ranges, single item, no nanomaterials", :with_stubbed_antivirus do
+    visit new_responsible_person_add_notification_path(responsible_person)
+
+    expect_to_be_on_was_eu_notified_about_products_page
+    answer_was_eu_notified_with "No"
+
+    expect_to_be_on_are_you_likely_to_notify_eu_page
```
slawosz: how about `expect_to_be_on__are_you_likely_to_notify_eu_page`, so introducing double `_` for better readability?

frankieroberto: Good idea - done!

PR #1522, opened 2020-01-06
Refactor ReadDataAnalyzer

frankieroberto:
This refactors the class which parses the XML files (within ZIP files) attached to `NotificationFile` models, and create `Notification` models instead.

The main motivation behind this is the discovery that within the existing `ReadDataAnalyzer` specs, the `ReadDataAnalyzer` class is being called automatically during the setup of the uploaded files, rather than (or sometimes as well as) explicitly within the `analyzer_instance.metadata` call. Whilst this had little effect when running the the jobs using the `:async` method (a separate thread), changing this to `:inline` caused all the tests to fail.

Whilst the XML parsing & importing can be thought of as a type of "analyzer", it doesn't feel like a good fit for `ActiveStorage::Analyzer`, as that is designed purely for adding metadata to existing files, rather than fundamentally processing them into records and then deleting the file.

## Before

`ReadDataAnalyzer` is a subclass of `ActiveStorage::Analyzer` and is run within `ActiveStorage::AnalyzeJob` immediately after every file is uploaded. Following the ActiveStorae::Analyzer` API, the `accept?` class method is first called to see whether or not to apply the analyzer, and if so, the `metadata` instance method is called.

The `metadata` method returns an empty hash, because that is what the ActiveStorage job is expecting to add to the metadata column in the database - but instead were using this method to analyze the xml file, create a notification record, and then delete the notification_file itself.

## After

There is a new `NotificationFileProcessorJob` which is queued (or run inline) with the `id` of the `NotificationFile`. Following the ActiveJob API, when the `perform` method is called, all of the same code is run as before – the attached file is downloaded, unzipped, and then a `Notification` record is saved, and the `NotificationFile` record is deleted.

When uploading the zip files, the `NotificationFilesController` now explicitly queues the job to process the files, via `NotificationFileProcessorJob.perform_later(notification_file.id)`

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
@@ -36,6 +36,8 @@ class ResponsiblePersons::NotificationFilesController < ApplicationController
         })
         return render :new
       end
+
+      NotificationFileProcessorJob.perform_later(notification_file.id)
     end
     redirect_to responsible_person_notifications_path(@responsible_person)
   end
--- a/cosmetics-web/app/jobs/notification_file_processor_job.rb
+++ b/cosmetics-web/app/jobs/notification_file_processor_job.rb
@@ -1,29 +1,19 @@
 require "zip"
 
-class ReadDataAnalyzer < ActiveStorage::Analyzer
-  extend AnalyzerHelper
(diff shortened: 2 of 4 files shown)


slawosz:
I will port this to my PR.

slawosz:
Its in https://github.com/UKGovernmentBEIS/beis-opss/pull/1514

PR #1523, opened 2020-01-09
Default to adding a validation on belongs_to relationships

frankieroberto:
This adds the `belongs_to_required_by_default = true` config option, which is the default for new Rails applications. This adds an automatic validation to all `belongs_to` relationships which presents the foreign key from being null.

By adding this, we can remove the explicit `validates :notification, presence: true` on the Component model.


cosmetics-web/spec/factories/responsible_person.rb
```diff
@@ -5,10 +5,9 @@
     address_line_1 { "Street address" }
     city { "City" }
     postal_code { "AB12 3CD" }
-    contact_persons { [create(:contact_person)] }
 
-    trait :with_no_contact_person do
-      contact_persons { [] }
+    trait :with_a_contact_person do
+      contact_persons { [create(:contact_person)] }
```
slorek: I'd argue this should probably be default behaviour if it's validated. If tests don't require a contact person they can set the property to `nil` easily.

frankieroberto: It's not validated - there's just some controller logic which redirects you to a different page if you've not yet added a contact person.

The main reason I switched from `with_no_contact_person` to `with_a_contact_person` is because there is now a validation on the `responsible_person` association on the `ContactPerson`, and so it was throwing errors & felt odd to have both factories create the other one. Seemed more sensible to only have the factories create the associations on the `belongs_to` association not the `has_many`.  The downside though is needing to use the `:with_a_contact_person` trait in most of the tests.

PR #1573, opened 2020-01-15
Dont perform component name validation on cpnp import

slawosz:
To prevent possible CPNP components validation issue, don't perform validation on name in case of zip file import.


cosmetics-web/app/services/cpnp_notification_importer.rb
```diff
@@ -15,6 +15,7 @@ def create!
     elsif @cpnp_parser.cpnp_notification_date >= EU_EXIT_DATE
       raise CpnpFileNotifiedPostBrexitError, "Product was notified to CPNP post-Brexit, which is not currently supported"
     else
+#      binding.pry
```
frankieroberto: remove commented out line

cosmetics-web/app/services/cpnp_parser.rb
```diff
@@ -63,7 +63,7 @@ def shades
 
   def components
     current_version_component_lists_node.xpath(".//component").collect do |component_node|
-      Component.new(name: component_name(component_node),
+      c = Component.new(name: component_name(component_node),
```
frankieroberto: rename variable `c` to `component` (to avoid single letter variable names)?

slawosz: Well, in this case, `c` is absolutely fine :) but will remove it anyway.

PR #1576, opened 2020-01-15
Bugfix: looping through nanomaterials

nasirkhanbeis:
Fix for a bug whereby in some conditions, not all nanomaterials are looped through and asked questions about before ending up on the Check your answers page.

See https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-788


frankieroberto:
@nasirkhanbeis think I've finally managed to understand and fix this bug. I had to update some of your test expectations (as they were in the wrong order now that we're looping through components and nanomaterials in strict ID order).

The fix isn't pretty - basically adding a lot more conditions into the `finish_wizard_path` method of the `nanomaterial_build_controller` - and should definitely all be refactored at some point.

nasirkhanbeis:
Tested this in review app, looks good as expected!

PR #1587, opened 2020-01-16
Replace declaration system spec with a request spec

frankieroberto:
This remove the "system" test (which uses a full browser) for the user declaration, with a request spec (which simulates a browser by sending HTTP requests directly to the app).

The objective here is to speed up the test suite by removing the dependency on having a full browser.


cosmetics-web/spec/requests/declaration_requests_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    let(:responsible_person) { create(:responsible_person) }
+    let(:user) { build(:user, first_login: true) }
+
+    before do
+      sign_in_as_member_of_responsible_person(responsible_person, user)
+    end
+
+    context "viewing the declaration page" do
+      before do
+        get "/declaration"
```
slorek: I think we should use URL helpers i.e. `declaration_path`

frankieroberto: Could do. I've tended to use full URLs on the other request specs so far, but happy to go with a consensus. I've had this debate with developers on other teams, and we concluded there’s pros and cons to both approaches.

## Full paths
Pros:
* also implicitly tests the routing – ie that the page is rendered from the exact url given
* _can_ be more readable, depending on how the route helpers are named.

Cons:
* Needs to be updated if the path changes

## Helpers
Pros:
* No changes needed if the path changes (so long as the helper doesn’t also get renamed)
* Can be more readable if lots of string interpolation would otherwise be needed.

Cons:
* Doesn’t provide test coverage of the URL helpers themselves (need to have a separate routing spec)

@slorek @slawosz thoughts? My mild preference goes to using full paths (mainly because I like seeing the actual URL path) – but I'm not too fussed.

slorek: I had been using full paths as well on PSD feature specs but we decided to go with helpers. I don't have a strong preference but it would be good to have a consistent approach.

slawosz: I prefer full paths so tests are more readable, but from testing perspective helpers make more sense.

frankieroberto: Maybe a good compromise is to have full paths on feature specs (which can also make sense if you’re writing the tests first before any implementation) but use helpers in request specs?

frankieroberto: We've resolved to do this.

PR #1588, opened 2020-01-16
Convert notification system specs to request specs

frankieroberto:
Removes the "resposible person dashboard" system specs (which require a full browser) and replaces these tests with request specs within `notification_page_spec` and `notification_files_spec`.


frankieroberto:
@nasirkhanbeis so it turns out that the code doesn’t actually check that the notification file had an error – you can delete _any_ notification file if you send the right request (but you only see a delete button in the browser for files with an error)...

PR #1598, opened 2020-01-17
convert "creating a responsible person" system tests to feature tests

nasirkhanbeis:


cosmetics-web/spec/support/feature_helpers.rb
```diff
@@ -529,3 +529,86 @@ def answer_does_product_contain_poisonous_ingredients_with(answer)
...
+    go_to_rp_declaration_page
+    select_individual_account_type
+
+    assert_text "UK Responsible Person details"
+
+    fill_in "Name", with: responsible_person.name
+    fill_in "Building and street", with: responsible_person.address_line_1
+    fill_in "Town or city", with: responsible_person.city
+    fill_in "County", with: responsible_person.county
+    fill_in "Postcode", with: responsible_person.postal_code
```
frankieroberto: I think it would be better to either hardcode these values, or pass them in as a parameters to the helper method, rather than expecting a `responsible_person` variable to be defined elsewhere.

cosmetics-web/spec/features/responsible_persons/create_rp_user_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+RSpec.describe "Resposible person creation journey", type: :feature do
+  let(:responsible_person) { create(:responsible_person_with_user, :with_a_contact_person) }
+
+  before do
+    sign_in as_user: create(:user, first_login: true)
+    stub_notify_mailer
+  end
+
+   scenario "Given as a new user, I should be able to create responsible user as a individual sole trader" do
```
frankieroberto: This reads a bit awkwardly. How about:

```suggestion
   scenario "creating a responsible user as a individual sole trader" do
```

slorek: Use context blocks around scenarios to make this DRY

cosmetics-web/spec/features/responsible_persons/create_rp_user_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe "Resposible person creation journey", type: :feature do
+  let(:responsible_person) { create(:responsible_person_with_user, :with_a_contact_person) }
+
+  before do
+    sign_in as_user: create(:user, first_login: true)
```
slorek: Use `let(:user) { create(:user, first_login: true) }` within a `context "as a new user"` block.

cosmetics-web/spec/support/feature_helpers.rb
```diff
@@ -529,3 +529,77 @@ def answer_does_product_contain_poisonous_ingredients_with(answer)
...
+    fill_in "Full name", with: business_responsible_person.contact_persons.first.name
+    fill_in "Email address", with: business_responsible_person.contact_persons.first.email_address
+    fill_in "Phone number", with: business_responsible_person.contact_persons.first.phone_number
+    click_on "Continue"
+  end
+
+
+  def select_business_account_type
+    assert_text "Is the UK Responsible Person a business or an individual?"
+    choose "responsible_person_account_type_business", visible: false
```
frankieroberto: I think ideally we should use the visible text rather than `name` attribute of the `<input>`, eg:

```suggestion
    choose "Limited company or Limited Liability Partnership (LLP)"
```

...also I think it should be visible?

cosmetics-web/spec/support/feature_helpers.rb
```diff
@@ -529,3 +529,77 @@ def answer_does_product_contain_poisonous_ingredients_with(answer)
...
+    assert_text "UK Responsible Person details"
+
+    fill_in "Name", with: responsible_person.name
+    fill_in "Building and street", with: responsible_person.address_line_1
+    fill_in "Town or city", with: responsible_person.city
+    fill_in "County", with: responsible_person.county
+    fill_in "Postcode", with: responsible_person.postal_code
+    click_on "Continue"
+
+    assert_text "contact person"
```
frankieroberto: can we make this a bit more specific?

PR #1607, opened 2020-01-20
Extract antivirus

slawosz:
Remove antivirus references. Antivirus is now in https://github.com/UKGovernmentBEIS/beis-opss-antivirus


frankieroberto:
@slawosz looks like you also need to remove [line 7 of build-docker-images.sh](https://github.com/UKGovernmentBEIS/beis-opss/blob/master/infrastructure/ci/build-docker-images.sh#L7)?

PR #1608, opened 2020-01-20
Update Keycloak SMS authenticator submodule - DO NOT MERGE

edwardhorsford:
Updates to latest Keycloak with [these updates](https://github.com/UKGovernmentBEIS/keycloak-sms-authenticator-sns/pull/126).

Note this changes the 2FA message to refer to the Product safety database rather than OPSS. It was OPSS so that these messages could work for both PSD and Cosmetics.

This change should be coordinated with turning off prepending the service name in Notify's settings.


slawosz:
As we are moving to https://github.com/UKGovernmentBEIS/beis-opss-keycloak I am closing this.

PR #1615, opened 2020-01-21
Use error summary helper

frankieroberto:
This updates the error summaries on all of the views to use a new `error_summary_for(model)` helper, which in turn uses a new `govukErrorSummary()` helper within the shared [`govuk-design-system-rails`](https://github.com/uKGovernmentBEIS/govuk-design-system-rails) gem.

The main difference is that this new helper follows the recommendations of the GOV.UK Design System in [linking from the error summary to each answer](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer). This means that when you click the link within the error summary, the relevant field is focused.


cosmetics-web/app/views/declaration/business_declaration.html.erb
```diff
@@ -1,9 +1,6 @@
 <% content_for :page_title, "Responsible Person Declaration" %>
 
 <%= form_with url: accept_declaration_path, local: true do |form| %>
-  <% if @errors.present? %>
```
slawosz: Why no helper here?

cosmetics-web/app/views/declaration/poison_centre_declaration.html.erb
```diff
@@ -1,9 +1,5 @@
 <% content_for :page_title, "Declaration" %>
 <%= form_with url: accept_declaration_path, local: true do |form| %>
-  <% if @errors.present? %>
```
slawosz: Why no helper here?

cosmetics-web/app/views/declaration/msa_user_declaration.html.erb
```diff
@@ -1,8 +1,5 @@
 <% content_for :page_title, "Declaration" %>
 <%= form_with url: accept_declaration_path, local: true do |form| %>
-  <% if @errors.present? %>
```
slawosz: Why no helper here?

cosmetics-web/app/views/declaration/business_declaration.html.erb
frankieroberto: I had a look at the controller, and we not longer ever set `@errors`, so this line isn't needed any more. I think it's a legacy from when you had to check a checkbox to accept the declaration, but now you just press the Continue button, so there are never any errors shown.

nasirkhanbeis:
Tested this in review app, looks good as expected!

PR #1618, opened 2020-01-21
add further create rp tests and improve test code

nasirkhanbeis:
Improve test coverage and  test code


cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe "Create multiple rp within same kc user", type: :feature do
```
frankieroberto: Can we try and avoid abbreviations like rp and kc in the test descriptions? (Just to make it less confusing for future developers to follow).

cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:responsible_person) { create(:responsible_person_with_user, :with_a_contact_person) }
+
+  before do
+    sign_in_as_member_of_responsible_person(responsible_person)
+    stub_notify_mailer
+  end
+
+  scenario "creating another respoonsible person as a limited company" do
```
frankieroberto: Typo:
```suggestion
  scenario "creating another responsible person as a limited company" do
```

cosmetics-web/spec/support/feature_helpers.rb
```diff
@@ -191,7 +191,6 @@ def expect_to_be_on__upload_formulation_document_page
   expect(page).to have_h1("Upload formulation document")
 end
 
-# rubocop:disable Naming/UncommunicativeMethodParamName
```
frankieroberto: I think we still need to disable this rubocop, as the keyword param `ph` is considered too short – but in this case is semantically meaningful.

cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    visit new_responsible_person_add_notification_path(responsible_person)
+    visit "/responsible_persons/account/overview"
+    expect(page).to have_h1("Are you or your organisation a UK Responsible Person?")
+    create_another_business_responsible_person
+  end
+
+  scenario "Verify dead end page when user choosen as account already exist" do
+    visit new_responsible_person_add_notification_path(responsible_person)
```
frankieroberto: Think we can drop this line as the next line immediately visits a different page?

cosmetics-web/spec/support/feature_helpers.rb
```diff
@@ -534,58 +532,57 @@ def add_an_item
 
 #Create responsible person helpers
 
-def go_to_rp_declaration_page
+def create_individual_responsible_person
+  go_to_rp_declaration_and_confirm
+  select_options_to_create_account
+  select_individual_account_type
+  expect(page).to have_h1("UK Responsible Person details")
```
frankieroberto: I think generally we've tried to have the interactions (select, fill_in, etc) and the expectations in separate helper methods – that way the scenario is a bit easier to follow.

cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
nasirkhanbeis: agree, addressed this now :)

cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
nasirkhanbeis: good spot, fixed!

cosmetics-web/spec/features/create_another_rp_user_within_same_kc_user_spec.rb
nasirkhanbeis: done

cosmetics-web/spec/support/feature_helpers.rb
nasirkhanbeis: yeah agree, I tried to put all page interactions in methods but yeah some tests may still have but I did that only coz I won't be using the same code anywhere, so to avoid one time usage of methods

PR #1646, opened 2020-01-28
Show deploy information for Staging and Production

scruti:
Github deployments API will track the deployment of both Staging and
Production environments when new commits are pushed into master branch.

When the deployments succeed, Github deploy information will show a
link for each of the deployed environments.

## Description
The process is the same in each environment deployed:

1. Create a Github deployment for the environment.
2. Set deployment status as initiated.
3. Deploy the environment.
4. Set deploy status as success or failure depending on the deploy result.
5. When successful, a link to the deployment environment will be displayed.

Related to https://github.com/UKGovernmentBEIS/beis-opss/pull/1637

## Alert for Reviewers

I don't recall a way to test this workflow without actually merging it into master and :crossed_fingers:. So if it doesn't work on the first attempt this can turn into a "fix/push/merge/deploy" nasty nightmare with the staging/production deploy...

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -58,7 +58,7 @@ jobs:
         LOG_URL: ${{ env.LOG_URL }}
       run: |
         source cosmetics-web/deploy-github-functions.sh
-        gh_deploy_success staging
+        gh_deploy_success staging https://staging-submit.cosmetic-product-notifications.service.gov.uk/
 
     - name: Update Staging deployment status (failure)
       if: failure()
@@ -109,7 +109,7 @@ jobs:
         LOG_URL: ${{ env.LOG_URL }}
       run: |
         source cosmetics-web/deploy-github-functions.sh
-        gh_deploy_success production
+        gh_deploy_success production https://submit.cosmetic-product-notifications.service.gov.uk/
 
     - name: Update Production deployment status (failure)
       if: failure()
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,44 +21,18 @@ jobs:
     - name: Create GitHub deployment for Staging
       env:
         GITHUB_TOKEN: ${{ secrets.GithubApiDeploymentToken }}
+        BRANCH: master
       run: |
-        DEPLOY_URL=$(curl -X POST \
-          -H "Authorization: token $GITHUB_TOKEN" \
-          https://api.github.com/repos/$GITHUB_REPOSITORY/deployments \
-          -d '{
-          "ref": "master",
-          "description": "Staging deploy",
-          "environment": "staging",
-          "auto_merge": false,
-          "required_contexts": []
-          }' | jq '.url?' | tr -d '"') # Gets 'url' field from the response and trims the surronding quotes.
-
-        if [ -z "$DEPLOY_URL" ]; then
-          echo "Failed to create Github deployment"
-        else
-          echo "::set-env name=DEPLOY_URL::$DEPLOY_URL"
-          echo "Github deployment created: $DEPLOY_URL"
-        fi
+        source cosmetics-web/deploy-github-functions.sh
+        gh_deploy_create staging
 
     - name: Initiate Staging deployment status
       env:
         GITHUB_TOKEN: ${{ secrets.GithubApiDeploymentToken }}
-        DEPLOY_STATUSES_URL: ${{ env.DEPLOY_URL }}/statuses
+        DEPLOY_STATUSES_URL: ${{ env.DEPLOY_STATUSES_URL }}
       run: |
-        ACTIONS_DEPLOY_URL=$(echo "https://github.com/$GITHUB_REPOSITORY/actions?query=branch%3Amaster+workflow%3ADeploy")
-        echo "::set-env name=ACTIONS_DEPLOY_URL::$ACTIONS_DEPLOY_URL" # Export for future steps
-
-        curl -X POST \
-          -H "Authorization: token $GITHUB_TOKEN" \
-          -H "Accept: application/vnd.github.ant-man-preview+json" \
-          -H "Accept: application/vnd.github.flash-preview+json" \
-          $DEPLOY_STATUSES_URL \
-          -d '{
-            "environment": "staging",
-            "state": "in_progress",
-            "description": "Staging deployment initiated",
-            "log_url": "'"$ACTIONS_DEPLOY_URL"'"
-          }'
+        source cosmetics-web/deploy-github-functions.sh
+        gh_deploy_initiate staging
 
     - name: Deploy to Staging
       env:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -32,6 +75,84 @@ jobs:
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GithubApiDeploymentToken }}
+      run: |
+        DEPLOY_URL=$(curl -X POST \
+          -H "Authorization: token $GITHUB_TOKEN" \
+          https://api.github.com/repos/UKGovernmentBEIS/beis-opss/deployments \
+          -d '{
+          "ref": "master",
+          "description": "Production deploy",
+          "environment": "production",
```
scruti: This can be set to "**prod**" to match the PaaS space, but there is already a "**production**" environment [listed in this project](https://github.com/UKGovernmentBEIS/beis-opss/deployments).

As Github does not seem to provide a way to delete deployment environments, if we create "**prod**" it will be listed together with "**production**". 

I personally get more confused by finding both "production" and "prod" environments around Github than understanding that "product" in Github corresponds to "prod" space in PaaS. But open to change this if the team does not agree.

frankieroberto: 👍 for keeping it as "production". Seems unnecessary and unhelpful to abbreviate.

.github/workflows/deploy.yml
```diff
@@ -17,7 +17,50 @@ jobs:
...
+          "description": "Staging deploy",
+          "environment": "staging",
+          "auto_merge": false,
+          "required_contexts": []
+          }' | jq '.url?' | tr -d '"') # Gets 'url' field from the response and trims the surronding quotes.
+
+        if [ -z "$DEPLOY_URL" ]; then
+          echo "Failed to create Github deployment"
+        else
+          echo "::set-env name=DEPLOY_URL::$DEPLOY_URL"
```
frankieroberto: Could you use the [outputs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/metadata-syntax-for-github-actions#outputs) syntax here, instead of setting environment variables?

.github/workflows/deploy.yml
```diff
@@ -17,7 +17,50 @@ jobs:
...
+          -H "Authorization: token $GITHUB_TOKEN" \
+          -H "Accept: application/vnd.github.ant-man-preview+json" \
+          -H "Accept: application/vnd.github.flash-preview+json" \
+          $DEPLOY_STATUSES_URL \
+          -d '{
+            "environment": "staging",
+            "state": "in_progress",
+            "description": "Staging deployment initiated",
+            "log_url": "'"$ACTIONS_DEPLOY_URL"'"
+          }'
```
frankieroberto: refactor to a separate script so that it can be shared between staging and production?

.github/workflows/deploy.yml
```diff
@@ -17,7 +17,50 @@ jobs:
...
         sudo cp /tmp/cf7 /usr/local/bin/cf7
-    - name: Deploy to staging
+
+    - name: Create GitHub deployment for Staging
+      env:
+        GITHUB_TOKEN: ${{ secrets.GithubApiDeploymentToken }}
+      run: |
+        DEPLOY_URL=$(curl -X POST \
+          -H "Authorization: token $GITHUB_TOKEN" \
+          https://api.github.com/repos/UKGovernmentBEIS/beis-opss/deployments \
```
frankieroberto: Use the [`GITHUB_REPOSITORY`](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables) environment variable here (so that it doesn't break if we rename the repo)?

.github/workflows/deploy.yml
scruti: I do remember checking that option but then, if I understood it correctly, realising that `outputs` belongs to the `metadata` file scope of a JS Github action. 

So is not something available for sharing information between steps inside a workflow but to define "action FOO has X input and Y outputs".

Did I understand it wrong? I find the boundaries between documentation on Github JS actions
and workflows (that can use those actions) quite blurry.

frankieroberto: @scruti I think you're right. Reading the documentation again, it seems to suggests that outputs can be used for sharing data between different actions within the same workflow, but environment variables can be used for sharing data between different steps in the same action?

scruti: Yeah, I understood that too (thanks for the better wording!)

.github/workflows/deploy.yml
scruti: Going to explore to extract all the related bash code from here and from review workflow into bash functions.

scruti:
Switching back to WIP while try to address some of the suggestions.

frankieroberto:
@scruti feel free to leave any refactoring for future branches/pull requests, if you just want to get this merged.

scruti:
@frankieroberto extracted all Github Deployment managament logic into a script with bash functions. Ready for another review.

cosmetics-web/deploy-github-functions.sh
```diff
@@ -0,0 +1,145 @@
...
+# - curl
+gh_deploy_success() {
+  environment_name=$1
+  if [ "$environment_name" == "int" ]; then
+    environment_url="https://cosmetics-pr-${PR_NUMBER}-submit-web.london.cloudapps.digital/"
+  elif [ "$environment_name" == "staging" ]; then
+    environment_url="https://staging-submit.cosmetic-product-notifications.service.gov.uk/"
+  elif [ "$environment_name" == "production" ]; then
+    environment_url="https://submit.cosmetic-product-notifications.service.gov.uk/"
+  fi
```
frankieroberto: Could we move this from the bash script into the GitHub Workflow definition, eg so that it can called by something like:

```bash
gh_deploy_success("int", "https://cosmetics-pr-${PR_NUMBER}-submit-web.london.cloudapps.digital/")
```

(that way we can re-use the exact same script for PSD... 😉  )

I also suspect that, rather than giving all the review apps the same environment name of `int`, they should be unique-per-review-app (eg `review-app-${PR_NUMBER}`)? That's how Heroku review apps seem to have done their integration anyhow.

cosmetics-web/deploy-github-functions.sh
```diff
@@ -0,0 +1,145 @@
...
+  environment_name=$1
+  deploy_url=$(curl -X POST \
+    -H "Authorization: token $GITHUB_TOKEN" \
+    https://api.github.com/repos/$GITHUB_REPOSITORY/deployments \
+    -d '{
+    "ref": "'"$BRANCH"'",
+    "description": "'"$environment_name"' deploy created",
+    "environment": "'"$environment_name"'",
+    "auto_merge": false,
+    "required_contexts": []
```
frankieroberto: should we add `transient_environment: true` for review apps? see https://developer.github.com/v3/repos/deployments/#parameters-1

cosmetics-web/deploy-github-functions.sh
```diff
@@ -0,0 +1,145 @@
...
+# - curl
+# - jq
+# - awk
+gh_deploy_create() {
+  environment_name=$1
+  deploy_url=$(curl -X POST \
+    -H "Authorization: token $GITHUB_TOKEN" \
+    https://api.github.com/repos/$GITHUB_REPOSITORY/deployments \
+    -d '{
+    "ref": "'"$BRANCH"'",
```
frankieroberto: I _think_ this should be the SHA of the current commit, rather than the branch. That way, I think GitHub additionally shows a "View deployment" button next to the commit that’s currently deployed, making it clearer to see whether there are some new commits within the branch that haven't yet been deploy?

cosmetics-web/deploy-github-functions.sh
scruti: Good ideas, going to explore it.

cosmetics-web/deploy-github-functions.sh
scruti: I've been struggling with the commit based pre-packed Github Deploy actions. So I switched to branch based approach in order to get the deploy correctly linked to the PR.
Maybe the issue wasn't on that, so going to give it a try as the "View deployment" button next to the commit is definitely an improvement. Good suggestion :+1:

cosmetics-web/deploy-github-functions.sh
scruti: I am terrified from creating new "environments" on Github as they cannot be deleted once created. I wonder if transient environments will actually avoid this. Will read more on this and maybe explore it in a separate branch afterwards.

cosmetics-web/deploy-github-functions.sh
scruti: Attempted this but the commit SHA used as a reference was not matching any commit in the branch, resulting in the deploy not being associated with the Pull Request.
I did reset it to use the branch as reference, taking in account that even with the branch approach, the Github Deploy info box will show when the last deployment is outdated due to new commits added to the branch.

scruti:
Ass suggested by @frankieroberto we moved moved the environment URLS logic from the bash script to the workflow, so the bash script is not project dependant.
Didn't implement the transient environments or use commits instead of branches as reference. This can be studied/implemented in further follow ups on these workflows.

Going to merge it in this state.

PR #1657, opened 2020-02-03
Set individual Github environments for reviewapp deploys

scruti:
## Description
All the reviewapps are being deployed under "int" environment on Github deployments, matching the BEIS PaaS "int" Space that hosts all the review apps.

This approach has a critical downside though: 
**New PRs deploys are considered as the latest version of the same "int" environment**, and Github Deployment tags
the previously deployed PRs as "innactive", removing he deployment information and environment URL.
As a consequence of this, we cannot have multiple PRs opened in parallel showing Deployment & Environment information.

## Solution
In order to avoid this, we are creating individual environments for each Pull Request number. So multiple review app environments will be cohabiting in Github Deployment domain without stepping over each other.
This approach was suggested by @frankieroberto [over previous PR](https://github.com/UKGovernmentBEIS/beis-opss/pull/1646#discussion_r373568175) (thanks! :clap: )

## Example of the issue when sharing "int"
Two different PRs, the first one has been deployed under "int". The second PR deployment has started:
Screenshot from 2020-02-03 15-22-20

Once the second PR finishes, the firs PR "View deployment" button disappears, as that deploy gets considered outdated and marked as "inactive":
Screenshot from 2020-02-03 15-26-12

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/deploy-github-functions.sh
+++ b/cosmetics-web/deploy-github-functions.sh
@@ -4,7 +4,7 @@ set -e
 # Creates a Deploy in Github
 #
 # Input:
-# - environment to deploy at.
+# - Name of the environment to deploy at.
 #   eg: $ gh_deploy_create staging
 #
 # Required environment variables:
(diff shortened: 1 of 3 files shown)


cosmetics-web/deploy-github-functions.sh
```diff
@@ -19,25 +19,30 @@ set -e
 # - awk
 gh_deploy_create() {
   environment_name=$1
+
+  if [[ $environment_name == review* ]]; then
+    transient_environment=true
+  else
+    transient_environment=false
+  fi
```
frankieroberto: Maybe make this a param of the function too, instead of making assumptions based on the environment name? (Could default to false, assuming bash has optional params?)

scruti: Sure, no idea why I keep coupling the Bash script with the workflow specific logic.

PR #1662, opened 2020-02-06
Default to Submit page when not specifically requesting Search pages

scruti:
[COSBETA-816](https://regulatorydelivery.atlassian.net/browse/COSBETA-816)

## Description
Until now, when loading the Cosmetics application in our machines for development, we needed to set up SUBMIT_HOST and SEARCH_HOST in our environment variables, and access the Cosmetics in our machines by using either of those values as URL.

Attempting to load the service by using: "http://localhost:300x" was resulting into an error page.

For development purposes we want to default to the Submit page when requesting the service from our computers with "http://localhost:300x" or from docker with "127.0.0.1:300x". 

## Additional changes
- Did some housekeeping with the route domain constraints, plus solved some issue with the actual routing. All the changes & reasoning behind them are described in their respective commits. 

- Changed the approach for stubbing search/submit hosts during testing. Instead of stubbing methods to return true/false, moved it to stub the request host directly. This makes code changes lest brittle and allows end to end request & feature testing without part of our codebase being stubbed.

## How to review this PR
I encourage to review it commit by commit. 

### Before
image

### After
image

## Checklist:

### DIFF
--- a/cosmetics-web/config/routes.rb
+++ b/cosmetics-web/config/routes.rb
@@ -34,7 +34,8 @@ Rails.application.routes.draw do
     end
   end
 
-  constraints DomainInclusionConstraint.new(ENV.fetch("SUBMIT_HOST") || ENV.fetch("COSMETICS_HOST")) do
+  domains = "#{ENV.fetch('SUBMIT_HOST') || ENV.fetch('COSMETICS_HOST')},localhost"
+  constraints DomainInclusionConstraint.new(domains) do
     root "landing_page#index"
 
 
--- a/cosmetics-web/config/routes.rb
+++ b/cosmetics-web/config/routes.rb
@@ -26,7 +26,7 @@ Rails.application.routes.draw do
     mount Sidekiq::Web => "/sidekiq"
   end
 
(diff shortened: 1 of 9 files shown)


cosmetics-web/config/routes.rb
```diff
@@ -34,7 +34,8 @@
     end
   end
 
-  constraints DomainInclusionConstraint.new(ENV.fetch("SUBMIT_HOST") || ENV.fetch("COSMETICS_HOST")) do
+  domains = "#{ENV.fetch('SUBMIT_HOST') || ENV.fetch('COSMETICS_HOST')},localhost"
```
slorek: This is better than we have currently, but I wonder if we can simply default to the service in the case there are no matches, instead of matching to the `localhost` string, in the event that the developer is using some other domain, for instance when loading via docker-compose.

cosmetics-web/spec/requests/notifications_page_spec.rb
```diff
@@ -5,17 +5,18 @@
     sign_out
   end
 
-  context "when signed in as a poison centre user", with_errors_rendered: true do
+  context "when signed in as a poison centre user but accesing from submit domain", with_errors_rendered: true do
```
slorek:
```suggestion
  context "when signed in as a poison centre user but accessing from submit domain", with_errors_rendered: true do
```

PR #1664, opened 2020-02-07
Fix coveralls

slawosz:


.github/workflows/tests.yml
```diff
@@ -93,4 +93,4 @@ jobs:
         COVERALLS_REPO_TOKEN: ${{secrets.CoverallsToken}}
       run: |
         cd cosmetics-web
-        bundle exec rspec spec
+        GIT_BRANCH=${GITHUB_REF##*/} bundle exec rspec spec
```
frankieroberto: Add a comment to explain what this does / why it’s needed?

PR #1751, opened 2020-09-17
New team member invitation flow

scruti:
## Description
Changing the process of inviting users to a team for Submit service

invite a team member to a cosmetics org@2x (1) (1)

## Highlights
- Now we are sending individual invitation tokens instead of the previous approach where invitation links were generic and relying
on the user being already registered in the service with the same email account as some pending invitation.
- Users will be able to register into the service through the team invitation link. Once completed their registration they will be automatically added to the team.

## Checklist:


cosmetics-web/app/controllers/responsible_persons/team_members_controller.rb
```diff
@@ -16,24 +18,36 @@ def create
...
-      Rails.logger.info "Team member added to Responsible Person"
-      pending_requests.delete_all
-    end
+    return render("join_signed_in_as_another_user", locals: { existing_user: invited_user }) if current_user && current_user.email != pending.email_address
+    return redirect_to registration_new_submit_user_path unless invited_user
+    return authenticate_user! unless current_user
+
+    # User at this point is signed as the invited user.
+    @responsible_person.add_user(current_user)
+    Rails.logger.info "Team member added to Responsible Person"
```
slawosz, line 31: For logging to make sense, I would recommend add some ids here.

cosmetics-web/spec/features/inviting_a_colleague_spec.rb
```diff
@@ -21,22 +13,181 @@
...
     expect(invited_user.responsible_persons).to include(responsible_person)
   end
 end
+
+def expect_to_be_on_check_your_email_page
+  expect(page).to have_css("h1", text: "Check your email")
+  expect(page).to have_css(".govuk-body", text: "A message with a confirmation link has been sent to your email address.")
+end
+
+def complete_secondary_authentication_for(user)
```
scruti, line 189: Comment by @slawosz : There is already a helper for this.
To Do: Use that helper.

nasirkhanbeis:
Tested this in review app
Need to fix the invite journey as there some confusion around the flow.
Also when I click on invite link again after invite complete, it redirects to empty page . 
image
Expected to return home page instead empty page

PR #1756, opened 2020-09-23
Inviting team members POC

slawosz:
Different approach to inviting team member. It:
- simplifies process by using current registration flow
- instead of using `PendingResponsiblePersonUser`, it simply adds invited user to `ResponsiblePerson` (via `ResponsiblePersonUser`
- on invite form, reuses `Registration::NewUserForm`

To make it working:
- [ ] send proper email notification
- [ ] determine team member status based on the `User` `registration_completed` field
- [ ] move tests from @scruti PR

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/team_members_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/team_members_controller.rb
@@ -1,34 +1,34 @@
 class ResponsiblePersons::TeamMembersController < ApplicationController
   before_action :set_responsible_person
-  before_action :set_team_member, only: %i[new create]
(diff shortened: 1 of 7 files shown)


nasirkhanbeis:
Invite happy path journey looks good however few issues to sort yet
1. Invite an existing user doesn't throw an error , expected to throw error as user already exist in different organisation
2. Invite partial user setup . e.g self registered user but didn't confirm email, when same user is invited to a team they should still be able to complete invite and join invited team ( need to check with Ralph)
3. Need to fix validation issues in invite form page

nasirkhanbeis:
Also when I click join link again in the email, it returns as token expired. Expected to return to home page I think

PR #1759, opened 2020-09-25
Submit and search subcontrollers with invitation

slawosz:


cosmetics-web/app/controllers/declaration_controller.rb
```diff
@@ -10,6 +8,7 @@ def show
 
   def accept
     current_user.update(has_accepted_declaration: true)
+    # return redirect_to root_path
```
scruti, line 11: :scissors: ?

scruti (approved):
LGTM! Great work on this.
Is worth it only for simplifying the before_action/skip_before_action filters :clap:

cosmetics-web/app/controllers/declaration_controller.rb
scruti, line 11:
```suggestion
```

PR #1760, opened 2020-09-29
Separate notify instances for search and submit

slawosz:
Split notify mailer to use 2 different accounts of service. Mainly to allow proper 'from' name in the email.


cosmetics-web/app/models/user.rb
```diff
@@ -11,6 +11,6 @@ class User < ApplicationRecord
   validates :name, presence: true, unless: -> { invite }
 
   def send_new_email_confirmation_email
-    NotifyMailer.new_email_verification_email(self).deliver_later
+    NotifyMailer.get_mailer(self).new_email_verification_email(self).deliver_later
```
scruti, line 14: I wonder if we could just define a `MAILER = SubmitNotifyMailer` on `SubmitUser` and `MAILER = SearchNotifyMailer` on `SearchUser` so we can remove the need of `get_mailer` and calls like this one can be turned into:

```ruby
MAILER.new_email_verification_email(self).deliver_later
```

and shared concerns like this call:
```ruby
NotifyMailer.get_mailer(user).update_email_address_notification_email(user, old_email).deliver_later
```
turned into:

```ruby
user.class::MAILER.update_email_address_notification_email(user, old_email).deliver_later
```
or even encapsulate this call withing the respective user class (like done in methods as `send_reset_password_instructions_notification(token)`):
```ruby
user.update_email_address_notification_email(old_email)
```

What do you think?

scruti (changes requested):
Looks good!

The requested changes are only on the Gemfile issue. 
The other comment about setting each mailer class in their respective user class is more a suggestion to consider.

PR #1729, opened 2020-07-15
WIP: Keycloak replacement

slawosz:
Before merge
- [x] @nasirkhanbeis to make sure all works
- [x] @ralph-hawkins to make sure all wording is spot on
- [x] @tobyhughes or Bruce to make sure it looks good
- [x] other developer will check for obvious pitfals

PR #1766, opened 2020-10-06
WIP/BLOCKED Remove any resp. person invitations when user joins a responsible person

scruti:
## Blocked

Currently blocked by https://regulatorydelivery.atlassian.net/browse/COSBETA-888

## Description
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-876)

An user joins a Responsible Person by either:
- Accepting an invitation.
- Creating a Responsible Person itself on registration.

In either case, any pending invitation to join a responsible person should be removed, so the user does not appear in the team member list as "pending" when it actually belongs to another resp. person.


cosmetics-web/app/controllers/responsible_persons/team_members_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
     return render :new if @responsible_person.errors.any?
 
     send_invite_email
-    redirect_to responsible_person_team_members_path(@responsible_person), confirmation: "Invite sent to #{team_member_params["email_address"]}"
+    redirect_to responsible_person_team_members_path(@responsible_person), confirmation: "Invite sent to #{team_member_params['email_address']}"
```
scruti, line 18: Unrelated rubocop autofix.

cosmetics-web/app/helpers/form_helper.rb
```diff
@@ -7,7 +7,7 @@ def form_input(user, field, options = {})
       name: "#{resource_form_name(user)}[#{field}]",
       type: "text",
       classes: "app-!-max-width-two-thirds",
-      label: { text:  user.class.human_attribute_name(field.to_sym)},
+      label: { text: user.class.human_attribute_name(field.to_sym) },
```
scruti, line 10: Unrelated rubocop autofix

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -51,7 +52,7 @@ def responsible_person_valid?
     @responsible_person.valid?(step)
   end
 
-  def responsible_person_saved?
+  def save_responsible_person
```
scruti, line 55: Changed the name as found the `?` naming misleading. This is not a query, is an command instruction that can success/fail.

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -19,7 +19,8 @@ def update
     when :create_or_join_existing
       create_or_join_existing_account
     when :enter_details
-      if responsible_person_saved?
+      if save_responsible_person
+        PendingResponsiblePersonUser.where(email_address: current_user.email).delete_all
```
scruti, line 23: I have considered the possibility of, as we have the business rule of deleting any matching pending responsible person user every time we add an user to a responsible person, to make this `Pending...delete_all` instruction part of the `ResponsiblePerson#add_user` implementation (or a callback over the method). 
Ended deciding against it:

This `add and delete` is happening now in two places: This bit and when [joining from invitation ](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1729/files#diff-603531de52ef0d10f13713c29d9d611eR32)

By leaving this in this state we have possibility of someone adding a new flow for joining a responsible person and forgetting about including the invitation removal. 
But given that is just a simple instruction I believe is worth the risk and outcomes the downsides of burying this business logic behaviour related to PendingResponsiblePersonUser into ResponsiblePerson model, forcing the later class to know too much about the former and coupling them.

If the amount of behaviour we want to associate around joining a responsible person growths, I personally would extract a class abstracting the `ResponsiblePersonUserAdition` concept and associated behaviour. 

But I think this would be an overkill now. So tried to be pragmatic and left it as simple as possible.

scruti:
Closing as this became incompatible with incoming feature for users belonging to multiple responsible persons.

PR #1772, opened 2020-10-12
Fix email validation error link on team user invite

scruti:
When clicking on the email validation error the focus wasn't switching to the email input.
Fixed if by changing the email input id to match the attribute with the validation error.


nasirkhanbeis (approved):
looks good as expected!

slawosz (approved):

PR #1787, opened 2020-10-15
Allow users to belong to multiple RPs

slawosz:


cosmetics-web/app/models/pending_responsible_person_user.rb
```diff
@@ -11,7 +11,7 @@ class PendingResponsiblePersonUser < ApplicationRecord
             }
   validates_presence_of :email_address, message: I18n.t(:blank, scope: EMAIL_ERROR_MESSAGE_SCOPE)
   validate :email_address_not_in_team?
-  validate :email_address_not_in_other_team?
+  # validate :email_address_not_in_other_team?
```
scruti, line 14:
```suggestion
```
I guess at this point the method [`#email_address_not_in_other_team?` ](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/blob/637390390a70166a186c6e6f528dbbec043f609f/cosmetics-web/app/models/pending_responsible_person_user.rb#L46-L51) is not used anymore and should be :scissors: too.

cosmetics-web/app/controllers/responsible_persons_controller.rb
```diff
@@ -1,5 +1,20 @@
...
+
+  def show; end
+
+  def select; end
+
+  def change
+    # TODO: assign current_responsible_user_id properly
+    # TODO: security - make sure in such case user cant access wrong rp
+
+    # TODO: spec for this line
```
scruti, line 14: :question:  Are these TODO going to be implemented in this PR?

cosmetics-web/db/migrate/20201013165932_add_current_responsible_person_id_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddCurrentResponsiblePersonIdToUsers < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :current_responsible_person_id, :integer
+  end
+end
```
scruti, line 5: Please change: :scissors:  
This has been moved to the session and we're not storing it in DB anymore right?

cosmetics-web/db/schema.rb
```diff
@@ -277,6 +278,7 @@
     t.string "new_email_confirmation_token"
     t.datetime "new_email_confirmation_token_expires_at"
     t.boolean "account_security_completed", default: false
+    t.integer "current_responsible_person_id"
```
scruti, line 281:
```suggestion
```

nasirkhanbeis (approved):
looks good as expected!

cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def validate_responsible_person
+    return if @responsible_person.nil?
+
+    if @responsible_person != current_responsible_person
+      redirect_to select_responsible_persons_path
+    end
+  end
+end
```
scruti, line 15: I found myself going around different files and scrolling up and down (and ended copy pasting bits in a notebook) when trying to understand things like:
- Where do we set `@responsible_person` ?
- Where do we set `current_responsible_person` ?
- What does it mean `validate_responsible_person`? 

In case it does help, this is the resulting structure of how we call responsible person related methods:
([URL for notes instead of images](https://www.notion.so/marcsardon/Review-Slawosz-multiple-RP-PR-590f5723534044488165a3ea0c47d14a))
image
image

Based on this, I think we could improve this by either:
- Rename this concern to `ValidateCorrectResponsiblePersonConcern` or similar, as the current name is so generic that I personally found confusing it having it around other `responsible_person` related calls/filters defined somewhere else.
- Instead of making the concern something very particular. Embrace the `ResponsiblePersonConcern` and move into it all the controller method definitions related to responsible persons. Things like:
  - `def current_responsible_person`
  - `def create_or_join_responsible_person`
  - `def set_current_responsible_person(responsible_person)` 
  - `def set_responsible_person`  
Then we would just to need to add `include ResponsiblePersonConcern` inside `SubmitApplicationController`, and, depending on the use of `included do ...` on the concern, play with `before_action` or `skip_before_action` in each controller inheriting from the `SubmitApplicationController`.  I personally favour this approach, but didn't attempt to implement it so may turn more complex than I originally foreseen.

scruti (commented):
Great work on this. 
Left a few suggestions, some questions and some change requests (the test descriptions one).

The main thing, besides wording, is a proposal to either group all the controller responsible_person related methods into the `ResponsiblePersonConcern` or either rename the concern to make it more obvious.

ralph-hawkins:
Looks good to me 👍

cosmetics-web/spec/factories/responsible_person.rb
```diff
@@ -1,13 +1,16 @@
 FactoryBot.define do
   factory :responsible_person do
     account_type { :individual }
-    name { "Responsible Person" }
+    sequence(:name) { |n| "Responsible Person #{n}" }
```
scruti, line 4: Nice!

scruti (approved):
Great work! :ship:

PR #1788, opened 2020-10-16
Remove some Keycloak config/mentions left behind

scruti:
This is not used or needed to be mentioned in the Readmes as we do not
use Keycloak anymore.


slawosz:
I docker removed from compose file as well?

scruti:
@slawosz there was already no mention to keycloak in docker compose file (while was indirectly referenced from `db` Dockerfile removed in this PR).

slawosz (approved):

PR #1792, opened 2020-10-19
Do not check for mobile number verification when 2FA is disabled

scruti:
Originally solving a bug, but now just turned into a method code simplification.
We do not need to check for user mobile number verification unless the 2FA is enabled in the environment, as the mobile number gets verified when passing 2FA.

### DIFF
--- a/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
@@ -3,7 +3,6 @@ module ResponsiblePersonConcern
 
   def create_or_join_responsible_person
     return unless fully_signed_in_submit_user?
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/controllers/submit_application_controller.rb
```diff
@@ -34,7 +33,7 @@ def create_or_join_responsible_person
 
   def fully_signed_in_submit_user?
     if Rails.configuration.secondary_authentication_enabled
-      user_signed_in? && secondary_authentication_present?
+      user_signed_in? && secondary_authentication_present? && current_user.mobile_number_verified?
```
scruti, line 36: The mobile number will be verified only when the secondary authentication is enabled.

slawosz (approved):

PR #1794, opened 2020-10-20
Remove header and "Active" from team members list

scruti:
[Jira Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-887)

When team members are listed, we don't want to display "Status" header or "Active" as a value.
We just want to display "Resend invitation" link when the team member is an user with an invitation instead of an active member.

Maintained the 3 columns (even when empty) for visual reasons so the table has the same proportions independently of the elements having a "Resend Invitation" link or not.

### Before
Screenshot from 2020-10-19 16-35-53

## After
Screenshot from 2020-10-19 16-36-29


nasirkhanbeis (approved):
looks good as expected

PR #1800, opened 2020-10-23
Minor user account section fixes

slawosz:
* Don't require password confirmation when changing password
* Don't allow Search Users to change email and mobile phone


scruti (approved):
What this required by any tickets? (Can you link the PR to them if so?)

PR #1802, opened 2020-10-27
Contact person styling fix

ralph-hawkins:
Simplifies the way we tell the user which responsible person they are creating a contact person for.


nasirkhanbeis (approved):
Looks good as expected 
image
I see code style issues and some tests failed, need to fix before merging this in

PR #1815, opened 2020-11-06
Show user responsible person pending invitations after registration

scruti:
[Ticket COSBETA-667](https://regulatorydelivery.atlassian.net/browse/COSBETA-667)

## Description
An email address has been invited to one/multiple responsible persons by one/multiple members of the team.
The user for that email instead of following the invitation link registers to the Submit Cosmetics service like any other user.
The user will be prompted with a page that will show the list of Responsible Persons they have been invited to.
For each Responsible Person the user has been invited to:
- If there are active invitations for that responsible person, the user will be displayed the date of the latest invitation sent.
- If all the invitations for that responsible person are expired, users will be displayed the name of the RP members who invited them, so they can contact them asking for be re-invited.

## Implementation notes
As part of the work, we added the possibility of having multiple invitations for the same email on the same responsible person, as long as the users who sent them are different. We modified the Team Members page to display the user who sent the invitation for the pending users elements, so they don't seem duplicated and each invitation can be re-sent independently through their link.

**Note: Ignore the line wrapping of the name on the screenshot, this has been fixed**
Screenshot from 2020-11-09 12-27-02

Screenshot from 2020-11-10 09-27-45

## Checklist:

### DIFF
--- a/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
@@ -9,6 +9,8 @@ module ResponsiblePersonConcern
     if responsible_person.blank?
       if current_user.responsible_persons.present?
         redirect_to select_responsible_persons_path
+      elsif pending_responsible_person_invitations.any?
+        redirect_to account_path(:pending_invitations)
       else
         redirect_to account_path(:overview)
       end
@@ -46,5 +48,10 @@ private
     else
       user_signed_in?
     end
+  end
+
+  def pending_responsible_person_invitations
+    @pending_responsible_person_invitations ||=
+      PendingResponsiblePersonUser.where(email_address: current_user.email).includes(:responsible_person)
   end
 end
--- a/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
@@ -9,6 +9,8 @@ module ResponsiblePersonConcern
     if responsible_person.blank?
       if current_user.responsible_persons.present?
         redirect_to select_responsible_persons_path
+      elsif pending_invitations.any?
+        redirect_to account_path(:pending_invitations)
       else
         redirect_to account_path(:overview)
       end
--- a/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
@@ -9,7 +9,7 @@ module ResponsiblePersonConcern
     if responsible_person.blank?
       if current_user.responsible_persons.present?
         redirect_to select_responsible_persons_path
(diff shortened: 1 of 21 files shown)


cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
```diff
@@ -47,4 +49,25 @@ def fully_signed_in_submit_user?
...
+  def invitations_grouped_by_responsible_person
+    pending_invitations.each_with_object(ActiveSupport::OrderedHash.new) do |invitation, hash|
+      rp_id = invitation.responsible_person_id
+      if hash[rp_id]
+        hash[rp_id] << invitation
+      else
+        hash[rp_id] = [invitation]
+      end
+    end
+  end
```
scruti, line 72: This is going to be refactored to somewhere else. Bear with me :raised_hands:

cosmetics-web/app/views/responsible_persons/account_wizard/pending_invitations.html.erb
```diff
@@ -0,0 +1,43 @@
...
+            Check your email inbox for your invite, sent <%= display_full_month_date(invitations.first.created_at) %>.
+          <% end %>
+        <% else %>
+          <% invitation = invitations.first %>
+          <% if invitation.expired? %>
+            Your invite has expired and needs to be resent. You were invited by <%= invitation.inviting_user.name %>."
+          <% else %>
+            Check your email inbox for your invite, sent <%= display_full_month_date(invitation.created_at) %>.
+          <% end %>
+        <% end %>
```
scruti, line 34: This is going to be refactored (I guess with a decorator). Bear with me :raised_hands:

cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
```diff
@@ -47,4 +49,17 @@ def fully_signed_in_submit_user?
       user_signed_in?
     end
   end
+
+  def pending_responsible_persons_invitations_text
```
slawosz, line 53: Is it something that should be in controller? Also name is bit ambiguous. But if you don't have better idea, good enough.

slawosz (approved):
LGTM

cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
scruti: I could move it to the view to be honest, specially when is only being used in one and the controller already provides the data needed for the presenter. Going to move it. About the naming... will try to figure out something better while I move it around, but naming things is hard xD

nasirkhanbeis (approved):
looks good as expected. 
Verified invite list page, expired invites, registration journey, declaration flag all good. 
Also back link works as expected. 
Note: currently every user within RP can invite same email which creates independent token which means team pages may show more one row invites for same email but invited by different user.

PR #1818, opened 2020-11-10
Refactor notification wizard

slawosz:
Reorganize files so its easier to follow what is happening in wizard.


cosmetics-web/app/controllers/responsible_persons/additional_information_controller.rb
```diff
@@ -1,4 +1,5 @@
-class AdditionalInformationController < SubmitApplicationController
+# candidate
```
scruti, line 1: :scissors: ?

scruti (approved):
Big thanks for doing this :pray:

PR #1827, opened 2020-11-17
Only allow one responsible person invitation per email and rp.

scruti:
[Jira Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-943)
- Only allow one invitation per email and team.
- Re-sending invitation update the inviting user with the user who clicked the link.

## Description
We are removing the possibility of having multiple invitations by different team members to the same email and same responsible person/team.
The service will allow a single invitation per email and responsible person/team.
This allows us to simplify a lot the view logic associated to the pending responsible person invitations page displayed when users with pending invitations register in the service.

On top of this, when a different team member (than the one who sent the team invitation originally) re-sends an invitation, the invitation will be updated so this user who clicked the link will figure as the "inviting user" on the invitation.

image


nasirkhanbeis (approved):
tested this on review app 1827, looks good as expected.

slawosz (approved):

PR #1834, opened 2020-11-19
Bugfix: Registration flow broken when 2FA is disabled

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-961)

**Bugfixes:**
- Fix registration when 2FA is disabled, avoiding declaration being shown before account security step.
- Fix users registered through an invitation link not being marked as `account_security_completed`.

**First issue:**
Currently the registration flow is broken **when 2FA is disabled** due to the Responsible person Declaration and Creation/Join  filters being triggered in the account security step.

These filters are skipped when 2FA is enabled as, when reached at registration account security point, the user is not "signed in". But when 2FA is disabled, the user is "signed in" and the filters are applied.

In order to solve this we're extending the conditions, so the user needs to have completed its registration in order for the filters to run.

**Second issue**
There were some issues with the declaration page in the feature specs corresponding to the user registration through an invitation flow. By flagging "account_security_completed" when the user submits their username, password and mobile number, now the behaviour of the Declaration page became consistent through the specs.

### DIFF
--- a/cosmetics-web/app/controllers/application_controller.rb
+++ b/cosmetics-web/app/controllers/application_controller.rb
@@ -73,8 +73,7 @@ private
   def authorize_user!; end
 
   def has_accepted_declaration
-    return unless user_signed_in?
-    return unless current_user.mobile_number_verified?
+    return unless user_signed_in? && current_user&.account_security_completed? && current_user&.mobile_number_verified?
 
     redirect_path = request.original_fullpath unless request.original_fullpath == root_path
 
--- a/cosmetics-web/app/controllers/application_controller.rb
+++ b/cosmetics-web/app/controllers/application_controller.rb
@@ -73,7 +73,9 @@ private
   def authorize_user!; end
 
   def has_accepted_declaration
-    return unless user_signed_in? && current_user&.account_security_completed? && current_user&.mobile_number_verified?
+    return unless user_signed_in?
+    return unless current_user&.account_security_completed?
+    return unless current_user&.mobile_number_verified?
 
     redirect_path = request.original_fullpath unless request.original_fullpath == root_path
 
--- a/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
@@ -44,10 +44,14 @@ private
 
   def fully_signed_in_submit_user?
     if Rails.configuration.secondary_authentication_enabled
-      user_signed_in? && secondary_authentication_present? && current_user.mobile_number_verified?
+      fully_registered_user? && secondary_authentication_present? && current_user.mobile_number_verified?
     else
-      user_signed_in?
+      fully_registered_user?
     end
+  end
+
+  def fully_registered_user?
+    user_signed_in? && current_user&.account_security_completed?
   end
 
   def pending_invitations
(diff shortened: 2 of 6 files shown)


cosmetics-web/app/controllers/registration/account_security_controller.rb
```diff
@@ -3,6 +3,8 @@ class AccountSecurityController < SubmitApplicationController
     before_action :check_user
     skip_before_action :require_secondary_authentication
     skip_before_action :try_to_finish_account_setup
+    skip_before_action :has_accepted_declaration
+    skip_before_action :create_or_join_responsible_person
```
slawosz, line 7: I wonder if there makes sense to add something like `if: two_fa_disabled?`, or maybe change the implementation of said methods to execute only for certain user state?

slawosz (approved):

cosmetics-web/app/controllers/registration/account_security_controller.rb
scruti: Good, point. Changing to explore that approach,.

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -73,8 +73,7 @@ def after_sign_in_path_for(resource)
   def authorize_user!; end
 
   def has_accepted_declaration
-    return unless user_signed_in?
-    return unless current_user.mobile_number_verified?
+    return unless user_signed_in? && current_user&.account_security_completed? && current_user&.mobile_number_verified?
```
slawosz, line 76: I personally prefer not to inline this conditions.

cosmetics-web/app/controllers/concerns/responsible_person_concern.rb
```diff
@@ -44,12 +44,16 @@ def validate_responsible_person
...
     if Rails.configuration.secondary_authentication_enabled
-      user_signed_in? && secondary_authentication_present? && current_user.mobile_number_verified?
+      fully_registered_user? && secondary_authentication_present? && current_user.mobile_number_verified?
     else
-      user_signed_in?
+      fully_registered_user?
     end
   end
 
+  def fully_registered_user?
```
slawosz, line 53: I am not sure if the naming is right. You check if user is registered (at least name suggests so) - which is DB state, but you also check if user is signed in - which is browser state.

cosmetics-web/app/controllers/users_controller.rb
```diff
@@ -30,7 +30,7 @@ def update
     @user = SearchUser.find(params[:id])
     return render("errors/forbidden", status: :forbidden) if params[:invitation] != @user.invitation_token
 
-    @user.assign_attributes(new_user_attributes)
+    @user.assign_attributes(new_user_attributes.merge(account_security_completed: true))
```
slawosz, line 33: Why we are adding `account_security_completed` here?

slawosz (approved):

cosmetics-web/app/controllers/users_controller.rb
scruti: Because, when registering through an invitation link, the user is setting its mobile number and password here instead of through the "account security" page. 
Without adding `account_security_completed` here, we end in an inconsistent state between users, as we would have:
- `account_security_completed? true` for self-registered users that have completed account security page.
- `account_security_completed? false` for users who registered through an invitation link and completed the account security through the completion form.

So, when wanting to filter the `declaration` page or the `create_or_join_responsible_person` page to users that have completed their registration (including account security) it would not work properly if the check returns `false` when the user registered through an invitation.

PR #1844, opened 2020-11-26
948 upload ingredients from zip

slawosz:
Proper template when adding ingredients files for components from zip file.


nasirkhanbeis:
looks good to me, detects type of formulation and prompts user on right page. However upload image seems to loop forever for post brexit zip file upload. Please try with attached test file.
[Uploading Multi-Item-RangeDoc_pHRange_ExactDoc_Nano 3.zip…]()

nasirkhanbeis:
Re-tested above issue, seems to be images not being attached at item level but rather at product which looks like different issue. so happy to approve this.

nasirkhanbeis (approved):
works fine as expected. detects formulation type and puts user to the right page.

scruti (approved):
LGTM!

PR #1850, opened 2020-12-01
Allow duplicated RP names

scruti:
[Jira ticket COSBETA-945](https://regulatorydelivery.atlassian.net/browse/COSBETA-945)

Allow duplicated RP names unless when attempting to create a responsible person with the same name as another RP either:
- we're member of.
- we have been invited to.

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
@@ -59,6 +59,14 @@ private
   def responsible_person_saved?
     return false unless responsible_person_valid?
 
+    if current_user.responsible_persons.any? { |rp| rp.name = @responsible_person.name }
+      @responsible_person.errors.add(:name, "You are already a member of #{@responsible_person.name}")
+      return false
+    elsif PendingResponsiblePersonUser.includes(:responsible_person).where(email_address: user.email, responsible_persons: { name: @responsible_person.name }).any?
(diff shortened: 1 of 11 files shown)


cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
```diff
@@ -9,6 +9,7 @@ class ResponsiblePersons::AccountWizardController < SubmitApplicationController
   before_action :pending_invitations, if: -> { step == :pending_invitations }
   before_action :clear_session, if: -> { step == :overview }
   before_action :set_responsible_person, only: %i[show update]
+  before_action :responsible_person_details_form, if: -> { step == :enter_details }
```
slawosz, line 12: how about `set_responsible_person_details_form`?

slawosz (approved):

cosmetics-web/app/controllers/responsible_persons/account_wizard_controller.rb
scruti: that was the original name but rubocop doesn't like it as it differs from the memorised variable name :/

nasirkhanbeis (approved):
Tested on review app 1850, looks good as expected. 
User can now create same name RP unless they are associated with same name RP or have invitation already.
However there is a chance of ending up duplicate RP in same user if a user invites to same RP who already created. ( which I think its fine)
Note: there are some accessibility issues around validation message for which I will raise a separate card.

PR #1854, opened 2020-12-07
Allow only one user session

slawosz:
Only one user session will exist. New sign in will invalidate existing session in other browser (if there is such session).


scruti (approved):
LGTM!

PR #1865, opened 2020-12-15
Change phone validations: Allow international phones. Apply validations everywhere.

scruti:
[Jira Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-896)

These changes include:
1. Allow international phone numbers for Submit Users.
2. Only allow UK phone numbers for Search Users.
3. Bring validations to the models, not only on the forms.
4. Align validations with GovUK API phone validations.
5. **Extra**: Refactor locales & form validation errors.

We want to allow Submit users to use international phone numbers.
As we want to avoid allowing users to set a phone number that will be rejected later by GovUK Notify API when trying to send a 2FA SMS, we decided to align our service phone number validations with GovUK Notify ones.

By duplicating the validations in both the forms and model we avoid situations where an user that got its phone number set outside one of the validated forms, ends with an invalid phone that later is rejected by GovUK Notify API (this is currently happening for users who registered through "Complete Registration" page following an invitation)

## Phone validations rules: 
- Phone is identified as UK phone if either:
  - Starts with "0" (but not "00"). 
  - Ignoring starting "0"s and removing special symbols (`+, spaces, -, (, )`):
    - Starts with UK prefix (44).
    - Starts with UK Mobile prefix(7) and its length is less than 11 digits.

- UK Phone numbers must: 
  - Start with UK Mobile prefix ("7") 
  - Be 10 digits long.

- Any phone number not identified as UK number is assumed to be an international phone number.

- International phone numbers must: 
  - Be between 8 and 15 digits length
  - Start with a prefix listed in the yaml file with the list of approved prefixes we introduced in the PR.

## Extra work
During the development of this ticket noticed that in multiple forms we had to specify the scope/route for the validation errors from the translations in the YAML locales files. 
This could be avoided by moving those errors into the correct YAML route/tree so the location can be automatically inferred by Rails and the error declaration becomes more succinct. 
As I was already doing it for the forms with the 2FA mobile phone validations, decided to refactor this on the rest of the forms and keep them consistent with the refactor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 ## Inviting Search Users
 
 1. SSH to rails console (described below)
-2. Run `InviteUser.call name: 'Joe Doe', email: 'email@example.org', role: :poison_centre`
+2. Run `InviteSearchUser.call name: 'Joe Doe', email: 'email@example.org', role: :poison_centre`
 3. Role could be `poison_centre` or `msa`
 
 ## Getting Setup
--- a/cosmetics-web/app/forms/concerns/email_form_validation.rb
+++ b/cosmetics-web/app/forms/concerns/email_form_validation.rb
@@ -6,9 +6,9 @@ module EmailFormValidation
 
     validates :email,
               email: {
-                message: I18n.t(:wrong_format, scope: "sign_in_form.email"),
+                message: I18n.t(:wrong_format, scope: :email_form_validation),
                 if: ->(sign_in_form) { sign_in_form.email.present? },
               }
-    validates_presence_of :email, message: I18n.t(:blank, scope: "sign_in_form.email")
+    validates_presence_of :email, message: I18n.t(:blank, scope: :email_form_validation)
   end
 end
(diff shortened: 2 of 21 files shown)


nasirkhanbeis:
It appears that, for international numbers, it is stripping out + and 00 so I think I'm not receiving messages. 
Eg. when user enters number 0091 or +91 (india code) then it should retain same format in db and when I click change number and see the pre-populated phone number field as 0091 or +91. This applies to all international numbers.

nasirkhanbeis:
spotted one more issue, so in search domain registration journey,
1. enter valid UK number, fill. in other fields and submit form
2. On 2fa security code, click on code not received and change phone number, enter international number +918374303074
3. submit form, it returns 422 error
Expected to return validation error  - Enter valid phone number

scruti:
Thanks for identifying these issues @nasirkhanbeis . Both issues have got addressed and covered with corresponding tests.

nasirkhanbeis:
Hi @scruti  re-tested above issues, looking good as expected. thanks!
All good

nasirkhanbeis (approved):
looks good as expected.

PR #1877, opened 2020-12-24
WIP: Don't process Notification file unless virus free

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1003)
The antivirus job and the notification file processing are both run as async jobs in parallel.

This allows the file to be processed while is still being analyzed for virus or, even worse, when it actually contains a virus file in the zip.

In order to avoid that we guard the process execution behind a check on
the antivirus output:
- If the antivirus process finished and the file is marked as safe,
  process the notification as usual.
- If the antivirus process hasn't run on the file, reschedule the
  processing of the notification to be run again later.
- If the antivirus process finished and detects a virus on the
  attachment, do not process the file.

### DIFF
--- a/cosmetics-web/spec/jobs/notification_file_processor_job_spec.rb
+++ b/cosmetics-web/spec/jobs/notification_file_processor_job_spec.rb
@@ -10,91 +10,175 @@ RSpec.describe NotificationFileProcessorJob, :with_stubbed_antivirus do
   let(:responsible_person) { create(:responsible_person) }
 
   describe "#perform" do
-    before do
-      described_class.new.perform(notification_file.id)
+    subject(:job) { described_class.new }
+
+    context "when the attached file is safe from virus" do
+      let(:notification_file) do
+        create(:notification_file, uploaded_file: uploaded_file)
+      end
+
+      before do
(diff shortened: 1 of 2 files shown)


cosmetics-web/spec/jobs/notification_file_processor_job_spec.rb
```diff
@@ -10,91 +10,175 @@
...
+          begin
+            job.perform(notification_file.id)
+          rescue described_class::AntivirusCheckPendingError
+            nil
+          end
+        }.not_to change(Notification, :count)
+      end
+
+      # Struggling to get this bit tested, as the "retry_on" magic does not seem to be happening on specs
+      xit "enqueues the job again" do
```
scruti, line 172: The only thing stopping from having the PR completely ready is this bit. I'm trying to test that the job gets re-enqueued, but with Rails magic over the `retry_on` it becomes quite hard to assert over it.
I don't want to simply switch the `retry_on` to an instance `retry_job` call because it misses critical functionality, like setting a max attempts, that comes with a default of 5 in the `retry_on`.
I prefer to keep the nice functionality (avoid endless loops of attempting the job) over the testing issue TBH.

PR #1884, opened 2021-01-01
Bugfix: incomplete accounts trying to reset password

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1014)
[Bugtracker exception](https://sentry.io/organizations/beis/issues/2123481455)

When an user with an incomplete account (missing their mobile number, name or password) attempts a password recovery the service is throwing an error.

Sadly seems some users get confused after introducing their email in the first step of the Submit user registration and, instead
of checking their email for the confirmation link (as prompted), they directly attempt to login, and not having set their password they attempt to reset it.

The cause of the exception is the 2FA attempt prior to the password reset, while the user actually doesn't have a mobile number to send the 2FA code to.

As this bit of the code is shared between Submit and Search users, we are changing this behaviour to:

- SUBMIT USERS:
  When a submit user that didn't complete its registration attempts to reset its password (that is not set up), will be resent the 
  account confirmation instructions instead.

- SEARCH USERS:
  When a search user that didn't complete its registration after having received an invitation link attempts to reset its password, 
  will be resent the invitation link instead.

### DIFF
--- a/cosmetics-web/app/controllers/users/passwords_controller.rb
+++ b/cosmetics-web/app/controllers/users/passwords_controller.rb
@@ -65,6 +65,9 @@ module Users
     end
 
     def resend_account_setup_link_for(user)
+      user.confirmed_at = nil
+      user.account_security_completed = false
+      user.save(validate: false)
       user.resend_account_setup_link
       redirect_to check_your_email_path
     end
--- a/cosmetics-web/app/controllers/users/passwords_controller.rb
+++ b/cosmetics-web/app/controllers/users/passwords_controller.rb
@@ -28,6 +28,9 @@ module Users
     end
 
     def create
+      user = user_class.find_by(email: params[user_param_key][:email])
+      return resend_account_setup_link_for(user) if user && !user.has_completed_registration?
+
       super do |resource|
         suppress_email_not_found_error
 
@@ -61,8 +64,8 @@ module Users
       user_signed_in? && is_fully_authenticated?
     end
 
-    def resend_invitation_link_for(user)
-      SendUserInvitationJob.perform_later(user.id, nil)
+    def resend_account_setup_link_for(user)
+      user.resend_account_setup_link
       redirect_to check_your_email_path
     end
 
--- a/cosmetics-web/app/models/search_user.rb
+++ b/cosmetics-web/app/models/search_user.rb
@@ -33,9 +33,8 @@ class SearchUser < User
     !msa_user?
   end
 
-  # TODO: is this used?
-  def send_confirmation_instructions
-    SearchNotifyMailer.send_account_confirmation_email(self).deliver_later
+  def resend_account_setup_link
+    SearchNotifyMailer.invitation_email(self).deliver_later
   end
 
   def send_reset_password_instructions_notification(token)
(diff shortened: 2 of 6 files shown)


cosmetics-web/app/controllers/users/passwords_controller.rb
```diff
@@ -61,8 +64,8 @@ def passed_secondary_authentication?
       user_signed_in? && is_fully_authenticated?
     end
 
-    def resend_invitation_link_for(user)
-      SendUserInvitationJob.perform_later(user.id, nil)
```
scruti, line 65: This class actually didn't exist. We messed up when  bringing code from PSD service :sweat_smile: :see_no_evil:

cosmetics-web/app/models/search_user.rb
```diff
@@ -33,9 +33,8 @@ def can_view_product_ingredients?
     !msa_user?
   end
 
-  # TODO: is this used?
-  def send_confirmation_instructions
-    SearchNotifyMailer.send_account_confirmation_email(self).deliver_later
```
scruti, line 38: Answer: NO.
Not used and.... not defined! Was only defined for `SubmitNotifyMailer`

cosmetics-web/app/controllers/users/passwords_controller.rb
```diff
@@ -61,8 +64,8 @@ def passed_secondary_authentication?
...
+    def resend_account_setup_link_for(user)
+      user.resend_account_setup_link
```
scruti, line 68: defined this method in both `SubmitUser` and `SearchUser` to encapsulate behaviour differences in the respective class.
For SubmitUser, the "account setup link" is the confirmation instructions from Devise.
For SearchUser is the invitation link for completing the registration.

cosmetics-web/spec/factories/user.rb
```diff
@@ -4,14 +4,15 @@
...
     password { "testpassword123" }
-    confirmed_at { 1.hour.ago }
     has_accepted_declaration { true }
     direct_otp_sent_at { Time.zone.now }
     direct_otp { "12345" }
     mobile_number_verified { true }
     account_security_completed { true }
 
     factory :submit_user, class: "SubmitUser" do
+      confirmed_at { 1.hour.ago }
```
scruti, line 14: Moved this because is only relevant/set for Submit Users.

slorek (approved):

bfanshawe (approved):
Yes - this works in both 'Submit' and 'Search' prior to inputting a password during the registration process - the confirmation email (for submit) or the invite (for search) is resent

cosmetics-web/app/controllers/users/passwords_controller.rb
```diff
@@ -65,6 +65,9 @@ def passed_secondary_authentication?
     end
 
     def resend_account_setup_link_for(user)
+      user.confirmed_at = nil
+      user.account_security_completed = false
+      user.save(validate: false)
```
slorek, line 70: This is needed to ensure the user can set up their account again if they failed to validate their phone number the first time.

tobyhughes (approved):

PR #1885, opened 2021-01-02
Fix zip file processing issues with zip files

slorek:
Fixes zip file uploads when the zip file:

* Contains MacOS metadata files
* Contains an XML file inside a folder which has the same name as the XML file

I have added a new fixture file `testExportFileWithMacMetadata.zip` to test these scenarios.


tobyhughes (approved):
Working as expected!

bfanshawe (approved):
all working good

PR #1886, opened 2021-01-02
Send unexpected errors to Sentry when processing file uploads

slorek:
Ensures that Sentry alerts us when there is an unexpected error processing a file upload. This will help us know when there is a problem and more easily diagnose it.


scruti (approved):
Thanks for noticing and adding this :clap:

PR #1896, opened 2021-01-05
Redirect to notification page if already submitted when attempting to access wizard pages

slorek:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1036

## Description
When the user has completed a notification they may subsequently use the browser's history to navigate back to a wizard page for that notification. Rather than throwing an uncaught exception this change redirects the user to the summary page for that notification.

So far we have seen 89 instances of this error in production so this will eliminate future occurrences.

Fixes these issues which can be marked as resolved once deployed:

https://sentry.io/organizations/beis/issues/2123870623/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2124963423/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2124963343/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2124722739/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2126674565/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2126152205/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2125879994/?environment=prod&project=1398436
https://sentry.io/organizations/beis/issues/2125878751/?environment=prod&project=1398436

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
@@ -3,7 +3,7 @@ require "will_paginate/array"
 class ResponsiblePersons::NotificationsController < SubmitApplicationController
   before_action :set_responsible_person
   before_action :validate_responsible_person
-  before_action :set_notification, only: %i[show edit confirm delete destroy]
+  before_action :set_notification, only: %i[show confirm delete destroy]
 
   def index
     @pending_notification_files_count = 0
@@ -39,6 +39,12 @@ class ResponsiblePersons::NotificationsController < SubmitApplicationController
 
   # Check your answers page
   def edit
+    @notification = Notification.find_by reference_number: params[:reference_number]
+
+    return redirect_to responsible_person_notification_path(@notification.responsible_person, @notification) if @notification.notification_complete?
+
+    authorize @notification, policy_class: ResponsiblePersonNotificationPolicy
+
     @previous_page_path = previous_path_before_check_your_answers(@notification)
 
     if params[:submit_failed]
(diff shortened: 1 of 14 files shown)


nasirkhanbeis (approved):
Verified zip file, manual journey back link and browser back. looking good as expected, 
Further browser back with intention to visit notification wizard will only redirect to summary page which is good.

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -39,6 +39,12 @@ def new
 
   # Check your answers page
   def edit
+    @notification = Notification.find_by reference_number: params[:reference_number]
```
slawosz, line 42: Why you don't do use `set_notification` anymore and why this particular order.

slorek: Because `set_notification` is a generic method used for all controller actions. In this case I only want to intercept and redirect for the `edit` action.

slawosz (approved):

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
slawosz: But how about simply adding `return redirect ..` there, why it would not work?

slorek: We could do so if we add a condition e.g. `if request.action == :edit`, but that adds additional complexity and the implementation becomes no longer generic in order to justify being a `before_action` filter. The alternative would be to split `set_notification` up to two filter methods, `set_notification` and `authorise_action` or similar, but I felt this was the most elegant solution and in line with our other implementations in PSD.

PR #1898, opened 2021-01-05
Turn email to case insensitive in db

scruti:
## Relevant links
[Jira ticket](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1894)
[Related Bug issue](https://sentry.io/organizations/beis/issues/2128065615/)
[Related PR with previous attempt to fix](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1894)
## Description of the issue
Users are mistakenly getting registered twice under the same email due to the search for existent users not being case insensitive. (example@example.com and EXAMPLE@example.com were taken as different emails).
This is causing [exceptions in our service](https://sentry.io/organizations/beis/issues/2128065615/) when users follow up on the registration link.
A [first attempt to solve this](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1894) assumed the users were trying to register themselves twice through the registration page.
But the issue kept showing up, and then we realised this is caused by:
- User registers with the service under "example@example.com" 
- Member of Responsible Person team invites the user to the team, but fills the invitation using a capitalised email ("Example@example.com" or similar).
- First user receives the email, attempts to join the team, gets redirected to complete registration instead to be added to the team.
- Service fails when submitting the registration as the email fails validation.

## Solution
The original approach requires developers to remember to, instead of calling `find_by(email: `, replace it with a more complex query each time we want to find if an user exists with the given email in order to achieve a case insensitive comparison.
This is far from ideal and error prone. As the fact that the error kept happening through the invitations after patching the main registration flow shows.

By using CITEXT type, we achieve case insensitive searches directly implemented by PostgreSQL and being transparent to the Rails ActiveRecord calls. So we can keep using `find_by(email:` as usual. And PostgreSQL will match the email texts ignoring the capitalisation.


slorek (approved):

nasirkhanbeis:
Test scenarios:
Validate invite email with different casing to an existing email which has self registered with account security but no rp yet- Should be able to join the invited RP - Failed
Validate invite email with upper case email and domain which has pre-registered by just submitting name and email- should be able to join invited RP successfully - Failed 
Validate invite email with lower case and create account with same email upper case- should be able to continue create account however should see page where they can see which RP has pre-invited and can continue - Passed
Validate  create account with same email but different casing(lower case/upper case and domains) - Should just detect as same email and send same token
Validate invite same email with different casing - should send same invite token 
Validate invite an existing user who has RP setup already - user should be able to join the invited RP

nasirkhanbeis:
1. Validate invite email with different casing to an existing email which has self registered with account security but no rp yet- Should be able to join the invited RP - **Passed**
2. Validate invite email with upper case email and domain which has pre-registered by just submitting name and email- should be able to join invited RP successfully **Passed** 
3. Validate invite email with lower case and create account with same email upper case- should be able to continue create account however should see page where they can see which RP has pre-invited and can continue - Passed
4. Validate create account with same email but different casing(lower case/upper case and domains) - Should just detect as same email and send same token  - **Passed**
5. Validate invite same email with different casing - should send same invite token - **Passed**
6. Validate invite an existing user who has RP setup already - user should be able to join the invited RP - **Passed**

Re-tested above scenarios and did some exploratory testing. looking good. Upper casing and lower casing detects as same email and sends same token. 
Verified invite and register with same email with different casing(lower and upper) and vice versa. Looks good.

Please note, there seems to be some issue with token expiry such as, when token is expired upon resend invite, new token doesn't seem to work. This appears to be a different issue for which I will raise a separate card.

nasirkhanbeis (approved):
looks good! see my earlier comment for the details.

scruti:
Thanks for the thoughtful check @nasirkhanbeis

PR #1899, opened 2021-01-05
Fix database connection pool errors

slorek:
Fixes connection timeout errors due to insufficient connection pool size as seen here: https://sentry.io/organizations/beis/issues/2124164600

### DIFF
--- a/cosmetics-web/manifest.review.yml
+++ b/cosmetics-web/manifest.review.yml
@@ -30,10 +30,14 @@ applications:
     - cosmetics-devise-env
   processes:
     - type: web
+      env:
+        RAILS_MAX_THREADS: ((web-max-threads))
       command: export $(./env/get-env-from-vcap.sh) && STATEMENT_TIMEOUT=60s bin/rake cf:on_first_instance db:migrate && bundle exec puma
       instances: 1
       memory: 1G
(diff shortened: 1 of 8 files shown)


scruti (approved):

cosmetics-web/manifest.review.yml
```diff
@@ -30,10 +30,14 @@ applications:
...
   processes:
     - type: web
+      env:
+        RAILS_MAX_THREADS: ((web-max-threads))
       command: export $(./env/get-env-from-vcap.sh) && STATEMENT_TIMEOUT=60s bin/rake cf:on_first_instance db:migrate && bundle exec puma
       instances: 1
       memory: 1G
     - type: worker
+      env:
+        RAILS_MAX_THREADS: ((worker-max-threads))
```
slawosz, line 40: why not `WORKER_MAX_THREADS`? If there is no obvious reason for those, I would rename this variable.

slawosz (approved):

cosmetics-web/manifest.review.yml
slorek: Because the application requires a generic name for this variable which is used in the `database.yml` for both worker and web instances.

PR #1905, opened 2021-01-06
Restrict Responsible Person postcode to UK

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1023)

It is a requirement that the Responsible Person is established in the UK.

Decided to implement this with a Regular Expression as should be good enough to check UK Postcode format.
Could have gone for gems or consulting third-party services to check for real addresses but would be too complex for our goal.

### DIFF
--- a/cosmetics-web/app/forms/responsible_person_details_form.rb
+++ b/cosmetics-web/app/forms/responsible_person_details_form.rb
@@ -1,6 +1,8 @@
 class ResponsiblePersonDetailsForm
   include ActiveModel::Model
   include ActiveModel::Attributes
+  include ActiveModel::Validations::Callbacks
+  extend AutoStripAttributes
 
   attribute :name
   attribute :address_line_1
@@ -9,6 +11,8 @@ class ResponsiblePersonDetailsForm
   attribute :county
   attribute :postal_code
   attribute :user
+
+  auto_strip_attributes :name, :address_line_1, :address_line_2, :city, :county, :postal_code, virtual: true
 
   validates :name, presence: true
   validates :address_line_1, presence: true
--- a/cosmetics-web/app/forms/responsible_person_details_form.rb
+++ b/cosmetics-web/app/forms/responsible_person_details_form.rb
@@ -14,6 +14,7 @@ class ResponsiblePersonDetailsForm
   validates :address_line_1, presence: true
   validates :city, presence: true
   validates :postal_code, presence: true
+  validates :postal_code, uk_postcode: true, if: -> { postal_code.present? }
   validate :user_not_member_of_rp_with_same_name
   validate :user_not_invited_to_rp_with_same_name
 
--- a/cosmetics-web/app/forms/responsible_person_details_form.rb
+++ b/cosmetics-web/app/forms/responsible_person_details_form.rb
@@ -1,8 +1,7 @@
 class ResponsiblePersonDetailsForm
   include ActiveModel::Model
   include ActiveModel::Attributes
-  include ActiveModel::Validations::Callbacks
-  extend AutoStripAttributes
+  include StripWhitespace
 
   attribute :name
   attribute :address_line_1
@@ -11,8 +10,6 @@ class ResponsiblePersonDetailsForm
   attribute :county
   attribute :postal_code
   attribute :user
-
-  auto_strip_attributes :name, :address_line_1, :address_line_2, :city, :county, :postal_code, virtual: true
 
   validates :name, presence: true
   validates :address_line_1, presence: true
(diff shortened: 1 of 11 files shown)


slorek (approved):

cosmetics-web/app/validators/uk_postcode_validator.rb
```diff
@@ -0,0 +1,17 @@
+class UkPostcodeValidator < ActiveModel::EachValidator
+  # We start by making sure our postcode isn't surrounded by anything else in our field,
+  # so we encase it between a beginning (\A) and an end (\z) of string.
+  # Then, we accept one or two letters [a-zA-Z]{1,2}
+  # Followed by either one digit and a letter, or between one and two digits: ([0-9]{1,2}|[0-9][a-zA-Z])
+  # An optional space \s*
+  # Finally, a digit followed by two letters: [0-9][a-zA-Z]{2}
+  # Source: https://coderwall.com/p/rnucjg/regexp-how-to-validate-a-uk-postcode
+  UK_POSTCODE_REGEX = /\A[a-zA-Z]{1,2}([0-9]{1,2}|[0-9][a-zA-Z])\s*[0-9][a-zA-Z]{2}\z/.freeze
+  ERROR_MSG = "Enter a UK postcode".freeze
```
slawosz, line 10: You could move this to internatiolization file.

cosmetics-web/app/models/responsible_person.rb
```diff
@@ -10,12 +10,17 @@ class ResponsiblePerson < ApplicationRecord
 
   enum account_type: { business: "business", individual: "individual" }
 
+  auto_strip_attributes :name, :address_line_1, :address_line_2, :city, :county, :postal_code
```
slawosz, line 13: I see this is used twice - in model in form. Could we avoid duplication?

cosmetics-web/app/forms/responsible_person_details_form.rb
```diff
@@ -10,10 +12,13 @@ class ResponsiblePersonDetailsForm
   attribute :postal_code
   attribute :user
 
+  auto_strip_attributes :name, :address_line_1, :address_line_2, :city, :county, :postal_code, virtual: true
```
slorek, line 15: Suggestion: this gem is very simple, and we had a similar behaviour in PSD: https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/models/correspondence.rb#L6. It might not be obvious to someone that this is happening due to non-standard declaration. I would prefer using a standard `before_validation` hook (which is what this gem does) to make it more obvious and reduce unnecessary external dependencies.

slawosz (approved):

cosmetics-web/app/validators/uk_postcode_validator.rb
scruti: I considered it but ended not doing so for keeping all the logic & data regarding "UK Postcode Validator" in the same class.

If different form/models using this validator had various error texts, I would have defined them in the int. file and pass them in the calls for validation. But as we're using the "default", I valued the class's cohesion defining the "default" as a constant.

nasirkhanbeis:
Re-tested this last week after new changes, looks good as expected. leading and trailing spaces are stripped out before it is saved in backend.

nasirkhanbeis (approved):
all good as expected!

cosmetics-web/app/models/responsible_person.rb
scruti: Ended removing duplication through a concern relying on its default behaviour: https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1905/commits/471af1ae3af60bce6c85eb8b873c9cc22b82dee3

Included both in model and form object as we want the whitespace striping happening even if the model is not instantiated from the form.

cosmetics-web/app/forms/responsible_person_details_form.rb
scruti: Moved it to a model concern. Thanks for the suggestion :+1:  https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1905/commits/471af1ae3af60bce6c85eb8b873c9cc22b82dee3

cosmetics-web/spec/models/concerns/strip_whitespace_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "rails_helper"
+require_relative "../../../app/models/responsible_person.rb"
```
scruti, line 2: Didn't find the way to do a specific class agnostic test while requiring an ActiveRecord class as we do. So ended picking up a particular class to test the behaviour of the concern over ActiveRecord classes.

PR #1909, opened 2021-01-07
Wait for virus check before processing notification file, and recover from S3 errors

slorek:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1003

## Description
Fixes the following issues:

* `Aws::S3::Errors::NotFound` exceptions not caught when attempting to analyse an uploaded file (most likely due to a request timeout). Now an error message will appear on the notification files list to inform the user of a problem and to try again. There were 78 instances of this error on production since go live, though the majority will be due to retries.
* Notification file processing could be started before the anti virus check has been completed. This is due to a race condition caused by the controller scheduling the job at the same time as the file is uploaded. Since both jobs are asynchronous, there's no guarantee that the analyser job is completed first. Now the analyser job will schedule the processing job only once it is completed.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/app/jobs/notification_file_processor_job.rb
```diff
@@ -11,6 +11,9 @@ class UnexpectedStaticFilesError < FileUploadError; end
 
   def perform(notification_file_id)
     @notification_file = NotificationFile.find(notification_file_id)
+
+    return if @notification_file.upload_error.present?
```
slorek, line 15: This is delayed to this point in order to guard against the job somehow being scheduled on an erroneous file.

cosmetics-web/spec/support/analyzers.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.shared_context "without default file analyzers", shared_context: :metadata do
```
slorek, line 1: This is pretty dirty but I can't think of a better pragmatic solution currently. We can improve this later if anyone has an idea.

cosmetics-web/spec/requests/notification_files_spec.rb
```diff
@@ -19,7 +19,8 @@
         create(:notification_file,
                uploaded_file: create_file_blob("testExportFile.zip"),
                responsible_person: responsible_person,
-               user: user)
+               user: user,
+               upload_error: "unknown_error")
```
slorek, line 23: Because the `MasterAnalyzer` is invoked automatically and schedules the `NotificationFileProcessorJob`, any `NotificationFile` created in tests will be deleted by the job automatically unless there are errors. This more realistically mirrors real world behaviour.

cosmetics-web/spec/features/zip_file_upload_notifications_spec.rb
```diff
@@ -458,4 +458,24 @@
       expect_to_see_notification_error("The uploaded file has been flagged as a virus")
     end
   end
+
+  feature "handling file uplaod errors", :with_stubbed_s3_returning_not_found do
```
scruti, line 462:
```suggestion
  feature "handling file upload errors", :with_stubbed_s3_returning_not_found do
```

scruti (approved):
Feels weird to have the file processing being triggered from a method meant to extract metadata.
But it works, and I like that sets allow us not to have to deal with re-scheduling of the jobs while setting a clear execution order dependency.  
So happy to go with this  ¯\_(ツ)_/¯

About the testing... I wonder if we would be able to skip the analyzer calls from the factory creation. But is something to explore once this is already patched.

:shipit: !

nasirkhanbeis (approved):
looks good as expected!

PR #1920, opened 2021-01-11
Fix back link from new RP type selection

scruti:
[Sentry related error](https://sentry.io/organizations/beis/issues/2052277308)
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1047)
When a user belonging to multiple RP reached this page before having selected their current RP, the backlink route was throwing an exception.

Fixed by pointing the backlink to the RP selection when the user has multiple RPs, none of them is defined as their current one.


nasirkhanbeis (approved):
Tested this on review app 1920, looking good as expected.
Please note, back link nav issues still persist for which we have separate ticket anyway.

PR #1923, opened 2021-01-12
Direct zip upload with worker to process uploaded files

slawosz:
Direct upload for zip files. Uses JS `ActiveStorage` library to send files directly to S3.


nasirkhanbeis:
Few issues spotted:
1. when I upload eg. 100 files, checking files counter still show as 1 - Expected to show correct count 100 files
2. when I upload 100 files, page becomes very slow like frozen. - Need show some spinner or disable the continue button 
image

cosmetics-web/app/services/direct_upload_handler.rb
```diff
@@ -0,0 +1,49 @@
...
+      @present_names << name
+
+      notification_file_ids << notification_file.id
+    end
+
+    handle_missing_files
+
+    notification_file_ids.each do |id|
+      NotificationFileProcessorJob.perform_later(id)
+    end
```
slorek, line 36: Please change: there [was a change](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1909) to the way this was scheduled to prevent race conditions. This is now automatically scheduled by the `MasterAnalyzer` when you call `attach` (see https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/blob/master/cosmetics-web/app/analyzers/master_analyzer.rb#L24). Please remove this because it will result in the job running twice and causing either a race condition or an error.

cosmetics-web/app/services/direct_upload_handler.rb
```diff
@@ -0,0 +1,49 @@
...
+        next
+      end
+
+      name = uploaded_file.filename.base
+      notification_file = NotificationFile.new(
+        name: name,
+        responsible_person: @responsible_person,
+        user: @submit_user,
+      )
+      notification_file.uploaded_file.attach(uploaded_file)
```
slorek, line 25: Suggestion: Delaying the `NotificationFile` record creation could result in the user seeing no uploaded files until the job was run. In normal conditions I suppose this shouldn't happen, but if there was a delay in the job running it could be confusing.

I think this `attach` call is the expensive one. To avoid the problem we could instead create the `NotificationFile` record immediately [here in the controller](https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1923/files#diff-e036390f6129d43999b8fa84fcf4a382a627cd0fffd1ab72b3956d4b185449bdR18) and then schedule the attachment of the file as the asynchronous job.

slorek (changes requested):

cosmetics-web/app/services/direct_upload_handler.rb
slawosz: Attach is not particulary expensive, its the loop which makes it expensive.

slawosz: User does not see notification files as such, so I don't think its an issue.

nasirkhanbeis (approved):
Tested this on review app, looking good as expected, 
Verified this new approach with 100 files upload at a time, feedbacks user appropriately if there's any error with any files however processes remainder files as expected. 
Also verified all other possible zip file upload scenarios such as duplicate file, non zip file, zip file with no product xmls, zip file with virus file . all returns back with expected errors. 
happy to approve this.

cosmetics-web/app/controllers/responsible_persons/notification_files_controller.rb
```diff
@@ -25,41 +45,25 @@ def create
...
-
-  def destroy
-    NotificationFile.destroy(params[:id])
-    redirect_to responsible_person_notifications_path(@responsible_person)
-  end
-
-  def destroy_all
-    @responsible_person.notification_files.where(user_id: current_user.id).where.not(upload_error: nil).destroy_all
-    redirect_to responsible_person_notifications_path(@responsible_person)
+      NotificationFileProcessorJob.perform_later(notification_file.id)
```
slorek, line 63: See previous comment. I think the `MasterAnalyzer` will schedule this job automatically, so this will result in it being run twice.

cosmetics-web/app/jobs/direct_upload_handler_job.rb
```diff
@@ -0,0 +1,5 @@
+class DirectUploadHandlerJob < ApplicationJob
+  def perform(uploaded_files, uploaded_file_names, responsible_person_id, current_user_id)
+    DirectUploadHandler.new(uploaded_files, uploaded_file_names, responsible_person_id, current_user_id).call
+  end
+end
```
slorek, line 5: ✂️

scruti, line 5: Since the job is not being called... could we remove it?

slawosz: I would keep it. If we will have some issues its just a matter of `git reverse` to start using it.

slorek (changes requested):
I don’t think you need to schedule `NotificationFileProcessorJob` manually. The `MasterAnalyzer` will do it.

Also I think we should add test coverage on the controller.

cosmetics-web/app/jobs/direct_upload_handler_job.rb
slawosz: As I mentioned, I left this as possibility when processing in request would take too much time.

scruti (approved):
image

Suggested to remove the uncalled job class.

cosmetics-web/app/jobs/direct_upload_handler_job.rb
scruti: Can we actually revert a squashed commit? (we squash on merge)
Even in that case... wouldn't a `git revert` re-introduce the class again if there is a commit removing it?

cosmetics-web/spec/requests/responsible_persons/notification_files_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+          uploaded_files_names: file_names,
+        }
+      end
+
+      it "processes upload using DirectUploadHandler" do
+        handler = double
+        expect(DirectUploadHandler).to receive(:new).with(file_ids, file_names, responsible_person.id, submit_user.id) { handler }
+        expect(handler).to receive(:call)
+
+        post "/responsible_persons/#{responsible_person.id}/notification_files", params: params
```
scruti, line 34: Suggestion:
About this:
```
RSpec/MessageSpies:
  Exclude:
    - spec/requests/responsible_persons/notification_files_spec.rb 
 ```
 This is the `to_receive` vs `have_received` rule on the spies right?
 Wouldn't this work without having to bypass the rule?
 
```suggestion
        handler = double
        allow(DirectUploadHandler).to receive(:new).with(file_ids, file_names, responsible_person.id, submit_user.id) { handler }
   
        post "/responsible_persons/#{responsible_person.id}/notification_files", params: params
        
        expect(handler).to have_received(:call)
```

cosmetics-web/.rubocop.yml
```diff
@@ -78,6 +78,10 @@ RSpec/LetSetup:
     - spec/features/account/change_name_spec.rb
     - spec/features/account/change_password_spec.rb
 
+RSpec/MessageSpies:
+  Exclude:
+    - spec/requests/responsible_persons/notification_files_spec.rb
+
```
scruti, line 84:
```suggestion
```

slorek (approved):

cosmetics-web/spec/requests/responsible_persons/notification_files_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        {
+          uploaded_files: file_ids,
+          uploaded_files_names: file_names,
+        }
+      end
+
+      it "processes upload using DirectUploadHandler" do
+        handler = instance_double(DirectUploadHandler, call: true)
+        allow(DirectUploadHandler).to receive(:new).with(file_ids, file_names, responsible_person.id, submit_user.id) { handler }
+
```
scruti, line 32: So sorry about this :/
```suggestion

```

cosmetics-web/spec/requests/responsible_persons/notification_files_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          uploaded_files_names: file_names,
+        }
+      end
+
+      it "processes upload using DirectUploadHandler" do
+        handler = instance_double(DirectUploadHandler, call: true)
+        allow(DirectUploadHandler).to receive(:new).with(file_ids, file_names, responsible_person.id, submit_user.id) { handler }
+   
+        post "/responsible_persons/#{responsible_person.id}/notification_files", params: params
+
```
scruti, line 34: So sorry about this :/
```suggestion

```

PR #1927, opened 2021-01-13
Don't allow users to have the option to select how they want to give a cosmetic product formulation

KyleMacPherson:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-987

## Description
After this change:
- users will not have the option to select how they want to give a cosmetic product formulation if they have notified before Jan 1st 2021 - they will go straight to the choose frame formulation page.
- Show poisonous ingredients file in check your answers page if user has uploaded this file.
- Make poisonous ingredients upload file non-mandatory if user notified after jan 1st.
- Link to upload poisonous ingredients file if user selects `yes` to poisonous ingredients but does not upload the file and has notified after jan 1st.

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
@@ -292,7 +292,10 @@ private
   def render_upload_formulation
     formulation_file = params.dig(:component, :formulation_file)
 
-    if formulation_file.present?
+    if @component.notification.was_notified_before_eu_exit?
+      @component.formulation_file.attach(formulation_file) if formulation_file.present?
+      redirect_to edit_responsible_person_notification_path(@component.notification.responsible_person, @component.notification)
+    elsif formulation_file.present?
       @component.formulation_file.attach(formulation_file)
       if @component.valid?
         redirect_to finish_wizard_path
@@ -300,9 +303,6 @@ private
         @component.formulation_file.purge if @component.formulation_file.attached?
         render step
       end
-    elsif @component.notification.was_notified_before_eu_exit?
-      @component.formulation_file.attach(formulation_file) if formulation_file.present?
-      redirect_to finish_wizard_path
     else
       @component.errors.add :formulation_file, "Upload a list of ingredients"
       render step
--- a/cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
@@ -37,6 +37,11 @@ class ResponsiblePersons::Wizard::ComponentBuildController < SubmitApplicationCo
       create_required_cmrs
     when :list_nanomaterials
       setup_nano_elements
+    when :select_formulation_type
+      if @component.notification.was_notified_before_eu_exit?
+        @component.update(notification_type: "predefined")
+        jump_to(:select_frame_formulation)
+      end
     end
     render_wizard
   end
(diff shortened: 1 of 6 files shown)


cosmetics-web/spec/features/manual_pre_brexit_notifications_spec.rb
```diff
@@ -7,144 +7,6 @@
     sign_in_as_member_of_responsible_person(responsible_person)
   end
 
-  scenario "Manual, pre-Brexit, exact ingredients, single item, no nanomaterials", :with_stubbed_antivirus do
```
slawosz, line 10: Why you removed some of those tests?

KyleMacPherson: I removed them because now we cannot have `manual, pre-brexit and with exact ingredients` because we are not allowing them to put in exact ingredients if they choose pre-brexit.

slawosz (approved):

cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
```diff
@@ -282,7 +282,7 @@ def update_contains_poisonous_ingredients
     end
 
     @component.update!(contains_poisonous_ingredients: params[:component][:contains_poisonous_ingredients])
-    if @component.contains_poisonous_ingredients?
+    if @component.contains_poisonous_ingredients? && !@component.notification.was_notified_before_eu_exit?
```
scruti, line 285: Suggestion:
There is a method for the `!was_notified_before_eu_exit?`, seems clearer to read:
```suggestion
    if @component.contains_poisonous_ingredients? && @component.notification.was_notified_after_eu_exit?
```

scruti: :thinking:  Did we even show the poisonous ingredients questions for "before EU Exit" case?

cosmetics-web/app/helpers/component_build_helper.rb
```diff
@@ -16,6 +16,8 @@ def previous_wizard_path
       wizard_path(:select_category, category: Component.get_parent_category(@category))
     elsif step == :upload_formulation && @component.predefined?
       wizard_path(:contains_poisonous_ingredients)
+    elsif step == :select_frame_formulation && @component.notification.was_notified_before_eu_exit?
+      wizard_path(:select_category)
```
scruti, line 20: If you get feedback from the review app asking you to redirect to the sub-category, you can disregard it as this is already being implemented in my PR fixing the back-links.

cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
KyleMacPherson: No, but we introduced the poisonous ingredients question in this PR but want to skip the file upload, only allowing the file to be uploaded on the check your answers page. Not 100% sure on the reasoning for this but was part of the flow change requirements bruce which you can see in the comments on the linked jira ticket (there are a lot!)

cosmetics-web/app/helpers/component_build_helper.rb
KyleMacPherson: Ah thanks @scruti

nasirkhanbeis:
@KyleMacPherson  just spotted a bug , back link on poisonous ingredient upload page to return to check your answer page when I visited this page via check your answer page. 
image

cosmetics-web/app/controllers/responsible_persons/wizard/component_build_controller.rb
scruti: thanks for clarifying it. Missed those comments.

scruti (approved):

PR #1937, opened 2021-01-18
Validate phone numbers on contact details page

KyleMacPherson:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1026

## Description
This PR adds validation to the phone number field on the contact details page.

I have made changes to the `phone_validator` class because we need to allow british landlines in this case, not strictly mobile phones (which is necessary validation for 2fa).

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


scruti (approved):
Left a comment in case we want to be pickier about which landline prefixes do we accept. But this seems good enough to restrict the contact details info to `0x` UK phone numbers while allowing both landlines and mobile.

LGTM!

PR #1946, opened 2021-01-21
Introduce stricter email format validation

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1061)

We had multiple occurrences of clearly wrong emails being accepted by our validations and failing when Notify service attempted to use the address.

The validate_email gem was severely deprecated, and its validation was so relaxed that was accepting emails like
`user @example.co,m`, `user @ example . com`...
The gem's only value was to define a custom email validator that checked that the email address could be parsed and that the email address domain is not local (`user@example`).

The approach taken was to remove the gem, create our own custom email validator reproducing the gem behaviour, and extend the validator to check the email against accepted email format REGEXP.

To validate the email format we are using `URI::MailTo::EMAIL_REGEXP`, that is standard in Ruby library and strict enough to catch the majority of typos users may be introducing.
The downside with this regexp is that accepts local domains, as they technically are valid email addresses.

But, as we combine both the REGEXP and the domain check, we have both fronts covered.

The Ruby regexp has also been set as the email format regexp used by Devise. As Devise default one is way too relaxed and allows multiple typos.

There is an [open issue on Devise](https://github.com/heartcombo/devise/issues/3820) pushing for using this regexp as default as the current default.

### DIFF
--- a/cosmetics-web/app/models/contact_person.rb
+++ b/cosmetics-web/app/models/contact_person.rb
@@ -3,12 +3,9 @@ class ContactPerson < ApplicationRecord
 
   validates :name, presence: true
   validates :email_address,
-            email: {
-              message: I18n.t(:wrong_format, scope: "contact_person.email_address"),
-            }
-  validates :email_address, presence: { message: I18n.t(:blank, scope: "contact_person.email_address") }
-  validates :phone_number, presence: true
+            presence: true,
+            email: { message: :wrong_format, if: -> { email_address.present? } }
   validates :phone_number,
-            phone: { message: I18n.t(:invalid, scope: "contact_person.phone_number"), allow_landline: true, allow_international: true },
-            if: -> { phone_number.present? }
+            presence: true,
+            phone: { message: :invalid, allow_landline: true, allow_international: true, if: -> { phone_number.present? } }
 end
--- a/cosmetics-web/app/validators/email_validator.rb
+++ b/cosmetics-web/app/validators/email_validator.rb
@@ -1,0 +1,19 @@
+class EmailValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    mail = Mail::Address.new(value)
+
+    unless email_accepted?(mail, value)
+      record.errors.add(attribute, options[:message])
+    end
+  rescue Mail::Field::ParseError
+    record.errors.add(attribute, options[:message])
+  end
+
+private
+
+  def email_accepted?(parsed_email, value)
+    parsed_email.address == value && # Parsed address corresponds to introduced value
+      parsed_email.address =~ URI::MailTo::EMAIL_REGEXP && # Valid email format
+      parsed_email.domain.split(".").length > 1 # Exclude local domains (eg: user@example)
+  end
+end
--- a/cosmetics-web/app/validators/email_validator.rb
+++ b/cosmetics-web/app/validators/email_validator.rb
@@ -13,7 +13,8 @@ private
 
   def email_accepted?(parsed_email, value)
     parsed_email.address == value && # Parsed address corresponds to introduced value
-      parsed_email.address =~ URI::MailTo::EMAIL_REGEXP && # Valid email format
-      parsed_email.domain.split(".").length > 1 # Exclude local domains (eg: user@example)
+      URI::MailTo::EMAIL_REGEXP.match?(value) && # Valid email format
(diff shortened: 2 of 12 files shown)


cosmetics-web/config/locales/en.yml
```diff
@@ -89,7 +89,6 @@ en:
               wrong_format: "Enter email address in the correct format, like name@example.com"
               blank: "Enter email address"
               this_team: "This email address already belongs to member of this team"
-              other_team: "You can not invite this email address to join your team"
```
scruti, line 92: This is not used as now we allow users to belong to multiple RP teams.

cosmetics-web/spec/models/contact_person_spec.rb
```diff
@@ -20,7 +20,7 @@
   it "fails if an email address is not specified" do
     contact_person.email_address = nil
     expect(contact_person.save).to be false
-    expect(contact_person.errors[:email_address]).to include("Enter your email address in the correct format, like name@example.com")
+    expect(contact_person.errors[:email_address]).to include("Enter your email address")
```
scruti, line 23: The format validation for contact person was overshadowing the presence check. We usually apply the format validation only if the presence validation passed. So swap this to be consistent with the rest of validations.

cosmetics-web/app/models/contact_person.rb
```diff
@@ -3,12 +3,9 @@ class ContactPerson < ApplicationRecord
 
   validates :name, presence: true
   validates :email_address,
-            email: {
-              message: I18n.t(:wrong_format, scope: "contact_person.email_address"),
-            }
-  validates :email_address, presence: { message: I18n.t(:blank, scope: "contact_person.email_address") }
-  validates :phone_number, presence: true
+            presence: true,
+            email: { message: :wrong_format, if: -> { email_address.present? } }
```
scruti, line 7: The format validation for contact person was overshadowing the presence check. We usually apply the format validation only if the presence validation passed. So swap this to be consistent with the rest of validations.

nasirkhanbeis:
hey @scruti  can yo also add validation for emails with trailing numbers ? such as nasirkhan.beis@gmail.com1122

nasirkhanbeis (approved):
Tested email registration, invite and contact person email field, looking good. Tried all possible invalid email with spaces and non email format, all being checked as expected. 
Also tested in backend through JMeter script looking good.

KyleMacPherson (approved):

cosmetics-web/app/validators/email_validator.rb
```diff
@@ -0,0 +1,20 @@
...
+    record.errors.add(attribute, options[:message])
+  end
+
+private
+
+  def email_accepted?(parsed_email, value)
+    parsed_email.address == value && # Parsed address corresponds to introduced value
+      URI::MailTo::EMAIL_REGEXP.match?(value) && # Valid email format
+      parsed_email.domain.split(".").length > 1 && # Exclude local domains (eg: user@example)
+      /[a-zA-Z]/.match?(value[-1]) # Last character of the Top Level Domain is a letter
```
KyleMacPherson, line 18: This is a weird one, is this a requirement?

cosmetics-web/config/initializers/devise.rb
```diff
@@ -191,7 +191,8 @@
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
+  # config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
```
KyleMacPherson, line 194: Do we need this comment>

cosmetics-web/app/validators/email_validator.rb
scruti: Nasir asked for it here in the PR: https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/1946#issuecomment-765382995

cosmetics-web/config/initializers/devise.rb
scruti: Good point. I was originally removing it, but given the comment on top making a clear reference about what does this regex does, this seems to be a default value provided by Devise (like the rails initialization files) to be commented/uncommented.
Preferred to leave it commented so we can  compare what Devise provides out of the box against what we use instead.

PR #1951, opened 2021-01-22
Deleting notification

slawosz:
Hard deletes notification and creates log record when notification was deleted.


cosmetics-web/app/models/notification.rb
```diff
@@ -170,6 +174,18 @@ def via_zip_file?
     notification_file_imported? || cpnp_reference
   end
 
+  def destroy_notification!(submit_user)
+    if notification_complete?
+      NotificationDeleteService.new(self, submit_user).call
+    else
+      destroy!
+    end
```
scruti, line 182: Nice

cosmetics-web/app/services/notification_delete_service.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    ActiveRecord::Base.transaction do
+      NotificationDeleteLog.create! do |n|
+        n.notification_product_name = @notification.product_name
+        n.submit_user = @submit_user
+        n.responsible_person_id = @notification.responsible_person_id
+        n.notification_created_at = @notification.created_at
+        n.notification_updated_at = @notification.updated_at
+        n.cpnp_reference = @notification.cpnp_reference
+      end
```
scruti, line 20: Question: Curious about why this is using a block instead of the `key: value` arguments. 
The block seems useful for overriding things like the primary key/id after initialisation, but this doesn't seem to be the case. Am I missing something?

cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -193,4 +193,35 @@
...
+
+      it "uses NotificationDeleteService" do
+        allow(NotificationDeleteService).to receive(:new) { service }
+        allow(service).to receive(:call)
+
+        notification.destroy_notification!(submit_user)
+
+        expect(NotificationDeleteService).to have_received(:new).with(notification, submit_user) { service }
+        expect(service).to have_received(:call)
+      end
```
scruti, line 225: Suggestion (not a request, I'm fine with this too): 
You can save a couple of lines by stubbing the `call method` in the double object definition and by restricting the return of the stub to when `.new` is called with the proper args. 
In this way the `expect(service).to have_received(:call)` will be enough to test both assertions.
```suggestion
      let(:service) { instance_double(NotificationDeleteService, call: nil) }

      it "uses NotificationDeleteService" do
        allow(NotificationDeleteService).to receive(:new).with(notification, submit_user) { service }

        notification.destroy_notification!(submit_user)

        expect(service).to have_received(:call)
      end
```

cosmetics-web/spec/services/notification_delete_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+    service.call
+
+    log = NotificationDeleteLog.first
+
+    expect(log.submit_user).to eq submit_user
+    expect(log.notification_product_name).to eq notification.product_name
+    expect(log.responsible_person).to eq notification.responsible_person
+    expect(log.notification_created_at).to eq notification.created_at
+    expect(log.notification_updated_at).to eq notification.updated_at
+    expect(log.cpnp_reference).to eq notification.cpnp_reference
```
scruti, line 40: RSpec provides a nice helper for avoiding the multiple expectations violation in cases like this:
```suggestion
    expect(log).to have_attributes(
      submit_user: submit_user,
      notification_product_name: notification.product_name,
      responsible_person: notification.responsible_person,
      notification_created_at: notification.created_at,
      notification_updated_at: notification.updated_at,
      cpnp_reference: notification.cpnp_reference,
    )
```

cosmetics-web/spec/services/notification_delete_service_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+      service.call
+
+      log = NotificationDeleteLog.first
+
+      expect(log.submit_user).to eq nil
+      expect(log.notification_product_name).to eq notification.product_name
+      expect(log.responsible_person).to eq notification.responsible_person
+      expect(log.notification_created_at).to eq notification.created_at
+      expect(log.notification_updated_at).to eq notification.updated_at
+      expect(log.cpnp_reference).to eq notification.cpnp_reference
```
scruti, line 56:
```suggestion
      expect(log).to have_attributes(
        submit_user: nil,
        notification_product_name: notification.product_name,
        responsible_person: notification.responsible_person,
        notification_created_at: notification.created_at,
        notification_updated_at: notification.updated_at,
        cpnp_reference: notification.cpnp_reference,
      )
```

cosmetics-web/.rubocop.yml
```diff
@@ -82,6 +83,7 @@ RSpec/MultipleExpectations:
   Max: 2
   Exclude:
     - 'spec/features/**/*'
+    - 'spec/services/notification_delete_service_spec.rb'
```
scruti, line 86: Following the suggestion in the specs to avoid this RSpec alert.
```suggestion
```

cosmetics-web/.rubocop.yml
```diff
@@ -59,6 +59,7 @@ ExampleLength:
   Exclude:
     - 'spec/system/**/*'
     - 'spec/features/**/*'
+    - 'spec/services/notification_delete_service_spec.rb'
```
scruti, line 62: Suggestion: 
I would personally disable the rules per/test using `# rubocop:disable RSpec/ExampleLength` over the offending specs than exclude the full file.

scruti (approved):
Nice! Added some suggestions (the main one being using the `have_attributes` RSpec expectation as it saves us from having to disable some Rubocop rule), but nothing "blocking" :+1:

cosmetics-web/app/services/notification_delete_service.rb
slawosz: Its just for readibility.

nasirkhanbeis (approved):
Tested on review app 1951, looking good as expected. 
Would be nice to add 2fa before rendering delete confirmation page.

PR #1985, opened 2021-02-08
1075 invite users

slawosz:
Simple code to automate inviting `SearchUser`.


cosmetics-web/app/services/one_off/name_extractor.rb
```diff
@@ -0,0 +1,26 @@
+module OneOff
+  class NameExtractor
```
scruti, line 2: **Suggestion** (not a requirement at all)
Given this is only called (as far as I know) from the `SearchUserBulkInviter` and given what this class is doing, I think that instead of being a service, this class would nicely fit better as an OOP class `Email` that could be embedded inside the `SearchUserBulkInviter`. So you could have an object with an interface like:
- `email = Email.new(email_address)`
- `email.name` -> Extracted name from the address 
- `email.email` -> Parsed email (only set if valid).
- `email.valid?` -> Would encapsulate the logic that currently lives in `SearchBulkInviter` (`email.name.nil? || email.email.nil?`)


In case of keeping this as a service, I would suggest `EmailParser` / `EmailNameParser` / `EmailNameExtractor` as the class output is both "name" and "email" after parsing.

scruti (approved):
LGTM. Just left a suggestion about the NameExtractor class.

PR #1998, opened 2021-02-18
Add 2fa authenticator app setup to user registration

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1079)

## Description
This Pull Request adds the possibility of setting up secondary authentication through an Authenticator APP (to generate TOTP codes) on top of the current approach of sending a direct OTP code through SMS to a mobile number.

Affected pages:
- For Submit service: Account Security page.
- For Search service: Complete Registration page.

**This Pull Request only covers secondary authentication setup during user registration**. The possibility of select & fill secondary authentication through Authenticator App code when secondary authentication is required will be added in a further Pull Request.

Issues with SMS secondary authentication:
- **Reliability:** Multiple users with international mobile numbers have reported issues with the SMS code reception. For UK numbers the Notify service is working very well, but with international numbers seems a hit&miss as some carriers are reporting to Notify that the SMS was delivered, while users don't receive it.
- **Security:** While secondary authentication through SMS is better than no secondary authentication at all. SMS authentication is vulnerable to some attacks. Authenticator APP is a more secure alternative.

We aim to enable users to use either or both methods indistinctly. So users will be able to do 2FA with their authentication app, SMS or both.
Users will need to set up at least one of the methods during their registration.

## Screenshots 
**The new interface for the account security page landing looks like this:**

____

**User must select at least one of the secondary authentication methods:**

____

**When selecting the authentication app, the user needs to set it up and fill in the generated code. Setup can be done by scanning a QR Code or by introducing the secret key in the app**

____

**If both methods are selected, both methods must be set in order to be able to complete the form submission:**


cosmetics-web/app/models/concerns/encryptable.rb
```diff
@@ -0,0 +1,71 @@
+# Allows models to define attributes that will be encrypted.
+#
+# Usage EG: "attr_encrypted :secret" will encrypt values given to
+# 'secret = "value"' into 'encrypted_secret'.
+#
+# Ideally the class will be an ActiveRecord class containing 'encrypted_secret'
+# (not 'secret') as table field.
+# Encryptable will define '#secret' and '#secret=' instance methods, so the
+# value can be set to and retrieved from the encrypted db field.
+module Encryptable
```
slawosz, line 10: Did you created this class yourself or this is from some other place?

slawosz (approved):

cosmetics-web/app/models/concerns/encryptable.rb
scruti: Created it in-house combining the ideas of a couple of blogposts on the subject.
Main sources of this approach:
- [Got the metaprogramming bit from herere](https://pawelurbanek.com/rails-secure-encrypt-decrypt)
- [Individual salt per user instead of global salt from here](https://dev.to/shobhitic/simple-string-encryption-in-rails-36pi)

nasirkhanbeis (approved):
looking good as expected.

PR #2012, opened 2021-03-03
Fix: Error when Trigger Question has empty elements

scruti:
Fixes the second issue raised on [Jira ticket 1095](https://regulatorydelivery.atlassian.net/browse/COSBETA-1095)
Some CPNP export files contain Trigger Questions with empty elements (no name, no value).
The file import works ok, but then, when trying to display them, the service attempts to format the answers/values and blows up.


nasirkhanbeis (approved):
tested the given zip file, working fine as expected.

slawosz (approved):
Just double check `filter_map`

scruti:
Check done, thanks @slawosz for pointing it.

PR #2017, opened 2021-03-05
CSV download

slawosz:
Basic implementation on CSV download. Might be slow for RP's which lot of notifications, but this is something we can easily monitor on the logs.

### DIFF
--- a/cosmetics-web/app/decorators/notification_decorator.rb
+++ b/cosmetics-web/app/decorators/notification_decorator.rb
@@ -1,14 +1,11 @@
 class NotificationDecorator
-
   def initialize(notification)
     @notification = notification
   end
 
-  def to_csv
-    row = []
-    NotificationsDecorator::ATTRIBUTES.each do |attr|
-      row << @notification.send(attr)
+  def as_csv
+    NotificationsDecorator::ATTRIBUTES.map do |attr|
+      @notification.public_send(attr)
     end
-    row.join(",")
   end
 end
--- a/cosmetics-web/app/decorators/notification_decorator.rb
+++ b/cosmetics-web/app/decorators/notification_decorator.rb
@@ -1,0 +1,14 @@
+class NotificationDecorator
+
+  def initialize(notification)
+    @notification = notification
+  end
+
+  def to_csv
+    row = []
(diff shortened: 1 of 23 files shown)


cosmetics-web/app/decorators/notification_decorator.rb
```diff
@@ -0,0 +1,14 @@
...
+    @notification = notification
+  end
+
+  def to_csv
+    row = []
+    NotificationsDecorator::ATTRIBUTES.each do |attr|
+      row << @notification.send(attr)
+    end
+    row.join(",")
+  end
```
slorek, line 13: Please change: this will break if any of the values contain a `"` or `,` character. I would recommend using the CSV standard library, see https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/app/decorators/businesses_decorator.rb#L4-L9 for an example we used in PSD, as this will handle all of the escaping automatically.

slorek (changes requested):

cosmetics-web/app/decorators/notifications_decorator.rb
```diff
@@ -0,0 +1,26 @@
...
+    "Reference number",
+    "Notification date",
+    "EU Reference number",
+    "EU Notification date"]
+
+  ATTRIBUTES = [:product_name,
+                :reference_number,
+                :updated_at,
+                :cpnp_reference,
+                :cpnp_notification_date]
```
scruti, line 13: :question: Shouldn't this constant be declared within `NotificationDecorator`?

slawosz: I prefer to have `ATTRIBUTES` in here, as it also has headers here and this is parent class.

scruti (approved):

nasirkhanbeis:
Spotted few issues
1. Link to read as "Download a CSV of notified products"
2. time stamp to be in GMT format - not in UTC
3. Level 1, level2 and level 3 missing - agreed to fix this separately

slorek (approved):

nasirkhanbeis (approved):
looking good as expected. Re: category and subcategory, we agreed to implement this separately.

PR #2025, opened 2021-03-15
Add 2fa authenticator app to 2fa login

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1080)
# What does this PR does?
- When secondary authentication is required, asks the user which method to use (when multiple methods are set).
- Introduces two different pages for going through the secondary authentication process: SMS and Authenticator App.
- Refactors the secondary authentication code through the service so:
  - All the related controllers and forms are namespaced under `SecondaryAuthentication` module.
  - Splits the `SecondaryAuthentication` model class into 3 classes under the `SecondaryAuthentication` module/domain:
    - `DirectOtp`
    - `TimeOtp`
    - `Operations`  
- Sets all the secondary authentication routes under the `/two-factor/` path:
```
new_secondary_authentication_method       GET   /two-factor/method(.:format)           secondary_authentication/method#new
secondary_authentication_method           POST  /two-factor/method(.:format)           secondary_authentication/method#create
new_secondary_authentication_sms          GET   /two-factor/sms(.:format)              secondary_authentication/sms#new
secondary_authentication_sms              POST  /two-factor/sms(.:format)              secondary_authentication/sms#create
new_secondary_authentication_app          GET   /two-factor/app(.:format)              secondary_authentication/app#new
secondary_authentication_app              POST  /two-factor/app(.:format)              secondary_authentication/app#create
new_secondary_authentication_sms_resend   GET   /two-factor/sms/not-received(.:format) secondary_authentication/sms/resend#new
secondary_authentication_sms_resend       POST  /two-factor/sms/not-received(.:format) secondary_authentication/sms/resend#create
```

# Screenshots
**Users that have set both App and SMS authentication must choose which method to use:**

**When selecting SMS authentication:**

**When selecting Authenticator app authentication:**

**The Secondary Authentication related controllers have been placed under `SecondaryAuthentication` module:**

**Same with the models:**


nasirkhanbeis:
when user has set both options as sms and authenticator then I think we may need to show back link so then can return back to choice page where they can choose diff option 
image

nasirkhanbeis:
On invite user, 2fa is always defaulting to enter access code although user had setup both options. expected to show choose option page then show 2fa page accordingly

PR #2034, opened 2021-03-23
1119 limit file size

slawosz:
Please note that JS validation in general is not adviced by GDS Design Patterns, but in this case, we don't have much choice. This image can be used for test: https://effigis.com/wp-content/uploads/2015/02/Airbus_Pleiades_50cm_8bit_RGB_Yogyakarta.jpg


scruti:
Cannot say much about JS code as is not my area of expertise, but tried it in a review app and it works.

image

scruti (approved):

nasirkhanbeis:
re-tested this with file size 30 MB and it threw js error which is good, however  when I tried to upload multiple files with size 4.5 mb each but it failed to upload
see attached example files [](url)

nasirkhanbeis (approved):
Tested this on review app 2034, looking good as expected. I can see if the size limit exceeds it gives user friendly js pop-up.

PR #2037, opened 2021-03-23
Remove pre-post brexit question from manual flow

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1114)

We are unifying pre-post Brexit flows into a unique flow that will behave like the current post brexit flow:
- Enforcing the presence of a product image.
- Enforcing the presence of formulations instead of allowing to choose a predefined one.

This PR bring those changes into the manual notification flow.

Previously submitted notifications will still have the `was_notified_before_eu_exit` flag in DB as `true`, but new notifications will rely on the default `false` value.

The question enquiring for Pre/Post EU exit date has been removed from the manual flow.

Opening the PR against a feature branch so both Manual and Zip flows can be released together on April 1st.

## Description

**Changes to the manual journey (only for pre 1 Jan CPNP notified products)**

**Current journey**: For manual notification of cosmetic products notified on CPNP prior to 1 January - the user proceeds along a different journey to a complete manual notification - image labels and formulation documents are optional, they can only provide a frame formulation etc.

**What to change**: For manual notification of cosmetic products notified on CPNP prior to 1 January - the user should now just be directed to the complete manual notification journey.

**Screen to remove on the manual journey:**
image

When a user clicks ‘no I’ll enter information manually' → the next screen should be removed:

The user will then go onto the manual journey as if they were submitting a full notification.


cosmetics-web/app/controllers/responsible_persons/wizard/notification_build_controller.rb
```diff
@@ -70,8 +70,12 @@ def previous_wizard_path
...
 
     if step == :add_product_name
-      last_eu_step = @notification.was_notified_before_eu_exit ? :was_product_on_sale_before_eu_exit : :will_products_be_notified_in_eu
-      responsible_person_add_notification_path(@notification.responsible_person, last_eu_step)
+      last_step = if request.referer&.end_with? "do_you_have_files_from_eu_notification"
+                    :do_you_have_files_from_eu_notification
+                  else
+                    :will_products_be_notified_in_eu
+                  end
+      responsible_person_add_notification_path(@notification.responsible_person, last_step)
```
scruti, line 78: #### Why am I using the request referer?
When building the Back link for the `Add product name` page, it can be reached from:
- `Have you notified the EU?` --> [YES] --> `Do you have files?` --> [NO] --> `Add product name`
- `Have you notified the EU?` --> [NO] --> `Add product name`
As `Have you notified the EU?` does not set any flag in DB, and the `Do you have files?` step is only reachable when the EU has been notified. Previous behaviour was to default the backlink to  `Have you notified EU?` step.

If the referer is available it allows us to point to the immediate previous question when possible. If is not available we default to the `Have you notified EU?` question as before.

nasirkhanbeis (approved):

slawosz (approved):

PR #2054, opened 2021-03-30
Fix 7 day window to delete notification after submission

scruti:
Add a timestamp to the "Notification complete" state on the notification state machine.

It does:
- Add a field to DB that matches the state in the state machine.
- Backfills the field for all the already completed notifications.
  As the notifications cannot be edited once completed/submitted, the
  updated_at timestamp is used for the backfill.
- Sets AASM option to automatically set the timestamp when the
  Notification transitions to the "notification_complete" state.


slawosz (approved):

nasirkhanbeis (approved):
Tested this on review app 2054, looking good as expected.

PR #2056, opened 2021-03-31
Reindex elasticsearch everyday at night

slawosz:
There are cases of product not present in search, caused by missing entry in ES index. To ensure that products are always present in index, lets reindex all notifications at night.


cosmetics-web/config/initializers/sidekiq.rb
```diff
@@ -1,6 +1,6 @@
 def create_log_db_metrics_job
   log_db_metrics_job = Sidekiq::Cron::Job.new(
-    name: "log db metrics, every day at 1 am",
+    name: "log db metrics",
```
scruti, line 3: :clap:

cosmetics-web/config/initializers/sidekiq.rb
```diff
@@ -11,9 +11,23 @@ def create_log_db_metrics_job
...
+def create_elasticsearch_index_job
+  job = Sidekiq::Cron::Job.new(
+    name: "Reindex Elasticsearch, every day at 1 am",
+    cron: "* 1 * * *",
+    class: "ReindexElasticsearchJob",
+    queue: "cosmetics",
+  )
+  unless job.save
+    Rails.logger.error "***** WARNING - Elasticsearch reindexing job was not saved! *****"
+    Rails.logger.error elasticsearch_job.errors.join("; ")
```
scruti, line 23: From the diff seems `elasticsearch_job` is not defined and we should use `job` instead:
```suggestion
    Rails.logger.error job.errors.join("; ")
```

scruti (changes requested):
Please change: I think there is a call to an undefined variable that probably got renamed.
Suggested the fix.

scruti (approved):

PR #2061, opened 2021-04-01
BugFix: CPNP importer with missing parsing date

scruti:
Bug: https://sentry.io/organizations/beis/issues/2306373786/

#### Bugfix: Default for cpnp_notification_date missing
When the `cpnp_notification_date` is missing in the parser, the importer is raising an exception as it tries to compare a "nil" value against a date.

To resolve this, the comparison will only happen if cpnp_notification_date is present, defaulting to a false value for `Notification#was_notified_before_eu_exit` if not.

Also done some minor refactors on the Importer class.

### DIFF
--- a/cosmetics-web/app/services/cpnp_notification_importer.rb
+++ b/cosmetics-web/app/services/cpnp_notification_importer.rb
@@ -42,7 +42,7 @@ private
       components_are_mixed: @cpnp_parser.components_are_mixed,
       ph_min_value: @cpnp_parser.ph_min_value,
       ph_max_value: @cpnp_parser.ph_max_value,
-      was_notified_before_eu_exit: (@cpnp_parser.cpnp_notification_date < EU_EXIT_DATE),
+      was_notified_before_eu_exit: @cpnp_parser.cpnp_notification_date&.<(EU_EXIT_DATE),
     }
   end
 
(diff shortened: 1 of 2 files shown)


cosmetics-web/app/services/cpnp_notification_importer.rb
```diff
@@ -1,42 +1,61 @@
...
+      cpnp_reference: @cpnp_parser.cpnp_reference,
+      industry_reference: @cpnp_parser.industry_reference,
+      cpnp_notification_date: @cpnp_parser.cpnp_notification_date,
+      responsible_person: @responsible_person,
+      under_three_years: @cpnp_parser.under_three_years,
+      still_on_the_market: @cpnp_parser.still_on_the_market,
+      components_are_mixed: @cpnp_parser.components_are_mixed,
+      ph_min_value: @cpnp_parser.ph_min_value,
+      ph_max_value: @cpnp_parser.ph_max_value,
+      was_notified_before_eu_exit: @cpnp_parser.cpnp_notification_date&.<(EU_EXIT_DATE),
```
scruti, line 45: The actual bugfix is this change that only compares against `EU_EXIT_DATE` if the `cpnp_notification_date` is not `nil`.

slawosz (approved):

nasirkhanbeis (approved):
Tested this on review app, looking good as expected. works fine in the UI, when there's no date , it will simply parse and doesn't show first notified in EU in summary page.

PR #2070, opened 2021-04-08
Invalidate session on logout

slawosz:
Thanks to `devise-security` we are already storing `unique_session_id` key in session. If the key in DB is different then session value, session is invalidated.

This PR invalidates session on logout by changing DB value of `unique_session_id`.


cosmetics-web/spec/features/account/multiple_sessions_spec.rb
```diff
@@ -32,6 +32,38 @@
...
+      end
+
+      visit path
+      click_link "Sign out"
+
+      Capybara.using_session("Other browser") do
+        visit path
+        expect(page).to have_css("h2", text: "Your login credentials were used in another browser. Please sign in again to continue in this browser.")
+      end
+    end
```
scruti, line 66: Veeery nice test. Thaks for covering this :clap:

scruti (approved):

PR #2077, opened 2021-04-13
Add "set up of authenticator app" to guidance pages

scruti:
[Jira Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1118)
Not much to say, we're adding a guidance page for the authenticator app set up.


nasirkhanbeis:
looking good as per the given content. 
Wonder if we could add some step by step screen grabs to hand held user more clearer

bfanshawe (approved):
All looks good

PR #2078, opened 2021-04-14
Secondary authentication: Remove 2FA method memorisation from session. Fix previous 2FA session polluting Registration

scruti:
[Fixes COSBETA-1128](https://regulatorydelivery.atlassian.net/browse/COSBETA-1128)

**TLDR:** There were two issues causing the previous user 2FA session to take over a new user session. One was the storing of the 2FA method selection into the session (that was unnecessary) and the second was the fact that the user session wasn't being wiped when creating a new user.

## Overview
- Remove usage of the user session for recording the secondary authentication methods selection when multiple options are available.
- Enforce the secondary authentication to have been triggered as a condition to allow access to "Resend secondary authentication SMS code".
- **Bugfix**: Cleans previous user session when initiating new user registration.

## Description

When the user has multiple methods configured for secondary authentication, they need to choose which method to use when secondary authentication is prompted.

The method selection was being set/read at the user session.

There were some situations where when abandoning the flow when prompted for 2FA and then initiating a different flow that triggered a 2FA required action, caused the previous method selection, being used instead of prompting the user again for selecting a method.

This can be avoided and simplified by removing the use of session to store the method selection and just trust in the correct redirection happening when the user selects a method.

So every time the user needs to fill 2FA it gets sent to the method selection page and this form sends it to the SMS or APP authentication page.

### Bugfix description
Users partially signed in but not having completed the secondary authentication can reach the service landing page and click on the "Create an account" link. This causes issues as the previous user ID is kept in the user session for secondary authentication purposes and overrides 2FA configuration from the newly created user.
To avoid that, when the user registration page is reached, we wipe any previous user session.

### DIFF
--- a/cosmetics-web/app/controllers/secondary_authentication/sms/resend_controller.rb
+++ b/cosmetics-web/app/controllers/secondary_authentication/sms/resend_controller.rb
@@ -7,19 +7,19 @@ module SecondaryAuthentication
                          :set_cache_headers
 
       def new
-        @user = find_user
+        @user = user_with_secondary_authentication_request
         return redirect_to(root_path) unless @user
       end
 
       def create
-        @user = find_user
+        @user = user_with_secondary_authentication_request
         return render("errors/forbidden", status: :forbidden) unless @user
-        return resend_code unless @user.mobile_number_change_allowed?
+        return redirect_to new_secondary_authentication_sms_path unless @user.mobile_number_change_allowed?
 
         @user.mobile_number = mobile_number_param
         if resend_code_form.valid?
           @user.save!
-          resend_code
+          redirect_to new_secondary_authentication_sms_path
         else
           @user.errors.merge!(resend_code_form.errors)
           render(:new)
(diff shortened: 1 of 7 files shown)


nasirkhanbeis (approved):
Tested this on review app, looking good as expected. Nice work @scruti

cosmetics-web/app/controllers/secondary_authentication/sms/resend_controller.rb
```diff
@@ -46,8 +40,12 @@ def user_id_for_secondary_authentication
         current_user&.id || session[:secondary_authentication_user_id]
       end
 
-      def find_user
-        current_user || User.find_by(id: session[:secondary_authentication_user_id])
+      def user_with_secondary_authentication_request
+        if current_user && current_user.id == session[:secondary_authentication_user_id]
```
scruti, line 44: Change: Check if we actually need this and either remove it or comment the reasoning behind needing it.

slawosz (approved):

PR #2084, opened 2021-04-22
Allow Responsible Person members to edit the RP contact person details

scruti:
[Jira Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1091)

## Description
This PR enables members of a Responsible Person to edit the RP Contact Person details from the Responsible Person page.
The approach taken is consistent with the user details edition, with an individual edit page for each of the editable fields:
- Name
- Email address
- Phone number


nasirkhanbeis (approved):
Tested this on review app, looks good as expected.

cosmetics-web/app/controllers/application_controller.rb
```diff
@@ -22,7 +22,7 @@ class ApplicationController < ActionController::Base
   before_action :configure_permitted_parameters, if: :devise_controller?
 
   rescue_from "ActiveRecord::RecordNotFound" do |_e|
-    redirect_to "/404", status: :not_found
+    redirect_to "/404"
```
scruti, line 25: I changed this after @nasirkhanbeis noticed that introducing a wrong Contact Person id in the URL for editing the Contact Person, the browser was displaying a "You are being redirected" static page instead of redirecting to the `/404` page.

It seems when you redirect with a non-standard redirection status, browsers try to load the body of the page, and in this case, being a redirection, it defaults to the "You are being redirected" message.

I removed the "not found" status so we can globally catch any `ActiveRecord::RecordNotFound` error and redirect to `/404` page without the "You are being redirected" message being shown instead.

cosmetics-web/app/views/responsible_persons/contact_persons/edit_email_address.html.erb
```diff
@@ -0,0 +1,23 @@
+<% title = "Change contact person email address for #{@responsible_person.name}" %>
```
slawosz, line 1: Isn't lot of repetition in those 3 templates?

cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
```diff
@@ -7,27 +13,38 @@ def new; end
...
+
+    render view
+  end
 
   def update
-    if @contact_person.update(contact_person_params)
-      redirect_contact_person
+    # As the edit pages have only one field per page, updates will come with a single field being updated.
+    # Possible field values: "name", "email_address", "phone_number"
+    field = contact_person_params.keys.first
```
slawosz, line 32: In this case its not a big issue, but if `ContactPerson` would have some confidential fields, this allows them to be accessed. I would run it via `EDIT_FIELD_VIEW` anyway.

slawosz (approved):

cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
scruti: Don't think could be possible, `contact_person_params` is only permitting the expected parameters: 
```
    params.fetch(:contact_person, {}).permit(
      :email_address,
      :phone_number,
      :name,
    )
 ```

PR #2089, opened 2021-04-23
Search user bulk inviter tweaks

scruti:
This PR does:
1. Fix an issue when parsing emails from the CSV file for the bulk inviter.
2. Logs when trying to invite an email address corresponding to a registered user.

### Bug explanation
The emails parsed from the CSV file contain "end of line" symbols, like `user@example.org\n`.

This causes issues when attempting to re-invite users: 
- The user inviter attempts to find an existent user by email using `user@example.org\n`. 
- Does not find it because of the end of line difference, as the existent user email is `user@example.org`.
- The service attempts to create a new user with `user@example.org\n` but fails validation.
- Service fails and does not send the invitation.

### DIFF
--- a/cosmetics-web/app/services/invite_search_user.rb
+++ b/cosmetics-web/app/services/invite_search_user.rb
@@ -25,13 +25,15 @@ private
   end
 
   def send_invite
-    return if user.account_security_completed?
+    if user.account_security_completed?
+      Rails.logger.info "[InviteSearchUser] #{user.email} is already registered in the service and cannot be re-invited."
+    else
+      if !user.invitation_token || (user.invited_at < 1.hour.ago)
+        user.update! invitation_token: (user.invitation_token || SecureRandom.hex(15)), invited_at: Time.zone.now
+      end
 
-    if !user.invitation_token || (user.invited_at < 1.hour.ago)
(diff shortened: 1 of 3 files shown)


cosmetics-web/app/services/invite_search_user.rb
```diff
@@ -25,13 +25,15 @@ def create_user
   end
 
   def send_invite
-    return if user.account_security_completed?
-
-    if !user.invitation_token || (user.invited_at < 1.hour.ago)
-      user.update! invitation_token: (user.invitation_token || SecureRandom.hex(15)), invited_at: Time.zone.now
+    if user.account_security_completed?
+      Rails.logger.info "[InviteSearchUser] #{user.email} is already registered in the service and cannot be re-invited."
```
slawosz, line 29: I would not store email in logs, as we run this script from console I would simply `puts`. In case when we would have sophisticated mechanism for csv upload for invitation, we could add one off ids to CSV and log those.

slawosz (approved):

cosmetics-web/app/services/invite_search_user.rb
scruti: Resolved on chat. We're moving to use user id's instead of emails.

PR #2091, opened 2021-04-27
Bugfix: New email confirmation link using outdated token

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1133)

### BUG:
Users are requesting to change their email and receiving confirmation links with invalid tokens.

These tokens seem to be the previous token that was changed with the new email request. So, instead of receiving the newly generated token, users receive the token that was previously set and is no longer valid.

### CAUSE OF THE BUG:
Change of token and mailer sending the new email confirmation link was under the same DB transaction block.

### EXPLANATION:
When code is wrapped into an Active Record transaction block, changes are not written in DB until the full transaction block has been run successfully.

This means that when the mailer is called inside the transaction, even when the `#save!` has been called, the DB value has not been updated as the transaction has not finished.

When this happens, the mailer retrieves the new email confirmation token from DB and gets the previous token instead of the updated value, as the update in DB happens when the transaction block ends.

### DIFF
--- a/cosmetics-web/app/controllers/my_account/email_controller.rb
+++ b/cosmetics-web/app/controllers/my_account/email_controller.rb
@@ -14,19 +14,14 @@ module MyAccount
         return render :edit
       end
 
-      @user.new_email = dig_params(:new_email)
-
-      ActiveRecord::Base.transaction do
-        @user.save!
-        @user.reload.send_new_email_confirmation_email
-        render "users/check_your_email/show"
-      end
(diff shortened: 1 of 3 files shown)


slawosz (approved):

nasirkhanbeis (approved):
Verified all possible scenarios, change email without clicking confirm link, latest token works as expected. Also verified reset, create user journey, all good.

PR #2101, opened 2021-04-30
Bugfix: Only allow changing mobile number when was set on Account Security during registration

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1141)

Only allow changing the mobile number when the user had set their mobile number through the account security step.

If a user who only setup APP authentication goes to its profile and tries to set their mobile number from the "change" link, gets
into a "broken" state when the user has a non-verified mobile number but cannot receive its SMS code as the user 2FA method is set to "APP".

Due to this, the user could not even navigate through the service, as it would need to verify its mobile number but 2FA is not sending its verification code.

Setting up changes on Account security and allowing setting SMS account security is an incoming work, but for the time being, we don't want to allow users who didn't set their phone number on registration to be able to set it afterwards.

### Before
Screenshot from 2021-04-30 12-11-46

### After
Screenshot from 2021-04-30 12-29-20


nasirkhanbeis (approved):
Tested this on review app, looking good as expected. Nice one @scruti

PR #2122, opened 2021-05-13
Reorganize search

slawosz:
Ticket: https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1145
First step for improving search results on search part of service. Used `kaminari` for pagination, as its much easier to override HTML.

Before:
Screenshot from 2021-05-13 10-04-34
After:
Screenshot from 2021-05-13 10-03-54


DH02:
Some quick Review notes – sorry this reads a bit bluntly:

1. Section element using different width class
2. Form group div wrapper is missing
3. H1 tag with wrong class
4. H1 tag doesn’t have the nested label elements
5. Hint text/element in wrong parent divs
6. Form tag is missing an attribute
7. Input element not wrapped in the correct divs and classes
8. Input element missing classes
9. Button using input (with old classes) and not button element

HTML is in the prototype. The top row - the section this work focuses on - is self contained with an opening and closing <SECTION> element.

DH02:
Thank you - this looks a lot better: there are some remaining issues...

- Form element still missing attribute
- As the name on the search input is now ‘q’ the ‘for’ attribute on the main label will also have to be ‘q’ (so the two reference each other)
- The parent element wrapping the search form is still a div (not Section)
- In the second Section (for the results list), in the first ‘.row’ is a ‘… grid-column-full’. This should be ‘govuk-grid-column-two-thirds’ in anticipation of later sorting functionality.
- We can remove ‘aria-live="polite"’ until (or if) javaScript sorting is added.
- (I don’t know if showing the keywords functionality in the search results caption is being added next time but the keywords should appear in the bold span)
- The link within each search result currently uses ‘govuk-link--no-visited-state’ instead of ‘govuk-link’
- Something has changed on the magnifier glass icon in the search bar: it is now off centre?

DH02:
Item number one on the list (Section element with wrong width class) hasn't been amended.

scruti (approved):

nasirkhanbeis:
just had a look this pr, search domain looks good, however few suggestions
1. message underneath search box to show only when user searched for products, otherwise default or empty search should not show any result message
2. submit domain product list needs to be consistent across incomplete and notified tabs. Also I think we don't bullet formatted list. see attached screen. 
image

3. Thumbnails are broken on search domain which are same in prod too( pre-existing issue) would be nice to fix in the pr if its quick one

nasirkhanbeis (approved):
Tested this on review app and latest changes, looks all good as  expected.
There are some improvements to be done which can go in separate pr.

PR #2129, opened 2021-05-21
Allow to navigate back from SMS code page when finishing registration

scruti:
Allows users to navigate back from the SMS code authentication page to the account security page.
This SMS code authentication page is reached when the user selects "Text message" as a secondary authentication method. 

## Why?
If a user selects "Text message" in the account security page when completing their registration security page, it gets sent to the SMS code authentication page to verify their mobile number. 
If the user decides that this was a mistake and didn't want to use this secondary authentication method, they cannot go back or change this. So they are forced to use a mobile phone as an authentication method.

## Description

When the user navigates back, will need to re-introduce:
-  a new password for the account.
- the Time-based OTP code from the authenticator app (if the authenticator app is selected).
- For Search users: Their full name.

When the user navigates back, the form will have pre-filled according to their previous selection:
- Authenticator app and/or SMS text authentication selection.
- Authenticator app secret key. This allows users to keep the configuration in their phone and not delete the configuration and set it up again.
- Mobile phone number.

## Implementation details/comments
There are 2 different flows here:
- Submit users go through the "account security" page.
- Search users go through the "complete registration" page (that requires an invitation token).

The SMS authentication page reached after submitting "account security/complete registration" is shared and their back-link sends to one or another depending on the domain.

In both controllers for Submit/Search, we need to point to a route that resets the user account security flag and redirects to the form in order to allow the user to access the form, as at the SMS auth page their "account security" flag is completed.

Now, these forms will remember the user security options (if they choose Text/App authentication and their content) so the user can use the backlink without having to repeat all the process of deleting the previous setup and setting the app authentication again for a new QR code.

### DIFF
--- a/cosmetics-web/app/controllers/concerns/secondary_authentication_concern.rb
+++ b/cosmetics-web/app/controllers/concerns/secondary_authentication_concern.rb
@@ -12,7 +12,11 @@ module SecondaryAuthenticationConcern
     return unless user && Rails.configuration.secondary_authentication_enabled
 
     if !user.account_security_completed?
-      redirect_to(registration_new_account_security_path)
+      if submit_domain?
+        redirect_to registration_new_account_security_path
+      else
+        redirect_to complete_registration_user_path(user, invitation: user.invitation_token)
+      end
     elsif !secondary_authentication_present_in_session? || user.mobile_number_pending_verification?
       session[:secondary_authentication_redirect_to] = redirect_to
       session[:secondary_authentication_user_id] = user.id
--- a/cosmetics-web/app/controllers/registration/account_security_controller.rb
+++ b/cosmetics-web/app/controllers/registration/account_security_controller.rb
@@ -5,15 +5,24 @@ module Registration
     skip_before_action :try_to_finish_account_setup
 
     def new
-      @account_security_form = AccountSecurityForm.new(user: current_user)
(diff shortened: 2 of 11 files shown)


nasirkhanbeis (approved):
Tested below scenarios on review app, looking good as expected. 

1. Verify back link in registration journey, user should be able to click back and choose different method.
2. Verify registration with text message 2fa and then add authenticator through my account page
3. Verify registration with authenticator app 2fa and then add sms text through my account page. 
4. Verify invite user to prompt correct 2fa before inviting a user
5. Verify reset password  to prompt correct 2fa before resetting the password
6. Verify invite user to prompt user to choose 2fa method when both are configured, should be able to click back and select different method
7. Verify reset password to prompt user to choose 2fa method when both 2fa methods are configured - should be able to click back and select different method.
8. Verify user sign-in to prompt user to choose 2fa method when both methods are configured, should be able to click back and select different method. 
9. Verify delete notification to prompt user choose 2fa method when both methods are configured, should be able to click back and select different method. 
Other regression tests:
Unlock link in email to prompt user choose to 2fa method when both methods are configured and should be able to click back and select different method
Delete notification in incomplete and notified tab  to prompt user choose to 2fa method when both methods are configured and should be able to click back and select different method

slawosz (approved):

PR #2130, opened 2021-05-25
1147 order by categories

slawosz:


cosmetics-web/app/models/elasticsearch_query.rb
```diff
@@ -0,0 +1,59 @@
...
+  def query
+    {
+      query: {
+        bool: {
+          must: search_query,
+          filter: filter_query,
+        },
+      },
+    }
+  end
```
scruti, line 20: Please feel free to ignore this if both `build_query` and `query` are needed. 
Seems that the public interface needs `build_query`, while `query` itself is not called anywhere. But I may be missing something.

```suggestion
  def build_query
    {
      query: {
        bool: {
          must: search_query,
          filter: filter_query,
        },
      },
    }
  end
```

cosmetics-web/app/views/poison_centres/notifications/_filters.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          <h2 class="govuk-fieldset__heading">
+            Filter the results by<span class="govuk-visually-hidden">:</span>
+          </h2>
+        </legend>
+
+        <div class="govuk-form-group">
+          <label class="govuk-label" for="notification_search_form_category">
+            Product category
+          </label>
+          <%= form.select :category, NotificationSearchForm::CATEGORIES, { include_blank: true },  class: "govuk-select" %>
```
scruti, line 17: In case we want to avoid the hardcoding and restrict the dependency to the available form object in the view:
```suggestion
          <%= form.select :category, form.class::CATEGORIES, { include_blank: true },  class: "govuk-select" %>
```

cosmetics-web/app/controllers/poison_centres/notifications_controller.rb
```diff
@@ -18,11 +19,15 @@ def show
...
   end
 
-  def query_params
-    params.permit(:q)
+  def search_params
+    if params[:notification_search_form].present?
+      params.require(:notification_search_form).permit(:q, :category)
+    else
+      {}
+    end
```
scruti, line 31: Suggesting for consistency with how is done through other param methods:
```suggestion
    params.fetch(:notification_search_form, {}).permit(:q, :category)
```
Maybe it doesn't work with the `@form = NotificationSearchForm.new(search_params)` initialisation and you discarded that approach though.

scruti (approved):

nasirkhanbeis (approved):
Tested on review app, looking all good.

PR #2134, opened 2021-06-01
Bugfix: Broken product image thumbnails in notification pages for search users

scruti:
[Ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1154)

Image thumbnails were broken for search users due to the active storage route being disabled (re-routed to "/").

Enabled the route and overwrote the ActiveStorage representations proxy controller to add an access policy to the images.


nasirkhanbeis (approved):
Tested this on review app, looks all good as expected. 
Verified thumbnails as as search user on all chrome, FF and IE11. I can see image is rendered as expected. 
Also tested as submit user,  all good.

slawosz (approved):

PR #2137, opened 2021-06-03
Paginate registered notifications by completion

scruti:
The default pagination order based on "created_at" causes notifications that
were created days ago but completed recently not be displayed as the first
pagination results.

We want to order the results by the completion date/time. So the user will
be shown what notifications were notified most recently.

Example of affected results:

### Default order
Screenshot from 2021-06-03 15-41-36

### Ordering by completion date/time
Screenshot from 2021-06-03 15-41-15


slawosz (approved):

nasirkhanbeis (approved):
Tested on review app, looking good as expected.

PR #2143, opened 2021-06-08
Faster CSV generation

slawosz:
Cache categories in `Notification` so they are being generated faster. After deployment, `#cache_categories_for_csv!` needs to be called on each completed notification.

### DIFF
--- a/cosmetics-web/app/decorators/notification_decorator.rb
+++ b/cosmetics-web/app/decorators/notification_decorator.rb
@@ -12,8 +12,6 @@ class NotificationDecorator
 private
 
   def categories_to_csv
-    @notification.components.map { |component|
-      [component.root_category, component.sub_category, component.sub_sub_category].map(&:to_s).map(&:humanize)
-    }.flatten
+    @notification.csv_cache&.flatten || []
   end
 end
--- a/cosmetics-web/app/models/notification.rb
+++ b/cosmetics-web/app/models/notification.rb
@@ -108,7 +108,11 @@ class Notification < ApplicationRecord
       transitions from: :notification_file_imported, to: :draft_complete, guard: :formulation_present?
     end
 
-    event :submit_notification do
+    event :submit_notification, after: :cache_categories_for_csv! do
+      after do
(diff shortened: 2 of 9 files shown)


cosmetics-web/db/schema.rb
```diff
@@ -190,6 +190,7 @@
     t.decimal "ph_min_value"
     t.decimal "ph_max_value"
     t.datetime "notification_complete_at"
+    t.json "csv_cache"
```
scruti, line 193: Suggestion:
I would be more specific on the field name: `csv_categories_cache` or even `csv_categories` as is stored in SQL DB and the cache concept seems is redundant at this level.

cosmetics-web/app/models/notification.rb
```diff
@@ -108,7 +108,11 @@ def as_indexed_json(*)
       transitions from: :notification_file_imported, to: :draft_complete, guard: :formulation_present?
     end
 
-    event :submit_notification do
+    event :submit_notification, after: :cache_categories_for_csv! do
+      after do
+        cache_categories_for_csv!
+      end
+
```
scruti, line 115: The `after` hook seems to be called twice when should only be called once, right?
```suggestion
    event :submit_notification, after: :cache_categories_for_csv! do   
```

cosmetics-web/app/decorators/notification_decorator.rb
```diff
@@ -12,8 +12,6 @@ def as_csv
 private
 
   def categories_to_csv
-    @notification.components.map { |component|
-      [component.root_category, component.sub_category, component.sub_sub_category].map(&:to_s).map(&:humanize)
-    }.flatten
+    @notification.csv_cache&.flatten || []
```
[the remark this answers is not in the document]
slawosz: I prefer to have some structure on the data.

scruti (approved):
LGTM!

nasirkhanbeis (approved):
Tested csv download on review app, looking good as expected.

PR #2145, opened 2021-06-09
Fix errors when sending confirmation email to invited Submit users who didn't finish their account setup

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1174)
[Related error tracking](https://sentry.io/organizations/beis/issues/2441301077/)

### TLDR
Invited Submit users have no name until their complete their account security setup. When trying to send them a confirmation email that depends on the name, the mailer throws an exception.
Fixed by re-sending them the invitation email instead.

### Description
Some Submit invited users who followed the invitation link but didn't complete their registration were attempting to either:
- create a new account.
- reset their password.
Both situations were triggering the "resend user confirmation email" intended for users who initiated the registration but didn't complete their account security setup to be pointed to complete their registration.

This email depends on the user having a "name" attribute. But invited users set this value in their account security page, being empty until they complete this stage.
So when an invited user was triggering the "resend user confirmation", the user has no "name" and the mailer was throwing an exception.

### Solution
For Submit invited incomplete users, we will re-send the latest invitation instead of sending the confirmation email.
This way they will be able to complete their account setup.

### This is a temporal solution
This fix is meant to temporally resolve the bug.
We are planning to start including the user name in the invitations, this will allow us to unify the logic between self-registered and invited Submit users and default both to the confirmation email in order to finish their account setup.

### DIFF
--- a/cosmetics-web/app/controllers/users/passwords_controller.rb
+++ b/cosmetics-web/app/controllers/users/passwords_controller.rb
@@ -68,7 +68,13 @@ module Users
       user.confirmed_at = nil
       user.account_security_completed = false
       user.save(validate: false)
-      user.resend_account_setup_link
+      if (invitation = PendingResponsiblePersonUser.where(email_address: user.email).last)
+        SubmitNotifyMailer.send_responsible_person_invite_email(
+          invitation.responsible_person, invitation, invitation.inviting_user.name
+        ).deliver_later
+      else
+        user.resend_account_setup_link
+      end
       redirect_to check_your_email_path
     end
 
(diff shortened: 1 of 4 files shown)


nasirkhanbeis (approved):
Tested this on review app, looks all good as expected.

Verified below scenarios:

Verify it re-sends invite email when user tries to reset password before account setup through invite link. 

Invite user- click invite link and sign-out , create account with same email- should re-send invite email - Passed

Also tested regression scenarios. 
Invite user- click invite link and sign-out , create account with same email  and then reset password - should resend invite link - Passed

reset password - should send reset pwd for existing users - Passed
resend invite link to send invite email - Passed
create user, reset pwd before account account setup- should send create user email  - Passed

slawosz (approved):
LGTM

PR #2151, opened 2021-06-17
1148 date filter

slawosz:
Adds date filtering to notification search for Poison Centre/MSA users.
Screenshot from 2021-06-17 10-02-48


nasirkhanbeis:
Functionally works, however I see we need to improve few things
1. Validation errors on invalid date or left empty 
2.  It needs to retain the filtered result when I return back to list from product detail page. Eg. I’ve filtered by date, open a product and then click back, I should see date filter is still retained

cosmetics-web/app/controllers/poison_centres/notifications_controller.rb
```diff
@@ -1,7 +1,11 @@
 class PoisonCentres::NotificationsController < SearchApplicationController
+  PER_PAGE = 20
+
   def index
-    @form = NotificationSearchForm.new(search_params)
-    @result = search_notifications(10)
+    @search_form = NotificationSearchForm.new(search_params)
+    @search_form.valid?
```
scruti, line 6: :raising_hand_man:  Question:
I don't understand this validity check here without any flow/action depending on it. Is this necessary?

scruti (approved):
This work is great @slawosz . Congrats :clap:

nasirkhanbeis (approved):
Tested this on review app 2151, looking good as expected. 
Verified all possible scenarios around filters, keyword search and result text. looks all good as agreed.
There's a know limitation
note: when keyword is applied and later when you clear keyword search and click apply filter it wouldn't remove keyword unless user clicks on search icon.  I don't see this as issue because remove filters is not associated with the keyword search.

PR #2160, opened 2021-06-29
Allow to replace formulation files before completing notifications

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1183)
Allow changing the component formulation file before completing the notification.

The notification draft prior to confirmation will have a link in each component formulation file that will point to a page where the current file can be removed and a new file uploaded.


nasirkhanbeis (approved):
Tested this on review app, verified all possible scenarios, also tried to edit as diff RP, looks all good as expected.

slawosz (approved):
LGTM

PR #2164, opened 2021-07-02
Bugfix: Don't allow to change email to registered one

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1134)

### TLDR
- Add validation to `new_email` so: 
  - it is unique (2 users cannot change their email to the same new email address)
  - cannot be set to any existing `email` address value.
- Add validation to `email` so cannot be set to any existing `new_email` value on a different user.

### Description
The service is triggering unhandled errors when users attempt to change their email address to an email address that is already in use.

The user receives the "change email address confirmation" email, but when following the link they find an error page.
Emails are unique so the error is raised when attempting to confirm the email address change and update it on DB.

### Solution
We don't want to allow:
- Users changing their email addresses to already registered emails.
- Multiple users attempting to change their email address to the same email.
- Users registering accounts with an email address that is a new email address pending confirmation for another user.

When a user attempts to use an existing email address as its new email, it will still see the "Check your email" page, but nothing will be triggered/changed in the background and the user won't receive any email. 
We avoid informing users about other email addresses existence due to security concerns.


cosmetics-web/app/controllers/my_account/email_controller.rb
```diff
@@ -16,8 +16,13 @@ def update
 
       @user.new_email_pending_confirmation!(dig_params(:new_email))
       render "users/check_your_email/show"
-    rescue StandardError
-      render :edit
+    rescue ActiveRecord::RecordInvalid => e
+      # We don't want to let the user know when validation failed due to new email being already registered
+      if e.message.include? "has already been taken"
```
scruti, line 21: Not very happy with this. 
We want to display an error on "edit" page when the user attempts to use a wrongly formatted email, while we want to fail silently and show the "check your email" page when the user attempts to use an email that already belongs to another user.
Didn't find a simpler/better way to separate both cases as the exception class is the same :man_shrugging:
Happy to get suggestions though.

nasirkhanbeis (approved):
Tested this on review app, looks good as expected.

slawosz (approved):

PR #2173, opened 2021-07-08
Show list of submitted nanomaterial notifications

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1184)

Redesign of the nanomaterial notifications index page to show a list of submitted nanomaterial notifications for the current responsible person.

Each notification will include basic information:
- Name of the nanomaterial.
- Number (id) of the notification.
- Date when was notified to the EU (if so).
- Date when was notified in the UK.

Based on [this prototype](https://opss-html.herokuapp.com/nanomaterials-1149/nanomaterials)
image


nasirkhanbeis (approved):
Tested this on review app, looking all good as expected. 
Suggestion: can hold this from merging until we implement view nano page? Don't think it will be an issue to release though but would be nice release as feature

scruti:
@nasirkhanbeis thanks for the review.
About the suggestion, it would defeat the purpose of separating the tickets into 3: Being able to quickly release useful bits one by one and iterate adding value to the user.

Showing a list of notifications with some basic info about their submission dates is good enough to add some value (at least in my opinion). The individual pages providing access to the PDF attachment will be released soon after, adding extra value.

Let's release fast and iterate :raised_hands:

slawosz (approved):

slawosz:
Maybe you can create 30 nanonotifications in seeds and change `PER_PAGE` to 30?

PR #2175, opened 2021-07-09
Add nanomaterial notifications pages

scruti:
For the given nanomaterial notification the page will display:
- Name
- Nanomaterial number
- Date when was notified in the UK
- Date when was notified in the EU (if notified in the EU)
- Link to download PDF file.

### DIFF
--- a/cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
+++ b/cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
@@ -4,7 +4,7 @@ class NanomaterialNotificationsController < SubmitApplicationController
   before_action :set_responsible_person, only: %w[index new create]
   before_action :validate_responsible_person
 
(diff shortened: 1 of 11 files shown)


cosmetics-web/app/controllers/nanomaterial_notifications_controller.rb
```diff
@@ -16,6 +16,8 @@ def index
                                     .paginate(page: params[:page], per_page: PER_PAGE)
   end
 
+  def show; end
```
slawosz, line 19: Would be nice to add requests specs regarding authorization.

nasirkhanbeis (approved):
Note: didn't tested multi page as we have 20 per page which means we need to notify 20+ to check pagination with details. I assume it should return to correct page on back link in notification detail page.

slawosz (approved):

PR #2184, opened 2021-07-23
Create fully registered Search users in the seeds

scruti:
Instead of sending Search users invitations with every newly seeded Review App, seeds will create completely registered users with the same credentials as the Submit ones.


slawosz (approved):

slawosz:
LGTM providing you tested it locally (or review app?)

scruti:
Thanks @slawosz :raised_hands:. 
Yes, tested locally (good call) reproducing the ENV vars and seeding my local environment. Managed to log in directly with the seeded user.

PR #2192, opened 2021-07-28
Primary cosmetics page

slawosz:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1179


nasirkhanbeis:
Looks over all good,
there are some minor issues to address
1. when I open notification on page2 in complete notifications, on back link in summary page should return to page 2( retain the page where user is)
2. UK number in the table header to read as  "UKCP number" should be within abbr tag as Darren suggested . This should apply for both incomplete and complete notification table headers.
image
3. Suggestion : Nice to have. Incomplete and errors accordions to be default collapsed when I load the page first time. ( after login - in same browser) 

@slawosz  can also pls remove old view as I see links all working and I see no reason to keep old view any longer.

nasirkhanbeis (approved):
Tested both tickets, Cosmetics new design and reorder notifications. looking good as expected.

cosmetics-web/app/models/component.rb
```diff
@@ -9,7 +9,7 @@ class Component < ApplicationRecord
   set_max_file_size 30.megabytes
 
   attr_writer :skip_name_uniqueness_on_import
-  belongs_to :notification
+  belongs_to :notification, touch: true
```
scruti, line 12: :ok_hand:

cosmetics-web/app/views/responsible_persons/notifications/index.html.erb
```diff
@@ -10,48 +10,100 @@
...
           <span class="loading-spinner__dot"></span>
           Checking <%= pluralize(@pending_notification_files_count, 'notification file') %>
-        </h2>
-        <div class="govuk-error-summary__body">
-          <%= link_to "Refresh the browser to see uploaded products", url_for, class: "govuk-link--no-visited-state" %>
-        </div>
+      </h2>
+      <div class="govuk-error-summary__body">
+        <%= link_to "Refresh the browser to see uploaded products", url_for, class: "govuk-link--no-visited-state" %>
+      </div>
```
scruti, line 16: **Please revert this bit** :no_entry_sign: 
This tab removal from 13 to 16 looks wrong (`/h2` is misaligned with `h2`, 2 `</div>` are at the same level instead of having one nested inside the other).

scruti (approved):

PR #2196, opened 2021-08-03
Add reference number to search

slawosz:
Cosmetic product can be searched using UKCP number. https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1165


cosmetics-web/spec/features/notification_search_spec.rb
```diff
@@ -1,18 +1,19 @@
...
+RSpec.feature "Search", :with_stubbed_mailer, :with_stubbed_notify, :with_2fa, :with_2fa_app, type: :feature do
   let(:user) { create(:poison_centre_user, :with_sms_secondary_authentication) }
 
-  let(:notification1) { create(:notification, :registered, :with_component, notification_complete_at: 1.day.ago, product_name: "Cream") }
-  let(:notification2) { create(:notification, :registered, :with_component, notification_complete_at: 2.days.ago, product_name: "Shower Bubbles") }
-  let(:notification3) { create(:notification, :registered, :with_component, notification_complete_at: 3.days.ago, product_name: "Bath Bubbles", category: :face_care_products_other_than_face_mask) }
+  let(:cream) { create(:notification, :registered, :with_component, notification_complete_at: 1.day.ago, product_name: "Cream") }
+  let(:shower_bubbles) { create(:notification, :registered, :with_component, notification_complete_at: 2.days.ago, product_name: "Shower Bubbles") }
+  let(:bath_bubbles) { create(:notification, :registered, :with_component, notification_complete_at: 3.days.ago, product_name: "Bath Bubbles", category: :face_care_products_other_than_face_mask) }
```
scruti, line 9: Easier to track around. Nice.

scruti (approved):
Nice!

nasirkhanbeis (commented):
When I try search by full number eg. 'UKCP-10513701' it is returning all the notifications with the searched on the top. I think it should return only 1 row. looks like it is considering UKCP which is matching all the rows

nasirkhanbeis (approved):
Happy with current implementation. please note, when user search by UKCP prefix, it returns exact match on the top + returns rest all other rows,. Agreed to keep as is and when Bruce around we will check if we need to any further improvements.

PR #2201, opened 2021-08-11
Add address to search

slawosz:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1168


nasirkhanbeis (approved):
looks good as expected!

scruti (approved):
LGTM
Not really sure about adding an ElasticSearch queries section into the specs instead of adding feature test scenarios for full end to end testing. But guess this is faster to run in order to test multiple specific scenarios.

slawosz:
@scruti reasons for `spec/search` are multiple:
* its faster
* we avoid the burden of checking results in features (data vs HTML matches)
* it allows to play with internals, much easier than in `spec/features`

We will fine tune test and this might come handy.

PR #2206, opened 2021-08-12
Responsible Person page changes

scruti:
- [Prefix RP page title and header with "UK"](https://regulatorydelivery.atlassian.net/browse/COSBETA-1221)
- [Add Business type to RP page (after the address)](https://regulatorydelivery.atlassian.net/browse/COSBETA-1225).

image


nasirkhanbeis (approved):
Looks good to me. 
Verified h1 and heading and also business type.
Do we need to add the "'<abbr>'" tag on the UK Responsible person detail page too?

slawosz (approved):

PR #2211, opened 2021-08-13
Highlight field with error and fix anchors

slawosz:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1209
https://regulatorydelivery.atlassian.net/browse/COSBETA-1210

The idea for anchors is following:
Instead of using different anchor and thus `id` for error messages, we could determine on which element in frontend (so probably only in `app/views/poison_centres/notifications/_date_filter_input.html.erb` we should place an `id`. This way it would be much easier to implement.

Review app:
https://cosmetics-pr-2211-search-web.london.cloudapps.digital/
https://cosmetics-pr-2211-submit-web.london.cloudapps.digital/


cosmetics-web/app/helpers/application_helper.rb
```diff
@@ -43,4 +50,33 @@ def reference_number_for_display(notification)
   def wrap_summary_address(address_array)
     address_array.join("<span class=\"govuk-visually-hidden\">,</span><br />").html_safe
   end
+
+  # def error_id(search_form, attribute, part)
```
slawosz, line 54: I will get rid of this soon

cosmetics-web/app/models/gov_uk/date_from_form.rb
```diff
@@ -0,0 +1,72 @@
+module GovUK
+  class DateFromForm
```
slawosz, line 2: I spend some time trying to name it. The thing is, this should be located in `ActiveModel::Types::GovUKDate` but it does not work with Rails reloading, so its get called from there.

cosmetics-web/app/helpers/application_helper.rb
```diff
@@ -43,4 +50,33 @@ def reference_number_for_display(notification)
...
+  #       attribute
+  #     end
+  #   elsif attr.is_a? Date
+  #     if search_form.errors[attribute].present? && part == :day
+  #       attribute
+  #     end
+  #   end
+  # end
+
+  def error_class(search_form, attribute, part)
```
scruti, line 68: does this helper have any specific dependency with search forms? `error_class` seems quite generic but `search_form` seems specific.

cosmetics-web/app/models/gov_uk/date_from_form.rb
scruti: Maybe `DateFrom` ? The `Form` suffix clashes with the form objects naming (and originally made me thought this was a form object :sweat_smile:

cosmetics-web/app/models/gov_uk/date_from_form.rb
```diff
@@ -0,0 +1,72 @@
...
+    IncompleteDate = Struct.new(:year, :month, :day) do
```
scruti, line 3: Given that this got public methods and is not simply wrapping some attributes... maybe should be turned into a class?

scruti (approved):

PR #2225, opened 2021-08-24
Fix #password error message on sign in

stdstaff:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1153

## Description
There was an incorrect and confusing validation message being shown when a user tried to sign in with incorrect details:
image
This has been updated to no longer display validations with blank messages:
image
These would previously still display due to infering the href even when `error.message` was `nil`: https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/blob/5ca0e8932df8d74cf024d21dcf1e87755f10747f/cosmetics-web/app/helpers/application_helper.rb#L18-L24

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/Gemfile.lock
+++ b/cosmetics-web/Gemfile.lock
@@ -169,7 +169,7 @@ GEM
     factory_bot_rails (6.2.0)
       factory_bot (~> 6.2.0)
       railties (>= 5.0.0)
-    faker (2.18.0)
+    faker (2.19.0)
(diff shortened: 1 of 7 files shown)


scruti (approved):
LGTM!

nasirkhanbeis (approved):
looks good as expected!

PR #2231, opened 2021-09-01
Redesign "Change Responsible Person" page

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1239)

- Switch from a page with links to a Radio buttons selection form.
- Align HTML with GDS standards.
- Hide currently selected user RP from the available options to switch to.
- Present different titles depending on the context:
  - Adding a new RP when the user belongs to a single RP.
  - Changing RP when user belongs to multiple RPs.
  - Selecting RP when the user belongs to multiple RPs but haven't selected any of them as their current one.


cosmetics-web/app/controllers/responsible_persons_controller.rb
```diff
@@ -2,14 +2,23 @@ class ResponsiblePersonsController < SubmitApplicationController
...
+  before_action :responsible_persons_selection_form, only: %i[select change]
 
   def show; end
 
   def select; end
 
   def change
-    set_current_responsible_person(current_user.responsible_persons.find(params[:id]))
-    redirect_to responsible_person_notifications_path(current_responsible_person)
+    render :select and return unless @responsible_persons_selection_form.valid?
```
slawosz, line 12: For me, its not clear what this line does from the first glance, maybe it would be good to write it more 'oldschool' style. Also `and` is rarely used and maybe we should avoid it as its different from `&&`?

cosmetics-web/app/controllers/responsible_persons_controller.rb
```diff
@@ -2,14 +2,23 @@ class ResponsiblePersonsController < SubmitApplicationController
...
+
+    if @responsible_persons_selection_form.selection == "new"
```
slawosz, line 14: how about 
```
SelectionForm::NEW = 'new'
```
?

cosmetics-web/app/forms/responsible_persons/selection_form.rb
```diff
@@ -0,0 +1,30 @@
...
+      @radio_items ||=
+        available.sort_by(&:name).map { |rp| { text: rp.name, value: rp.id } }.tap do |items|
+          items << { divider: "or" } if items.any?
+          items << { text: "Add a new Responsible Person", value: :new }
+        end
+    end
+
+    # We remove the currently selected RP as a selection option
+    def available
+      @available ||= super.excluding(previous)
```
slawosz, line 19: Why super here?

slawosz: Oh, does it references 'bare' attribute? Wouldn't be `self['available']` more readable then?

cosmetics-web/app/controllers/responsible_persons_controller.rb
scruti, line 12: @slawosz commented _"For me, its not clear what this line does from the first glance, maybe it would be good to write it more 'oldschool' style. Also and is rarely used and maybe we should avoid it as its different from &&?"_

I intended to delete the draft of my response from VisualStudioCode Github integration but it actually deleted @slawosz comment, silly UX, sorry.

cosmetics-web/app/forms/responsible_persons/selection_form.rb
scruti: Gave it a go, but triggers  `SystemStackError stack level too deep` error after falling into an endless loop of calling itself.

slawosz (approved):

PR #2238, opened 2021-09-08
Add common password validation

slawosz:
https://regulatorydelivery.atlassian.net/browse/COSBETA-963

### DIFF
--- a/cosmetics-web/app/models/user.rb
+++ b/cosmetics-web/app/models/user.rb
@@ -15,6 +15,8 @@ class User < ApplicationRecord
   validates :new_email, email: { message: :invalid, allow_nil: true }
   validates :new_email, uniqueness: true, allow_nil: true
   validate  :new_email_not_registered
+  validates :password,
+            common_password: { message: I18n.t(:too_common, scope: %i[activerecord errors models user attributes password]) }
 
   validates :name, presence: true, unless: -> { invite }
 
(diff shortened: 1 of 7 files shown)


scruti (changes requested):
Please change: Add validation to the account security form.

When users attempt to use a common password during registration, as this validation is not added on the account security form, the form submission attempts to update the user and raises an exception due to user validation failing.

cosmetics-web/app/models/user.rb
```diff
@@ -15,6 +15,8 @@ class User < ApplicationRecord
   validates :new_email, email: { message: :invalid, allow_nil: true }
   validates :new_email, uniqueness: true, allow_nil: true
   validate  :new_email_not_registered
+  validates :password,
+            common_password: { message: I18n.t(:too_common, scope: %i[activerecord errors models user attributes password]) }
```
scruti, line 19: I18n can infer the scope and we can turn the validation message into simply `:too_common` 
```suggestion
  validates :password, common_password: { message: :too_common }
```

cosmetics-web/spec/forms/registration/account_security_form_spec.rb
```diff
@@ -191,6 +191,17 @@
       end
     end
 
+    context "when password is too common" do
+      let(:password) { "passowrd" }
+
+      it "does not validate user" do
+        user.password = "password"
```
scruti, line 198: :question: do we need this?

scruti (approved):
LGTM!

PR #2241, opened 2021-09-09
Reindex Elastic Search for each Review App deploy

scruti:
Search Cosmetics on Review Apps is failing to load the notifications page (500 error) due to a missing Elastic Search index.
This forces developers to connect to the review app console and manually reindex ES.

We're automating this to happen on deploy by adding a task triggering the reindex job and calling that task on the review app deploys manifest.

### Before 
image
### After
image

### DIFF
--- a/cosmetics-web/lib/tasks/elasticsearch.rake
+++ b/cosmetics-web/lib/tasks/elasticsearch.rake
@@ -1,0 +1,6 @@
+namespace :elastic_search do
+  desc "Reindex Elastic Search"
+  task reindex: :environment do
+    ReindexElasticsearchJob.perform_later
+  end
+end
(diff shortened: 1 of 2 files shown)


cosmetics-web/lib/tasks/elasticsearch.rake
```diff
@@ -0,0 +1,6 @@
+namespace :elastic_search do
+  desc "Reindex Elastic Search"
+  task reindex: :environment do
+    ReindexElasticsearchJob.perform_later
```
slawosz, line 4: How about `ReindexElasticsearchJob.new.perform`?

slawosz (approved):

PR #2243, opened 2021-09-10
Fix add component page back-link raising exceptions

scruti:
When reaching the page for a draft complete notification but not having
available completed components, the backlink generation was causing
an exception.

There is no point in trying to load the last component trigger
question page when there is no "last component".

Sentry related exceptions [1](https://sentry.io/organizations/beis/issues/2602567600/?project=1398436) and [2](https://sentry.io/organizations/beis/issues/2602567571/?project=1398436)

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/wizard/notification_build_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/wizard/notification_build_controller.rb
@@ -64,8 +64,7 @@ class ResponsiblePersons::Wizard::NotificationBuildController < SubmitApplicatio
                     :will_products_be_notified_in_eu
                   end
       responsible_person_add_notification_path(@notification.responsible_person, last_step)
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/controllers/responsible_persons/wizard/notification_build_controller.rb
```diff
@@ -64,8 +64,7 @@ def previous_wizard_path
                     :will_products_be_notified_in_eu
                   end
       responsible_person_add_notification_path(@notification.responsible_person, last_step)
-    elsif step == :add_new_component && @notification.state == "draft_complete"
-      last_component = @notification.components.complete.last
+    elsif step == :add_new_component && @notification.state == "draft_complete" && (last_component = @notification.components.complete.last)
```
scruti, line 67: was causing an exception when `last_component` was `nil`.

slawosz (approved):

PR #2252, opened 2021-09-16
Improve anti-spam validation on registration user names

scruti:
Follow ups on the anti-spam validation on registration user names.

### DIFF
--- a/cosmetics-web/app/forms/registration/new_account_form.rb
+++ b/cosmetics-web/app/forms/registration/new_account_form.rb
@@ -1,13 +1,13 @@
 module Registration
   class NewAccountForm < Form
-    BANNED_STRINGS = %w[: @ / http www].freeze
+    BANNED_REGEXP = /:|\/|@|www|http/.freeze
     FULL_NAME_MAX_LENGTH = 50
 
     attribute :full_name
 
     validates_presence_of :full_name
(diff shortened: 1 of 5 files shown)


cosmetics-web/app/forms/registration/new_account_form.rb
```diff
@@ -42,8 +46,8 @@ def send_link(user)
     def full_name_not_spam
       return if full_name.blank?
 
-      %w[: @ /].each do |invalid|
-        if full_name.include? invalid
+      BANNED_STRINGS.each do |banned|
```
slawosz, line 49: how about something like this?
```
class NameValidator < ActiveModel::EachValidator
  BANNED_REGEXP = /:|@|/|http|www/

  def validate_each(record, attribute, value)
    unless value.match?(BANNED_REGEXP)
      record.errors.add(attribute, options[:message])
    end
  end
end
```

And using in all forms and models that is using this

scruti: Nice, the regexp is way more efficient than looping. Will extract to the custom validator in further commits when extending to invitations.

nasirkhanbeis (approved):
looks good as expected.

slawosz (approved):

PR #2256, opened 2021-09-23
1157 notification soft delete

slawosz:
Instead of deleting notifications, removes all data from notification and save into `DeletedNotification` record. Original notification receives `deleted` state and hence is invisible to the system. As followup, we need to implement simple way to do actual deletion.


cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -251,4 +251,65 @@
...
+          notification.destroy!
+          expect(notification.reload.image_uploads).to eq [image_upload]
+        end
+
+        it "has responsible_person" do
+          notification.destroy!
+          expect(notification.reload.responsible_person).to eq responsible_person
+        end
+
+        it "is linked to deleted_notification" do
```
scruti, line 309: I think this is missing `notification.destroy!` call, and the current test may be passing because `expect(nil).to eq nil` returns true.

cosmetics-web/app/models/notification.rb
```diff
@@ -206,6 +226,27 @@ def destroy_notification!(submit_user)
...
+    end
+  end
+
+  def delete
+    raise "Not supported"
+  end
+
+  def delete!
+    raise "Not supported"
+  end
```
scruti, line 248: We could map alias these methods to `destroy!` instead?

scruti (approved):
A couple comments but LGTM !

PR #2284, opened 2021-10-14
Spec warning messages fixes

scruti:
## Description
Fixing issues regarding shared examples that were triggering warning messages when running the tests:
- Shared specs naming clashing.
- Shared specs file being loaded twice.

Also removed a couple of concern files for Trigger Rules since their definitions weren't used.

## Review apps


cosmetics-web/spec/support/shared_examples/common_user_tests.rb
```diff
@@ -53,17 +53,29 @@
...
-        include_examples "name format validations"
-      end
-
       describe "when changing the name" do
         before do
           user.name = "Emma McCay"
           user.save
         end
 
-        include_examples "name format validations"
```
slawosz, line 66: Why you removed shared examples here?

scruti: Having Shared examples defined inside other shared examples definition (here was happening in line 36) was causing a "definition is being overwritten" warning for every single spec run.
Every time the common user tests (parent shared example) got included in a spec file, would re-define the inner shared examples and trigger the warning.

slawosz (approved):

PR #2315, opened 2021-11-29
1299 nanomaterials

slawosz:


nasirkhanbeis:
Spotted a few minor issues in the latest build
Nano-materials question Yes/No - when nothing is selected then no validation message
Nano once selected Yes and given the count, then no means to remove - No remove functionality yet
Nano when left orphan then no warning yet - don’t think it’s implemented yet
When new nano is added on the task list page, then the product/items details page show completed but the “accept and submit section” goes in cannot start yet state - expected to show product details section “Not started” to be able to revisit and select new nano
Back link on a few pages is not working as expected. (There’s a ticket for this)
when I re-visit a task on the task list page, some questions are not pre-populated such as internal reference, product image, does product contains hair dye,pH range when mixed, item shades, product/item contained in
In single item journey - 
 The product details section is clickable even before nano details are entered yet. This means users can submit notifications although nanos are declared but not entered yet.

PR #2318, opened 2021-12-09
Allow users to cancel invitations to their Responsible Person

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1293)

## Description

Allowing Responsible Person users to cancel pending invitations to their Responsible Person.

Invitation routes have been refactored, so now we have:
- `/responsible_person/:rp_id/invitations/new` to add a new invitation from the team members page.
- `/responsible_person/:rp_id/invitations/:inv_id/resend` to resend an existing invitation.
- `/responsible_person/:rp_id/invitations/:inv_id/cancel` to cancel an invitation.

Team members controller now is only in charge of:
- listing the team members and pending invitations
- enable people with invitation links to join the team/RP.

## Screenshots
The cancel invitation button is now available on the team members page:
image

Clicking on it takes the user to a confirmation page:
image

User is redirected back to team members page and a confirmation message is displayed once the invitation has been cancelled:
image

## Review apps

https://cosmetics-pr-2318-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2318-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/invitations_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/invitations_controller.rb
@@ -42,7 +42,7 @@ private
   end
 
   def authorize_responsible_person
-    authorize @responsible_person, :show?
+    authorize @responsible_person, :update?
   end
 
   def send_invite_email
--- a/cosmetics-web/app/controllers/responsible_persons/invitations_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/invitations_controller.rb
@@ -21,6 +21,16 @@ class ResponsiblePersons::InvitationsController < SubmitApplicationController
     end
   end
 
+  def resend
+    ActiveRecord::Base.transaction do
+      @invitation.refresh_token_expiration!
+      @invitation.update!(inviting_user: current_user)
+      send_invite_email
+    end
+
+    redirect_to responsible_person_team_members_path(@responsible_person), confirmation: "Invite sent to #{@invitation.email_address}"
+  end
+
 private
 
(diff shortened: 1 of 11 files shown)


nasirkhanbeis (approved):

1. Any confirmed team member within the team should be able to remove the unconfirmed invited user
3. Validate when I click cancel the invitation, then I should be directed back to remove confirm page.
5. Validate when nothing is selected on the cancel invitation confirm page, then it should throw a validation error. Validate the error title in HTML(accessibility) 
7. Validate backlink cancel invite page, should return back to the team page
9. Validate that when I select “No” and click save and continue then I should return back to the teams page and the invitation is not canceled.
11. Validate when the user is removed, then invited token when clicked should return 404 page
13. Validate that when the invitation is canceled, then I should be able to invite the same email within the same team or any other team.
15. Validate that when the invite is removed for a user who self-registered but not confirmed, then the registration token should still work and I should not see an interim page where it shows the user is invited to the team.
17.  Validate that when an existing user is invited to a team, the user clicks the token but didn’t sign in yet, and when the invitation is cancelled then the token shouldn't let the user join the team.
19. Validate that when I click a cancelled invite token as another user, then nothing should happen, it should redirect to home page

Regression test:

1. Validate re-send should resend the invite token
2. Validate register user 
3. Validate invite user journey
4. Validate user self-registered but didn’t complete the 2fa yet and then the same user is invited, when I click register token then I should see an interim page where It should show I am invited to a team. (existing feature)

slawosz (approved):
Nicely done :)

PR #2277, opened 2021-10-12
New flow for adding notifications

slawosz:
https://cosmetics-pr-2277-submit-web.london.cloudapps.digital/

# New, simplified notification flow

Instead of having one, complicated flow with many redirections, this introduces task based notification flow.
To create a notification, the user have to finish several tasks:

1. Define a product
2. Define a nanomaterials
3. Define how multiple items are used (only for multi item products)
4. Define items
5. Submit

Each task is available from main, `draft` page. Unlike in old flow, it is possible to revisit and edit already completed task. 
It is also possible to add/remove nanomaterials and items.

# Why?

## Technical why

The code of the wizard was quite complicated - the biggest problems were *hard to follow* redirections which were causing a lot of troubles. There was some legacy code, no longer used - this work gave good opportunity to review it and get rid of it.

## UI why

The new task based flow gives user several benefits:
* predictable behaviour
* possibility of editing notification at any stage
* possibility of stopping adding notification without loosing progres

# Tasks
## Add product

Adding basic product informations: name, if contains nanomaterials or multiple items.

## Add nanomaterials

This task is available only if product notification has nanomaterial. In the task, the user needs to define nanomaterial name and confirms that nanomaterial conforms to regulations.

## Define multi-item kit

This task is available only if product notification has more than one item. It is used to gather information of PH in case when elements are mixed.

## Define items

The 'main' task. Gathers all infrmation of the product. Collects information about nanomaterials (if present), harmful ingredients and item photo and formulation

# Technical documentation

## State machine migration

This work introduces new states. `component_complete` is semi-new state - in old wizard, it was one of the intermediate state without any function. Now, it indicates that notification can be submited.
The mapping between old and new states is as following:

```
draft_complete => product_name_added
```

This transition occurs when `draft_complete` notification has missing some information. The user has to go through all of the steps again. `product name_added` is being listed so user can revisit such notification.

```
draft_complete => components_complete
```

This transition occurs when `draft_complete` notification has all information. The user can submit the notification right away.

```
product_name_added => empty
```

When notification had `product_name_added` state, only name was added. There is no point to keep such notification in the system. `empty` is state for new notification, however we don't list this state, so in practice it 'hiddes' notification. In the future we should remove all empty notifications.

 ```
import_country_added => empty
```
The `import_country_added_state` is old, long unused state. 

```
notification_file_imported => empty
```

This is old zip file state. We never listed this state so it is set to empty.

```
components_complete => empty
```

[...]

PR #2322, opened 2021-12-15
Enable notification hard deletion

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1333)

## Description
We need to enable the hard deletion of notifications while keeping the "soft delete by default" approach.
In this way, the notification can be removed from DB but only if we explicitly command to.

To achieve this, I re-structured the logic about Notification deletion:
- `#hard_delete!` <--- Aliased to the original `ActiveRecord::Persistence#destroy` behaviour.
- `#soft_delete!` <--- Explicit naming so we can distinguish between soft/hard deletion. The logic here was previously implemented as `#destroy!`
- `#destroy`, `#destroy!` <-- Aliased to `#soft_delete!`. So we keep soft deletion as default for AR calls.
- `#delete`, `#delete!` <--- Still disabled. 
- `#destroy_notification!` <--- Removed from the model. `DeleteNotification` service object wraps all logic now and gets directly called from the controller.

### Consequences of these changes.
We will be able to remove a notification (and its associated `NotificationDelete` instance) altogether. We will need to call `notification.hard_delete!` explicitly to do so.

Given the small amount of RPs we need to remove and how infrequent is this requirement, we will be able to check each case and, if required, hard delete their notifications before deleting the Responsible Person.
We can do it through a Rails Console on production, and we don't need more automation so far.

## Review apps

https://cosmetics-pr-2322-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2322-search-web.london.cloudapps.digital/


cosmetics-web/app/controllers/responsible_persons/delete_notification_controller.rb
```diff
@@ -6,7 +6,7 @@ class ResponsiblePersons::DeleteNotificationController < SubmitApplicationContro
   def delete; end
 
   def destroy
-    @notification.destroy_notification!(current_user)
+    NotificationDeleteService.new(@notification, current_user).call
```
scruti, line 9: Removed the `destroy_notification!` method as all the logic resides now inside the `NotificationDeleteService`.

cosmetics-web/app/models/notification.rb
```diff
@@ -2,23 +2,6 @@ class Notification < ApplicationRecord
...
-  has_many :components, dependent: :destroy
-  has_many :image_uploads, dependent: :destroy
-
-  has_one :deleted_notification
-
-  accepts_nested_attributes_for :image_uploads
-
-  index_name [ENV.fetch("ES_NAMESPACE", "default_namespace"), Rails.env, "notifications"].join("_")
-  scope :elasticsearch, -> { where(state: "notification_complete") }
-
```
scruti, line 21: Just reordering lines

cosmetics-web/app/models/notification.rb
```diff
@@ -65,6 +64,8 @@ def self.completed
                                           allow_nil: true
   validate :max_ph_is_greater_than_min_ph
 
+  delegate :count, to: :components, prefix: true
```
scruti, line 67: Moved it up.

cosmetics-web/app/models/notification.rb
```diff
@@ -35,6 +18,22 @@ class DeletionPeriodExpired < ArgumentError; end
...
+  include Searchable
+  include AASM
+  include CountriesHelper
+
+  belongs_to :responsible_person
+
+  has_many :components, dependent: :destroy
+  has_many :image_uploads, dependent: :destroy
+
+  has_one :deleted_notification, dependent: :destroy
```
scruti, line 30: Added the `dependent: :destroy` so when we call `notification.hard_delete!` the associated `deleted_notification`, if present, gets also removed.

cosmetics-web/app/models/notification.rb
```diff
@@ -235,20 +252,19 @@ def destroy!
...
-    raise "Not supported"
-  end
+  alias_method :destroy, :soft_delete!
+  alias_method :destroy!, :soft_delete!
 
   def delete!
     raise "Not supported"
   end
 
+  alias_method :delete, :delete!
```
slawosz, line 262: Is this a way to disable `delete`? If so, maybe a comment would be helpful here?

cosmetics-web/app/models/notification.rb
```diff
@@ -207,19 +208,35 @@ def via_zip_file?
...
-      destroy!
-    end
-  end
-
-  def destroy
-    destroy!
-  end
-
-  def destroy!
+  # =========================================
```
slawosz, line 211: There is not that much code regarding deletion, but with this comment and 'significance' for the service, maybe it makes sense to move it to concerns - as I moved states here: `https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/2277/files?authenticity_token=Pfd6cBgD2qdTW%2FCrwSi%2BCIh2aMd6V6rXgtk1o%2Bx3A%2BrHI%2FLi5iU6stmmIbiqU34natpqV9tzySMio5d%2BeStp9g%3D%3D&file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.rb&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&hide-deleted-files=true#diff-594f0156bc80880638e5a11ce1ad0ef70c212eb59d7bc626eb86198114ae270dR1`

slawosz (approved):
Simple and big improvement 😃

cosmetics-web/app/models/notification.rb
scruti: it is just pointing to the `delete!` implementation that is defined immediately above. I don't see much value in a comment here :thinking:

cosmetics-web/app/models/notification.rb
scruti: I was considering it when I wrote the comment and grouped the logic but decided against it because, not being used anywhere else, the cohesive concept of soft deletion wouldn't pay off for the cost of the added indirection.

If in the future we need this logic in another model I will be all in for the concern. For the time being, I thought the comment and grouping in the same model are enough.

PR #2347, opened 2022-01-05
Guard against missing frame formulations IDs

scruti:
## Description
When the wrong ID or SubID is given in the URL for the frame formulation page, instead of assuming the formulation/subformulation are found and trying to load the template, will redirect to 404 page, treating it as an ActiveRecord "not found" situation.

Noticed this after an attempt of SQL attack against this controller, passing SQL commands as ID/SUBID parameters.

The attack had no danger (especially as this controller does not query the DB), but caused 500 errors in the service that brought attention to the need to redirect to 404 instead.

Related Sentry issues:
- https://sentry.io/organizations/beis/issues/2908098191/?project=1398436
- https://sentry.io/organizations/beis/issues/2908098178/?project=1398436

## Review apps

https://cosmetics-pr-2347-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2347-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/controllers/frame_formulations_controller.rb
+++ b/cosmetics-web/app/controllers/frame_formulations_controller.rb
@@ -5,6 +5,9 @@ class FrameFormulationsController < PubliclyAccessibleController
 
   def show
     @formulation = FrameFormulations::ALL.find { |formulation| formulation["number"] == params[:id].to_i }
+    return redirect_to "/404" if @formulation.blank?
+
     @subformulation = @formulation["data"].find { |formulation| formulation["childNumber"] == params[:sub_id].to_i }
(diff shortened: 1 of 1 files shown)


slawosz (approved):

nasirkhanbeis (approved):
Verified frame formulation ids through guidance link in the footer, it is now returning 404 for any nonexisting frame formulation ids. Tested main and sub ids, all good.

Also, done regression testing on various other ids which are returning from db such as 
Responsible Person ids, notification ids, nanomaterial ids, etc. all are returning 404 for non-existing ones. 
Also tested formulation and images, it is returning undesigned 404 page which I can raise it separately. (not a high priority though)

PR #2356, opened 2022-01-10
Filter Pundit auth exceptions from Sentry

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1359)

These errors are meant to block users to access resources they shouldn't be able to. The exceptions are not runtime exceptions that may be caused by bugs or need to be resolved.

Given we have a very tight error quota in our Sentry plan and that Pundit errors give us no value, we will benefit from not sending them to Sentry.

 
## Review apps

https://cosmetics-pr-2356-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2356-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/config/initializers/sentry.rb
+++ b/cosmetics-web/config/initializers/sentry.rb
@@ -7,5 +7,11 @@ Rails.application.configure do
     config.breadcrumbs_logger = [:active_support_logger] # Inject Sentry logger breadcrumbs
     config.dsn = ENV["SENTRY_DSN"]
     config.send_default_pii = false
+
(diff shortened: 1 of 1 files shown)


slorek (commented):
The solution we used in PSD is little more elegant: https://github.com/UKGovernmentBEIS/beis-opss-psd/blob/master/config/initializers/sentry.rb#L11

scruti:
Thanks @slorek . I always forget to check PSD before googling around :sweat_smile:

slawosz (approved):

PR #2360, opened 2022-01-12
Bug Fix: date helper called when date is not present

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1363)

## Description
The construction of this line was causing the `#display_full_month_date` helper method to be called with nil values instead of only being called when the notification completion date is present.

This is causing 500 errors in production when Search users attempt to view the Search notifications index while having notifications that lack completion date in DB (probably ZIP imported).

## Review apps

https://cosmetics-pr-2360-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2360-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/views/poison_centres/notifications/_registered_product.html.erb
+++ b/cosmetics-web/app/views/poison_centres/notifications/_registered_product.html.erb
@@ -19,7 +19,7 @@
       Metadata
     </th>
     <td headers="uknotified <%= cell_headers %>" class="govuk-table__cell">
-      <%= display_full_month_date notification.notification_complete_at.presence || "&nbsp;".html_safe %>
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/views/poison_centres/notifications/_registered_product.html.erb
```diff
@@ -19,7 +19,7 @@
       Metadata
     </th>
     <td headers="uknotified <%= cell_headers %>" class="govuk-table__cell">
-      <%= display_full_month_date notification.notification_complete_at.presence || "&nbsp;".html_safe %>
+      <%= notification.notification_complete_at.present? ? display_full_month_date(notification.notification_complete_at) : "&nbsp;".html_safe %>
```
slawosz, line 22: As this is view helper, how about including this logic inside helper?

slawosz (approved):

PR #2375, opened 2022-01-21
1361 error display fix

slawosz:


nasirkhanbeis (approved):
Ah there's a problem here, showing wrong error message here

nasirkhanbeis (changes requested):
wrong error message Nano page, please take a look

PR #2420, opened 2022-02-16
Only offer date range to Search non-js users

scruti:
Instead of having all the date input filter fields expanded, causing
confusion when filling multiple fields but not selecting the radio
button for either date/date range filtering, now Search users using
browsers without JS, will only have the date range inputs displayed
and pre-selected.

## Review apps

https://cosmetics-pr-2420-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2420-search-web.london.cloudapps.digital/


scruti:
Found issues in the approach. Closing PR until we decide how to resolve it.

PR #2444, opened 2022-03-31
Add tool to set service on/off from maintenance

scruti:
## Description 
Turning the service into maintenance mode is a task that requires multiple map/unmap routes commands.

To avoid the requirement to have to do this manually each time we want to set the service on or off from maintenance, with the possibility of executing the wrong command or the confusion for new developers in the service, we have written a Thor client tool that automates this process.

This tool allows us to set the production pages on/off from maintenance and will ensure that the required services are available before attempting to do re-mappings.

## Screenshots
Screenshot from 2022-03-31 09-59-48
Screenshot from 2022-03-31 10-10-18

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ And those routes are explicitly disabled in application:
 ## Deployment
 
 Anything which is merged to `master` (via a Pull Request or push) will trigger the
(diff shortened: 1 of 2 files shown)


slorek (approved):
Very nice! 🙇

PR #2478, opened 2022-04-26
Remove unused code

slawosz:
Some unused code that was causing errors.


scruti (approved):

nasirkhanbeis (approved):
Tested poisonous and non-poisonous ingredients within a notification, looking good as expected. 
Also, verified on search domain, all good!

PR #2479, opened 2022-04-27
Add nanomaterial name uniquenes per notification validation

slawosz:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1436

Screenshot 2022-04-27 103554

### DIFF
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -82,10 +82,10 @@ private
   end
 
   def unique_name_per_nanomaterial
-    nano_elements = nano_material.notification.nano_materials.map(&:nano_elements).flatten
-    other_names = nano_elements.reject { |nano_element| nano_element.id == id }.map(&:inci_name)
-    other_names = other_names.compact.map(&:downcase).map(&:strip)
-    if other_names.include? inci_name.downcase.strip
+    nano_elements_with_same_name = NanoElement.where(nano_material: nano_material.notification.nano_materials)
+                                              .where.not(id: id)
+                                              .where("trim(lower(inci_name)) = ?", inci_name.downcase.strip)
+    if nano_elements_with_same_name.any?
       errors.add(:inci_name, "Enter a name which has not already been used in this notification")
     end
   end
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -86,7 +86,7 @@ private
                                               .where.not(id: id)
                                               .where("trim(lower(inci_name)) = ?", inci_name.downcase.strip)
     if nano_elements_with_same_name.any?
-      errors.add(:inci_name, "Enter a name which has not already been used in this notification")
+      errors.add(:inci_name)
     end
   end
 end
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -83,8 +83,9 @@ private
 
   def unique_name_per_nanomaterial
     nano_elements = nano_material.notification.nano_materials.map(&:nano_elements).flatten
-    other_names = nano_elements.map(&:inci_name)
-    if other_names.include? inci_name
+    other_names = nano_elements.reject{ |nano_element| nano_element.id == id }.map(&:inci_name)
+    other_names = other_names.compact.map(&:downcase).map(&:strip)
+    if other_names.include? inci_name.downcase.strip
       errors.add(:inci_name, "Enter a name which has not already been used in this notification")
     end
   end
--- a/cosmetics-web/app/models/nano_element.rb
+++ b/cosmetics-web/app/models/nano_element.rb
@@ -3,6 +3,7 @@ class NanoElement < ApplicationRecord
 
   # TODO: add uniqueness validation across notifications
   validates :inci_name, presence: true, on: :add_nanomaterial_name
+  validate :unique_name_per_nanomaterial, on: :add_nanomaterial_name
 
   after_save do
     if blocked?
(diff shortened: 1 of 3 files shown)


nasirkhanbeis (approved):
Tested this on the review app 2479, looks good as expected.

cosmetics-web/app/models/nano_element.rb
```diff
@@ -79,4 +80,13 @@ def usage_confirmed_required?
   def notification
     nano_material.notification
   end
+
+  def unique_name_per_nanomaterial
+    nano_elements = nano_material.notification.nano_materials.map(&:nano_elements).flatten
+    other_names = nano_elements.reject { |nano_element| nano_element.id == id }.map(&:inci_name)
+    other_names = other_names.compact.map(&:downcase).map(&:strip)
+    if other_names.include? inci_name.downcase.strip
```
scruti, line 88: I think we could turn all this into a single search in DB and let PostgreSQL deal with the matching without having to load all the notification nano elements in the application memory:

```suggestion
    nano_elements_with_same_name = NanoElement.where(nano_material: nano_material.notification.nano_materials)
                                              .where.not(id: id)
                                              .where("trim(lower(inci_name)) = ?", inci_name.downcase.strip)
    if nano_elements_with_same_name.any?
```

An example of this query running on my local machine and matching ` Nanom2 ` from DB with `nanom2` given string:
Screenshot from 2022-04-27 18-15-44

scruti: Query without spaces/formatting:
```ruby
NanoElement.where(nano_material: nano_material.notification.nano_materials).where.not(id: id).where("trim(lower(inci_name)) = ?", inci_name.downcase.strip)
```

cosmetics-web/app/models/nano_element.rb
```diff
@@ -79,4 +80,13 @@ def usage_confirmed_required?
...
+      errors.add(:inci_name, "Enter a name which has not already been used in this notification")
```
scruti, line 89: Could we add this error message to the `en.yml` locale model errors?

cosmetics-web/spec/models/nano_element_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+      let(:nano_element1) { create(:nano_element, inci_name: existing_name, nano_material: nano_material1) }
+      let(:nano_element) { build(:nano_element, inci_name: new_name, nano_material: nano_material2) }
+
+      before do
+        nano_element1
+      end
+
+      it "is valid with same name without context" do
+        nano_element.valid?
+        expect(nano_element.errors).to be_empty
```
scruti, line 49: Suggestion:
```suggestion
        expect(nano_element).to be_valid
```

cosmetics-web/spec/models/nano_element_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+        nano_element.valid?(:add_nanomaterial_name)
+        expect(nano_element.errors).to be_present
+      end
+
+      context "when saving" do
+        # rubocop:disable RSpec/PredicateMatcher
+        it "does not cause error on self" do
+          expect(nano_element1.valid?(:add_nanomaterial_name)).to be_truthy
+        end
+        # rubocop:enable RSpec/PredicateMatcher
```
scruti, line 62: Suggestion to avoid having to disable the Rubocop rule:
```suggestion
        it "does not cause error on self" do
          expect(nano_element1.valid?(:add_nanomaterial_name)).to be(true)
        end
```

cosmetics-web/spec/models/nano_element_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+          end
+        end
+
+        context "when new name differs only by case" do
+          let(:new_name) { existing_name.upcase }
+
+          it "is valid with same name with context" do
+            nano_element.valid?(:add_nanomaterial_name)
+            expect(nano_element.errors).to be_present
+          end
```
scruti, line 81: Spec description needs to say **in**valid. Suggestion on assertion.
```suggestion
          it "is invalid with same name with context" do
            expect(nano_element.valid?(:add_nanomaterial_name)).to be(false)
          end
```

cosmetics-web/spec/models/nano_element_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+      end
+
+      context "when names are similar" do
+        context "when new name differs only by whitespaces" do
+          let(:new_name) { "#{existing_name} " }
+
+          it "is valid with same name with context" do
+            nano_element.valid?(:add_nanomaterial_name)
+            expect(nano_element.errors).to be_present
+          end
```
scruti, line 72: Spec description needs to say **in**valid. Suggestion on assertion.
```suggestion
          it "is invalid with same name with context" do
            expect(nano_element.valid?(:add_nanomaterial_name)).to be(false)
          end
```

cosmetics-web/spec/models/nano_element_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+
+      it "is valid with same name without context" do
+        nano_element.valid?
+        expect(nano_element.errors).to be_empty
+      end
+
+      it "is valid with same name with context" do
+        nano_element.valid?(:add_nanomaterial_name)
+        expect(nano_element.errors).to be_present
+      end
```
scruti, line 55: Spec description needs to say **in**valid. Suggestion on assertion.
```suggestion
      it "is invalid with same name with context" do
        expect(nano_element.valid?(:add_nanomaterial_name)).to be(false)
      end
```

cosmetics-web/app/models/nano_element.rb
slawosz: Its a bit more confusing - you need to go back to notification, as new relation is really one to one for `NanoElement` and `NanoMaterial`. I used ruby as there wont be notifications with many nanomaterials so the code wont be a bottleneck.

scruti (approved):
Thanks for addressing all the comments! :bow:

PR #2488, opened 2022-05-05
Dont switch to new RP right away

slawosz:
When we are creating new RP, we don't want to switch to new RP right away.
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1468


cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
```diff
@@ -13,7 +13,11 @@ def new; end
 
   def create
     if @contact_person.save
-      redirect_to responsible_person_path(@responsible_person)
+      # After contact person creation we might like to go back to previous RP as
+      # we don't want to switch to new RP
+      # See also ResponsiblePerspon::AccountWizardController
+      set_proper_current_responsible_person
```
scruti, line 19:
```suggestion
      set_current_responsible_person_from_previous
```

nasirkhanbeis (approved):
looks good as expected!

PR #2489, opened 2022-05-09
WIP: Automate science team nanomaterials data pull

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1464)

## Review apps


cosmetics-web/app/jobs/upload_nanomaterials_pdfs_job.rb
```diff
@@ -0,0 +1,41 @@
+require "zip"
+require "fileutils"
+
+class UploadNanomaterialsPdfsJob < ActiveStorageUploadJob
+  FILE_NAME = "NanomaterialsPDFs.zip".freeze
```
slawosz, line 5: I wonder if we could include date of generation in the file name. It could allow users to track potential issues with task.

cosmetics-web/spec/jobs/upload_cosmetic_products_containing_nanomaterials_job_spec.rb
```diff
@@ -0,0 +1,5 @@
+require "rails_helper"
```
slawosz, line 1: I think you could add simple test to see if the generated files are correct. In all 3 cases.

slawosz (approved):

cosmetics-web/app/jobs/upload_nanomaterials_pdfs_job.rb
scruti: I don't see much value in attempting this given that whatever issue the user may find, will be with the latest version of the ZIP file, generated the night before.
We are also deleting every previous version of the files with each new upload. So a file name that changes with the date would also cause issues (or at least add extra complexity) with the search/delete of the previous version (we search/delete by name).

PR #2498, opened 2022-05-18
Fix label images display for MSA users

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1510)
[Error tracking issue](https://sentry.io/organizations/beis/issues/3281115279/?project=1398436)

The view for label images is shared between MSA users and Submit users.

When the label image has not passed the antivirus check, it shows a "Refresh" link that reloads the edit page for Submit users.

The same link was being called from Search MSA users, causing a routing error (as MSA users are not associated with a Responsible Person and the route parameter is missing).

MSA users shouldn't even know if the image is processing or not. Just having it displayed when is available and not displayed when it is not.

The fix only shows the link if edits are allowed for the user when loading the page (submit user belonging to the RP who created the notification).

## Review apps

https://cosmetics-pr-2498-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2498-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
+++ b/cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
@@ -195,10 +195,10 @@ module ResponsiblePersons::NotificationsHelper
      .compact
   end
 
-  def notification_summary_label_image_link(image, responsible_person, notification)
+  def notification_summary_label_image_link(image, responsible_person, notification, allow_edits: false)
     if image.passed_antivirus_check?
       link_to(image.filename, url_for(image.file), class: "govuk-link govuk-link--no-visited-state", target: "_blank", rel: "noopener")
-    elsif image.file_exists?
+    elsif image.file_exists? && allow_edits
       "Processing image #{image.file.filename}..." \
       "<br>" \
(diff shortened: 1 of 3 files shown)


cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
```diff
@@ -195,10 +195,10 @@ def notification_summary_component_rows(component, include_shades: true, allow_e
      .compact
   end
 
-  def notification_summary_label_image_link(image, responsible_person, notification)
+  def notification_summary_label_image_link(image, responsible_person, notification, allow_edits: false)
     if image.passed_antivirus_check?
       link_to(image.filename, url_for(image.file), class: "govuk-link govuk-link--no-visited-state", target: "_blank", rel: "noopener")
-    elsif image.file_exists?
+    elsif image.file_exists? && allow_edits
```
scruti, line 201: This branch is the bit that `msa` users were getting access to and was only intended for Submit users.

scruti, line 201: This branch is the bit that `msa` users were getting access to and was only intended for Submit users.

slawosz (approved):

PR #2514, opened 2022-06-13
Change cookie redirection fallback path

slawosz:


nasirkhanbeis (approved):
looks good as expected.

PR #2517, opened 2022-06-14
Add ingredients ranges concentrations new notification

scruti:
## Description

[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1501)

Manually add ingredients selecting the concentration range for each of the ingredients.
If the user marks the ingredient as poisonous, they need to introduce their exact concentration instead.

## Screenshots
image
----
image
----
image
----
image

## Review apps

https://cosmetics-pr-2517-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2517-search-web.london.cloudapps.digital/


cosmetics-web/app/helpers/component_build_helper.rb
```diff
@@ -3,4 +3,36 @@ def cmr_errors(component)
     component.cmrs.each_with_index.flat_map { |cmr, index| cmr.errors.map { |error| { text: error.message, href: "#component_cmrs_attributes_#{index}_#{error.attribute}" } } } +
       component.errors.reject { |error| error.attribute.to_s.include? "cmrs." }.map { |error| { text: error.message, href: "#component_cmrs_attributes_0_name" } }
   end
+
+  def ingredient_range_concentration_html_for(above, up_to)
```
slawosz, line 7: Personally, when reading code, I prefer to have first the 'main' methods that are used somewhere, then the helpers. So I would suggest such order:
```ruby
def ingredient_range_concentration_items; end

def ingredient_range_concentration_item(above, up_to); end

def ingredient_range_concentration_html_for(above, up_to); end
```
I think its easier to read, as `ingredient_range_concentration_items` is very undenstandable,  and 2 others are very detailed, probably used only in `ingredient_range_concentration_items` so when I see them being used there, I don't need to try to understand them as I know in which context they are used.

cosmetics-web/app/helpers/component_build_helper.rb
```diff
@@ -3,4 +3,36 @@ def cmr_errors(component)
...
+    "Above <span class='govuk-!-font-weight-bold'>#{above}%</span> " \
```
slawosz, line 8: Maybe partial instead of inline HTML would work here?

slawosz (approved):

nasirkhanbeis (approved):
Tested this on the review app, looks all good as expected

PR #2532, opened 2022-07-04
Add date range to search

slawosz:
https://regulatorydelivery.atlassian.net/jira/software/projects/COSBETA/boards/24?selectedIssue=COSBETA-1521
Screenshot 2022-07-04 133425


cosmetics-web/app/models/open_search_query/ingredient.rb
```diff
@@ -5,14 +5,22 @@ class Ingredient
 
     FIELDS = %w[searchable_ingredients].freeze
 
-    def initialize(keyword:, match_type:)
+    def initialize(keyword:, match_type:, from_date:, to_date:)
+      Rails.logger.debug match_type
```
scruti, line 9: Is this debugging line meant to be in production?

cosmetics-web/config/locales/en.yml
```diff
@@ -181,6 +181,19 @@ en:
               must_be_real: 'The To date must be a real date'
               incomplete: The To date must include a %{missing_date_parts}
               date_from_is_later_than_date_to: The to date must be after From date
+        ingredient_search_form:
+          attributes:
+            date_from:
+              blank: Enter a From date
+              in_future: The From date must be today or in the past
+              must_be_real: 'The From date must be a real date'
```
scruti, line 189:
```suggestion
              must_be_real: The From date must be a real date
```

cosmetics-web/app/forms/ingredient_search_form.rb
```diff
@@ -4,8 +4,47 @@ class IngredientSearchForm < Form
...
+            not_in_future: true,
+            if: :date_range_selected?
+
+  validates :date_to,
+            presence: true,
+            real_date: true,
+            complete_date: true,
+            if: :date_range_selected?
+
+  validate :date_from_lower_then_date_to
```
scruti, line 39:
```suggestion
  validate :date_from_lower_than_date_to
```

cosmetics-web/app/forms/ingredient_search_form.rb
```diff
@@ -4,8 +4,47 @@ class IngredientSearchForm < Form
...
+
+  def date_from_lower_then_date_to
```
scruti, line 41:
```suggestion
  def date_from_lower_than_date_to
```

cosmetics-web/app/forms/ingredient_search_form.rb
```diff
@@ -4,8 +4,47 @@ class IngredientSearchForm < Form
...
+    return unless valid?
+
+    return date_from if date_range_selected?
+  end
+
+  def date_to_for_search
+    return unless valid?
+
+    return date_to if date_range_selected?
+  end
```
scruti, line 24: You could simplify the public interface of the form by redefining `date_from` and `date_to` (using `super` to call the attributes) instead of using the `_for_search` suffix.
Example from `AccountSecurityForm`:

```ruby
    def app_authentication_code
      app_authentication_selected? ? super : nil
    end

    def mobile_number
      sms_authentication_selected? ? super : nil
    end
 ```
 
 Another bit is that, from what I read here, `return date_from if date_range_selected?` and `return date_to if date_range_selected?` do not seem to be needed.
 
The methods are early returning as `nil` if the validations fail.
So if either `date_from` or `date_to` are present they will both either be present or return `nil` from the validation guard clause.

File: cosmetics-web/app/forms/ingredient_search_form.rb [hunk 6]
```diff
@@ -4,8 +4,47 @@ class IngredientSearchForm < Form
...
+            not_in_future: true,
+            if: :date_range_selected?
+
+  validates :date_to,
+            presence: true,
+            real_date: true,
+            complete_date: true,
+            if: :date_range_selected?
+
+  validate :date_from_lower_than_date_to
```
scruti, line 39: Tiny please change: Move validations on :top: of the method declarations.

PR #2541, opened 2022-07-15
Unify ingredient model & migrate data

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1509)

## Description

We're replacing the `ExactFormula` and `RangeFormula` models/DB tables with a single `Ingredient` model.

There is a rake task that will migrate all the valid data from both tables into the new `Ingredient` one. 

**This ticket shouldn't modify the service interface/behaviour.**

### Data integrity:

The `Ingredient`s are associated with `Component`, and enforces data integrity through validations:
- An ingredient must contain `exact_concentration` or `range_concentration` but cannot have a value in both fields.
- `poisonous` flag only can be set with ingredients that have an `exact_concentration` value.
- An ingredient with `range_concentration` cannot belong to an `exact` or `predefined` component.
- Repeated ingredient names are not allowed for the same component unless they belong to a ZIP imported notification (in order to keep validity of old imported notifications).

### Data migration:
We introduce a rake task that will migrate all valid `ExactFormula` and `RangeFormula` entities into `Ingredient`. 
Some invalid entities have been detected and won't be imported:
- `ExactFormula`s where the quantity value was `0`.
- `RangeFormulas` without a range set.
- `RangeFormula` and `ExactFormula`s without an `inci_name`.
- `RangeFormula` and `ExactFormula`s not associated with any component/notification.

This data migration has been applied against a DB clone of Staging (currently running with this PR Review App).

Existing `ExactFormula` and `RangeFormula` existing data will be kept until we can ensure that staging/production new flow works well and that the data has been migrated correctly.
Then we will be able to open a further PR deleting those models/tables and all the data from them.

## Review apps

https://cosmetics-pr-2541-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2541-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
@@ -372,15 +372,13 @@ private
   def ingredient_form_values(ingredient, ingredient_number)
     return {} if ingredient.blank?
 
-    values = { name: ingredient.inci_name,
-               cas_number: ingredient.cas_number,
-               updating_ingredient: ingredient,
-               ingredient_number: ingredient_number }
-
-    case ingredient
-    when ExactFormula then values.merge!(exact_concentration: ingredient.quantity, poisonous: ingredient.poisonous)
-    when RangeFormula then values.merge!(range_concentration: ingredient.range, poisonous: false)
-    end
+    { name: ingredient.inci_name,
+      cas_number: ingredient.cas_number,
+      exact_concentration: ingredient.exact_concentration,
+      range_concentration: ingredient.range_concentration,
+      poisonous: ingredient.poisonous,
+      updating_ingredient: ingredient,
+      ingredient_number: ingredient_number }
   end
 
   def component_params
--- a/cosmetics-web/app/helpers/formula_helper.rb
+++ b/cosmetics-web/app/helpers/formula_helper.rb
@@ -1,9 +1,0 @@
-module FormulaHelper
-  def format_exact_formulas(exact_formulas)
-    exact_formulas.collect { |formula| { inci_name: formula[:inci_name], quantity: display_concentration(formula[:quantity]) } }
-  end
-
-  def format_range_formulas(range_formulas)
-    range_formulas.collect { |formula| { inci_name: formula[:inci_name], range: display_concentration_range(formula[:range]) } }
-  end
-end
--- a/cosmetics-web/app/helpers/ingredient_helper.rb
+++ b/cosmetics-web/app/helpers/ingredient_helper.rb
@@ -13,4 +13,16 @@ module IngredientHelper
                        .map(&:presence) # Converts ["", "0.1"] to [nil, "1"]
     range_klass.new(above, upto)
   end
+
+  def format_exact_ingredients(exact_ingredients)
+    exact_ingredients.map do |ingredient|
+      { inci_name: ingredient.inci_name, quantity: display_concentration(ingredient.exact_concentration) }
+    end
+  end
+
+  def format_range_ingredients(range_ingredients)
+    range_ingredients.map do |ingredient|
+      { inci_name: ingredient.inci_name, range: display_concentration_range(ingredient.range_concentration) }
+    end
+  end
 end
--- a/cosmetics-web/app/helpers/ingredient_helper.rb
+++ b/cosmetics-web/app/helpers/ingredient_helper.rb
@@ -16,13 +16,13 @@ module IngredientHelper
 
   def format_exact_ingredients(exact_ingredients)
     exact_ingredients.map do |ingredient|
-      { inci_name: ingredient.inci_name, quantity: display_concentration(ingredient.exact_concentration) }
+      { inci_name: ingredient.inci_name, exact_concentration: display_concentration(ingredient.exact_concentration) }
     end
   end
 
   def format_range_ingredients(range_ingredients)
     range_ingredients.map do |ingredient|
-      { inci_name: ingredient.inci_name, range: display_concentration_range(ingredient.range_concentration) }
+      { inci_name: ingredient.inci_name, range_concentration: display_concentration_range(ingredient.range_concentration) }
     end
   end
 end
--- a/cosmetics-web/app/models/component.rb
+++ b/cosmetics-web/app/models/component.rb
@@ -206,7 +206,11 @@ class Component < ApplicationRecord
     # Purge formulation files added in old flow.
     # Now ingredients need to be added manually or use a predefined formulation.
     formulation_file.purge
-    ingredients.destroy_all if old_type != notification_type
+    if old_type != notification_type
+      ingredients.destroy_all
+      exact_formulas.destroy_all
+      range_formulas.destroy_all
+    end
     update!(frame_formulation: nil, contains_poisonous_ingredients: nil) unless predefined?
   end
 
@@ -255,5 +259,6 @@ private
 
   def remove_poisonous_ingredients!
     ingredients.poisonous.destroy_all
+    exact_formulas.where(poisonous: true).destroy_all
   end
 end
--- a/cosmetics-web/app/models/component.rb
+++ b/cosmetics-web/app/models/component.rb
@@ -218,15 +218,6 @@ class Component < ApplicationRecord
     predefined?
   end
 
-
-  def poisonous_ingredients
-    exact_formulas.where(poisonous: true)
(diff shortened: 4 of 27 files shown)


cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
```diff
@@ -372,15 +372,13 @@ def ingredient_concentration_form
...
-    when ExactFormula then values.merge!(exact_concentration: ingredient.quantity, poisonous: ingredient.poisonous)
-    when RangeFormula then values.merge!(range_concentration: ingredient.range, poisonous: false)
-    end
+    { name: ingredient.inci_name,
+      cas_number: ingredient.cas_number,
+      exact_concentration: ingredient.exact_concentration,
+      range_concentration: ingredient.range_concentration,
+      poisonous: ingredient.poisonous,
+      updating_ingredient: ingredient,
+      ingredient_number: ingredient_number }
```
scruti, line 381: Now `Ingredient` contains `exact_concentration`, `range_concentration` and `poisonous` attributes. So they can be queried without conditional logic. When the value is not set will be returned as `nil` (or `false` for `poisonous`).

cosmetics-web/app/helpers/formula_helper.rb
```diff
@@ -1,9 +0,0 @@
-module FormulaHelper
-  def format_exact_formulas(exact_formulas)
-    exact_formulas.collect { |formula| { inci_name: formula[:inci_name], quantity: display_concentration(formula[:quantity]) } }
-  end
-
-  def format_range_formulas(range_formulas)
-    range_formulas.collect { |formula| { inci_name: formula[:inci_name], range: display_concentration_range(formula[:range]) } }
-  end
-end
```
scruti, line 9: Moved to `IngredientHelper`

cosmetics-web/app/helpers/ingredient_helper.rb
```diff
@@ -13,4 +13,16 @@ def ingredient_concentration_range(range)
...
+    exact_ingredients.map do |ingredient|
+      { inci_name: ingredient.inci_name, quantity: display_concentration(ingredient.exact_concentration) }
+    end
+  end
+
+  def format_range_ingredients(range_ingredients)
+    range_ingredients.map do |ingredient|
+      { inci_name: ingredient.inci_name, range: display_concentration_range(ingredient.range_concentration) }
+    end
+  end
```
scruti, line 27: Moved (and adapted to new `Ingredient` model) from `FormulaHelper`.

cosmetics-web/app/models/component.rb
```diff
@@ -16,6 +16,7 @@ class Component < ApplicationRecord
 
   has_many :exact_formulas, dependent: :destroy
   has_many :range_formulas, dependent: :destroy
+  has_many :ingredients, dependent: :destroy
```
scruti, line 19: We keep  `exact_formulas` and `range_formulas`, even when they're not called anymore. To be able to query and debug existing data against the migrated data that will be available as `ingredients`. Once all is good in production, a further PR will be open deleting the `exact_formulas` and `range_formulas` associations and DB tables.

cosmetics-web/app/models/component.rb
```diff
@@ -205,30 +206,14 @@ def update_formulation_type(type)
...
-    (exact_formulas + range_formulas).sort_by(&:created_at)
-  end
-
-  def poisonous_ingredients
-    exact_formulas.where(poisonous: true)
-  end
-
-  def non_poisonous_ingredients
-    (exact_formulas.where.not(poisonous: true) + range_formulas).sort_by(&:created_at)
-  end
```
scruti, line 230: Moved to`Ingredient` model as scopes.

cosmetics-web/app/models/component.rb
```diff
@@ -205,30 +206,14 @@ def update_formulation_type(type)
     # Purge formulation files added in old flow.
     # Now ingredients need to be added manually or use a predefined formulation.
     formulation_file.purge
-
-    if old_type != notification_type
-      exact_formulas.destroy_all
-      range_formulas.destroy_all
-    end
+    ingredients.destroy_all if old_type != notification_type
```
slorek, line 209: We may need to keep the destruction of `exact_formulas` and `range_formulas` for now to ensure investigative comparisons (as described in https://github.com/UKGovernmentBEIS/beis-opss-cosmetics/pull/2541/files#r935437128) are accurate.

cosmetics-web/app/models/ingredient.rb
```diff
@@ -0,0 +1,74 @@
...
+private
+
+  def unique_inci_name
+    return if inci_name.blank? || component.blank?
+
+    notification = component.notification
+    return if notification&.via_zip_file? || notification&.deleted?
+    return if inci_name_was&.casecmp(inci_name)&.zero? # Do not validate uniqueness if name is unchanged.
+
+    if self.class.where(component_id: component).where("LOWER(inci_name) = ?", inci_name.downcase).any?
```
slorek, line 52: Strong suggestion: Rails' `uniqueness` validation takes a `scope` option. We could consider using a `CITEXT` column type for case insensitive comparison. All the other conditionals inside this method can be implemented in a helper method and attached as an option to the built-in `validates` call. Given all these changes this whole method could then potentially be rewritten simply as something like:

`validates :inci_name, uniqueness: { scope: :component_id }, allow_blank: true, if: :notification_present?`

cosmetics-web/app/models/ingredient.rb
```diff
@@ -0,0 +1,74 @@
...
+      errors.add(:inci_name, :taken)
+    end
+  end
+
+  def poisonous_on_exact_concentration
+    if poisonous? && range_concentration.present?
```
slorek, line 58: Please change: We have a mixture of conditions inside the validation method but also on the `validate` method call with the `unless` option. Prefer one or the other.

cosmetics-web/app/models/ingredient.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+  end
+
+  def range_component_type
+    if range_concentration.present? && component && component.notification_type != "range"
+      errors.add(:range_concentration, :non_range_component)
+    end
+  end
+
+  def non_poisonous_exact_component_type
```
slorek, line 69: Please change: We have a mixture of conditions inside the validation method but also on the `validate` method call with the `unless` option. Prefer one or the other.

cosmetics-web/db/migrate/20220714124143_create_ingredient.rb
```diff
@@ -0,0 +1,27 @@
...
+                     greater_than_01_less_than_1_percent
+                     greater_than_1_less_than_5_percent
+                     greater_than_5_less_than_10_percent
+                     greater_than_10_less_than_25_percent
+                     greater_than_25_less_than_50_percent
+                     greater_than_50_less_than_75_percent
+                     greater_than_75_less_than_100_percent]
+
+      create_table :ingredients do |t|
+        t.string "inci_name", null: false
```
slorek, line 15: Suggestion: CITEXT type might be advantageous for the uniqueness validation here.

slorek (changes requested):

nasirkhanbeis (approved):
Tested this on the review app with stage data, looking good as expected.

cosmetics-web/app/models/ingredient.rb
```diff
@@ -0,0 +1,74 @@
...
+  validate :range_component_type
+
+private
+
+  def unique_inci_name
+    return if inci_name.blank? || component.blank?
+
+    notification = component.notification
+    return if notification&.via_zip_file? || notification&.deleted?
+    return if inci_name_was&.casecmp(inci_name)&.zero? # Do not validate uniqueness if name is unchanged.
```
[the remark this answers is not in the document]
scruti: The `inci_name_changed?` would be true even if the change is just on the capitalisation. Eg: `Ingredient` into `ingredient`.  

Seeing your comments about the conditionals being in different scopes (will fix that) and CITEX I think we can clean up quite a bit and this line should become unnecessary.

cosmetics-web/app/models/ingredient.rb
scruti: Very good suggestion :clap: . I have already used `CITEX` for non-case-sensitive email uniqueness validation. Totally forgot about it.

slorek (approved):

PR #2545, opened 2022-07-25
Bug - don't create nanomaterials if no is selected.

slawosz:
To reproduce:
Select "Yes" on question "Does {notification_name} contains nanomaterials" and enter amount:
Screenshot 2022-07-25 093404
Than change your answer to "No":
Screenshot 2022-07-25 093450
After submit, there will be 4 nanomaterials created.

This PR contains fix.


nasirkhanbeis (approved):
Tested this on the review app, looks good as expected!

PR #2564, opened 2022-08-16
Fix the Single or Multi component prefilling

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1507)

When revisiting the notification builder for an already completed task,
prefill the correct data in the Single or Multi component form.

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
@@ -150,9 +150,25 @@ private
       ResponsiblePersons::Notifications::Product::ContainsNanomaterialsForm.new(contains_nanomaterials_params)
   end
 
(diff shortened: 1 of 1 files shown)


nasirkhanbeis (approved):

cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
```diff
@@ -150,9 +150,25 @@ def contains_nanomaterials_form
       ResponsiblePersons::Notifications::Product::ContainsNanomaterialsForm.new(contains_nanomaterials_params)
   end
 
+  def single_or_multi_component_values
```
slawosz, line 153: I would move this method below the `single_or_multicomponent_form`. It is easier to read from broader context code to narrow.

slawosz (approved):

PR #2572, opened 2022-08-19
Fix missing shades inputs

scruti:
## Description
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1376)

The check over the component having `nil` as shades value works fine for pre-creating two empty shades when reaching the "add shades" page for the first time.

The problem is that when the user skips the addition of the shades by not introducing any value in the shades fields, the component shades value will become an empty array `[]`.

This is not very consistent from a model data point, and we will probably want to migrate all production data to ensure we go for either `nil` or `[]` as shades attribute value when the component has no shades.

Besides that, there is a user interface issue when not filling any shades and then re-visiting the step and moving back to the same point again.

With the current code, a component with shades `[]` won't match the condition to display two empty shades in the form so the form will be displayed without any input field.

Checking that there are no shades (empty array or nil) resolves this and ensures user flow consistency. Either when visiting it for the first time or revisiting it.

## Screenshots of flow to reproduce the bug
Screenshot from 2022-08-19 11-58-31
Screenshot from 2022-08-19 11-58-42
Screenshot from 2022-08-19 11-58-53
Screenshot from 2022-08-19 11-59-05
Screenshot from 2022-08-19 12-00-27

## Review apps

https://cosmetics-pr-2572-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2572-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
@@ -71,7 +71,7 @@ class ResponsiblePersons::Notifications::Components::BuildController < SubmitApp
     when :select_nanomaterials
       return jump_to_step(:number_of_shades) if @component.notification.nano_materials.blank?
     when :add_shades
-      @component.shades = ["", ""] if @component.shades.nil?
+      @component.shades = ["", ""] if @component.shades.blank?
     when :add_cmrs
       create_required_cmrs
     when :after_select_nanomaterials_routing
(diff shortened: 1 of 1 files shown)


cosmetics-web/app/controllers/responsible_persons/notifications/components/build_controller.rb
```diff
@@ -192,7 +192,7 @@ def update_add_exposure_routes
   def update_add_shades
     @component.update(component_params)
 
-    if params.key?(:add_shade) && params[:add_shade]
+    if params[:add_shade]
```
scruti, line 195: This is just a refactor.

nasirkhanbeis (approved):
Agree that shades info is not mandatory so it is stored null too in case, so we don't need validation around but we need to show the default field when the user revisits the shades page. Thanks, Marc!

slawosz (approved):

PR #2573, opened 2022-08-19
Github workflows upgrades

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1557)

## Description

- Upgrade Postgres to v13.6 in Github workflows to match with the Postgres version running in review apps and
production environments.
- Migrate away from deprecated action to set up Ruby.
- Upgrade actions to their latest versions.
- Split Rspec in two parallel running jobs (one for feature specs and one for the rest), moving from 8-11 minutes test runs to 4-5 minutes.
- Fix bug in workflow for deleting dangling review apps.
- Do not attempt to create/delete PR for PRs initiated by Dependabot. Since they will fail, better to skip them.

### DIFF
--- a/cosmetics-web/app/models/concerns/searchable.rb
+++ b/cosmetics-web/app/models/concerns/searchable.rb
@@ -65,7 +65,10 @@ module Searchable
       # This line makes sure opensearch index is recreated before we search
       # It fixes the issue of getting no results the first time product list page is loaded
       # It's only used in dev because it lowers performance and the issue it fixes should be an edge case in production
-      __elasticsearch__.refresh_index! if Rails.env.development? || Rails.env.test?
+      if Rails.env.development? || Rails.env.test?
+        __elasticsearch__.create_index! unless __elasticsearch__.index_exists?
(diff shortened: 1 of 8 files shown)


cosmetics-web/app/models/concerns/searchable.rb
```diff
@@ -65,7 +65,10 @@ def self.full_search(query)
       # This line makes sure opensearch index is recreated before we search
       # It fixes the issue of getting no results the first time product list page is loaded
       # It's only used in dev because it lowers performance and the issue it fixes should be an edge case in production
-      __elasticsearch__.refresh_index! if Rails.env.development? || Rails.env.test?
+      if Rails.env.development? || Rails.env.test?
+        __elasticsearch__.create_index! unless __elasticsearch__.index_exists?
```
scruti, line 69: Since I did split the Rspec tests, the job in the job matrix running the feature specs was failing due to the ES index not existing. So guess better safe than sorry checking/ensuring this at this point.

slawosz (approved):

PR #2575, opened 2022-08-23
Fix error when removing images without JS-enabled

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1199)

## Description

When browsing without JS enabled, the link to remove an image was triggering a server error.
This is caused by the delete links mapping to an AJAX call that cannot be executed without JavaScript.

Fixed it by replacing the deletion link with a deletion button that triggers a form POST request.

### When JS is enabled
Screenshot from 2022-08-23 14-21-46

### When JS is disabled/not available
image

## Review apps

https://cosmetics-pr-2575-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2575-search-web.london.cloudapps.digital/


nasirkhanbeis (approved):
Tested this on the review app, looking good as expected. Verified on both JS and nonJS browser.

slawosz (approved):

PR #2621, opened 2022-10-11
[DO NOT MERGE] Test

slawosz:
After creation of new review app, the initial indexing is not working - we need to do manuall reindexing.


slawosz:
Looks like changes in previous PR made it working

PR #2638, opened 2022-10-18
Updates to navigation and URLs within the main header

AndrewCowle:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1588

## Description
Updates to the main header. Added in a products-redirect on the responsible persons controller to use the concern to get a current responsible person. This ensures the nav works correctly when visiting a shared controller (My Account) without exposing the responsible person to search.

## Review apps

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only

### DIFF
--- a/cosmetics-web/.gitignore
+++ b/cosmetics-web/.gitignore
@@ -1,4 +1,4 @@
-.env.development
+.env
 .env.test
 # Created by https://www.gitignore.io/api/ruby,node,rails
 
--- a/cosmetics-web/.gitignore
+++ b/cosmetics-web/.gitignore
@@ -1,3 +1,4 @@
+.env.development
 .env
 .env.test
 # Created by https://www.gitignore.io/api/ruby,node,rails
--- a/cosmetics-web/app/controllers/responsible_persons_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons_controller.rb
@@ -1,8 +1,12 @@
 class ResponsiblePersonsController < SubmitApplicationController
+  include ResponsiblePersonConcern
+
   before_action :set_responsible_person, only: %i[show edit update]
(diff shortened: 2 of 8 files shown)


cosmetics-web/.gitignore
```diff
@@ -1,4 +1,4 @@
-.env.development
```
slawosz, line 1: Some of us still use `.env.develepoment`

cosmetics-web/app/controllers/responsible_persons_controller.rb
```diff
@@ -1,9 +1,13 @@
 class ResponsiblePersonsController < SubmitApplicationController
+  include ResponsiblePersonConcern
```
slawosz, line 2: `ResponsiblePersonConcern` is already in `SubmitApplicationController`

cosmetics-web/app/controllers/responsible_persons_controller.rb
```diff
@@ -36,6 +40,10 @@ def update
     end
   end
 
+  def products_redirect
```
slawosz, line 43: I think better name would be `products_page_redirect`

cosmetics-web/.gitignore
AndrewCowle: But this shouldn't be committed to git right - env.development.example should be there as a template for others to use, you could commit things like your notify key etc

slawosz: Yes, so now, without `.env.development` in `.gitignore` my `.env.development` is on untracked files list. We don't want to remove it as we want to ignore it.

slawosz (approved):

PR #2642, opened 2022-10-19
Dont recreate index

slawosz:
We were removing index on every reindexing job. When they were failing, this was causing lack of searchable notifications in the index. Now, index is created only when it does not exists - it will be used mostly by review apps.


slorek (approved):
Bear in mind that (I think) if you ever change the JSON Opensearch representation schema you'll need to manually re-import with the `force` option going forward.

PR #2646, opened 2022-10-21
Update notification summary page with new nano info

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1544)
## Description

On the notification summary pages for both Submit and Search users, we want to change how we display the non-standard nanomaterials.

### Submit users will see:
- A new nanomaterial listing format including the UKN code, name and a link to their notification PDF.
- A new section displaying when the non-standard nanomaterials review period ends.

### Search OPSS users will see:
- A new nanomaterial listing format including the UKN code, name and a link to their notification PDF.
- A new section displaying when the non-standard nanomaterials review period ends.

### Search MSA users will see:
- A new nanomaterial listing format including the UKN code and name.
- A new section displaying when the non-standard nanomaterials review period ends.

### Search Poison Centre users will see:
- A new nanomaterial listing format including the UKN code and name.

### Submit user's view:
image
----
### Search OPSS Science user's view:
image
----
###  Search MSA user's view:
image
----
### Search Poison centre user's view:

image

## Review apps

https://cosmetics-pr-2646-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2646-search-web.london.cloudapps.digital/


cosmetics-web/app/models/nanomaterial_notification.rb
```diff
@@ -85,6 +87,10 @@ def can_be_made_available_on_uk_market?
     can_be_made_available_on_uk_market_from <= Time.zone.now
   end
 
+  def ukn
```
AndrewCowle, line 90: should this go into the decorator?

AndrewCowle (approved):

cosmetics-web/app/models/nanomaterial_notification.rb
scruti: Good point, @AndrewCowle. I initially thought the same, but when I opened the corresponding decorator found that our "decorators" are POROs for generating CSV files, and they are not using Draper gem. 

So, as far as I know, without Draper, would be quite a bit of a pain to instantiate for every use case of the `#ukn` method.

nasirkhanbeis (approved):
Tested this on the review app, looking all good expected.

slawosz (approved):

PR #2649, opened 2022-10-25
Error messages updates during account RP creation

AndrewCowle:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1029

## Description
Issue when not selecting either option doesn't show error message as intended. Logic was already applied just not checking for `nil?` - as it's only a single form field and logic already present doesn't warrant the time/effort for a FormObject.

Screenshot 2022-10-25 at 11 20 40

## Review apps

https://cosmetics-pr-2649-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2649-search-web.london.cloudapps.digital/


slawosz (approved):

cosmetics-web/app/views/responsible_persons/account_wizard/create_or_join_existing.html.erb
```diff
@@ -4,7 +4,7 @@
 <% end %>
 
 <%= form_with url: wizard_path, method: :put do |form| %>
-  <% error_message = @nothing_selected && { text: "Select an option before continuing", href: "#option-error" } %>
+  <% error_message = @nothing_selected && { text: "Select yes, if you or your organisation have an account", href: "#option-error" } %>
```
scruti, line 7: @nasirkhanbeis may confirm this. And seems it wasn't mentioned in the ticket as requirement. 
When displaying validation errors on radio/check forms, we follow the practice of pointing the summary error link to the first option. In this case to `#option_join_existing` instead of `#option-error`.

This way when we click on the summary error message we get here:
image

AndrewCowle: Ah thanks for flagging, swapped that over now.

PR #2665, opened 2022-11-01
Use .profile to set up the app environment

scruti:
Shamelessly copied from PSD: https://github.com/OfficeForProductSafetyAndStandards/product-safety-database/pull/2110

- Using new app initialisation method with a .profile file: https://docs.cloudfoundry.org/devguide/deploy-apps/deploy-app.html#profile
- Provides a shorthand for running processes within a launched environment: We moved from: `/tmp/lifecycle/launcher /home/vcap/app 'bin/rails c' ''` to: `bin/tll bin/rails c`


README.md
```diff
@@ -10,7 +10,7 @@
 
 ## Inviting Search Users
 
-1. SSH and run rails console: `cd app && export $(./env/get-env-from-vcap.sh) && /tmp/lifecycle/launcher /home/vcap/app 'rails c' ''`.
+1. SSH and run rails console: `cd app && bin/tll bin/rails c`.
```
stdstaff, line 13: FWIW you don't need to be in `./app` so it's possible to do:

```sh
app/bin/tll bin/rails c
```

Which is shorter, albeit possibly a little more confusing to the user, since the lifecycle launcher _is_ run in the context of `./app` regardless.

stdstaff (approved):

README.md
scruti: Thanks @stdstaff , going to correct it since I think the value of a shorter command outplays the possible confusion.

PR #2674, opened 2022-11-10
Do not allow to introduce long ingredient names

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1661)

## Description

We are introducing a form validation to avoid very long text introduced as an ingredient name when creating an ingredient.

Since we moved from PDF formulations to requiring RP users to introduce their product ingredients manually, some users used the lack of length restriction on the name to paste the whole product formulation (sometimes thousands of characters) as an ingredient.

This validation will avoid that for future ingredients. What to do with the current ingredients/products violating this will be decided in the future. At the moment those product notifications should pass validations, hence not introducing a model validation together with the form one.

image

## Review apps

https://cosmetics-pr-2674-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2674-search-web.london.cloudapps.digital/


nasirkhanbeis (approved):

PR #2684, opened 2022-11-16
Product seach on submit

AndrewCowle:
## Review apps

https://cosmetics-pr-2684-submit-web.london.cloudapps.digital/

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


cosmetics-web/config/routes.rb
```diff
@@ -129,6 +129,7 @@
           get :resend # Ideally this would be a PATCH action, but doesn't play well with redirection after 2FA triggered by the link to resend with patch method.
         end
       end
+      resource :products, controller: "responsible_persons/search_products", path: "search-products", as: :search_products, only: %i[show]
```
slawosz, line 132: I wonder if it would make sense to have search nested as `Search` inside `responsible_person/:id/notifications/search`?

cosmetics-web/app/controllers/poison_centres/notifications_search_controller.rb
```diff
@@ -23,6 +23,7 @@ def search_notifications
       sort_by: @search_form.sort_by,
       match_similar: @search_form.match_similar,
       search_fields: @search_form.search_fields,
+      rp_id: nil,
```
slawosz, line 26: I would use full variable name: `responsible_person_id`. As in the other places.

cosmetics-web/config/routes.rb
AndrewCowle: I did consider this, but thought ahead - having `search_products` and `search_ingredients` under the responsible person folder structure seemed more readable.... happy to change if you prefer?

slawosz (approved):

PR #2697, opened 2022-11-25
Improve display of poisoning information

slawosz:
Cleaning up the way we display poisons information:
 * in case when notification had poison information as formulation document, we link this document
 * in case when notification has poisons listed (from 3rd October 2022), we are hiding row that displays formulation document
 * dont display `acute poisoning information` in case when its empty. Only `CPNP` imported notification had this field
 * don't clone `acute_poisoning_info` as its imposible to clean up this value


cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
```diff
@@ -160,7 +160,7 @@ def notification_summary_component_rows(component, include_shades: true, allow_e
           value: { text: component_special_applicator_name(component) },
         }
       end,
-      if current_user.can_view_product_ingredients?
+      if current_user.can_view_product_ingredients? && component.acute_poisoning_info.present?
         {
           key: { text: "Acute poisoning information" },
           value: { text: component.acute_poisoning_info.presence || "None" },
```
scruti, line 166:
```suggestion
          value: { text: component.acute_poisoning_info },
```

cosmetics-web/spec/support/matchers/capybara_matchers.rb
```diff
@@ -33,6 +33,16 @@ def failure_message
         "Could not find sibling <td> or <dd> containing ‘#{@value}’ within #{@key_element.find(:xpath, '..').native}"
       end
     end
+
+    def failure_message_when_negated
```
scruti, line 37: :question:  Is this used anywhere?

slawosz: Here:
```
expect(page).not_to have_summary_item(key: "Acute poisoning information", value: "Toxins")
```
When matcher fails, it calls `failure_message_when_negates` as part of the interface.

scruti (approved):

nasirkhanbeis (approved):
Tested this on the review app, looking good as expected. Darren to review from design perspective.

PR #2698, opened 2022-11-29
Search on submit bug

AndrewCowle:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1705

Added safe operator and spec tp date_helper.

## Review apps

https://cosmetics-pr-2698-submit-web.london.cloudapps.digital/

## Checklist:

### General testing
- [ ] Test without javascript
- [ ] Test on small screen
### Accessibility testing
- [ ] Works keyboard only


scruti (changes requested):
I've just noticed these exceptions are highlighting a bigger issue: 
Some of the search results have no completion date.

I have investigated some of the exceptions, and they are caused by a deleted notification being returned by the ElasticSearch index. And this shouldn't happen.

When I removed the `completed` scope from the result filters, it was after checking that the notification was deleted from the search results seconds after having been deleted from the index (and we tested that with Nasir). 

So I am surprised that the ES results are returning deleted notifications. 

Going to investigate/patch this issue with ES, while keeping these exceptions as a failsafe to let us know if the search results are still returning empty values or not.

Sorry for the annoyance. Working on it!

scruti (approved):

scruti (approved):

PR #2735, opened 2023-01-12
Fix display of fractional ing. concentration range

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1688)

## Description

When the ingredient range has a fractional value, display it properly. At the moment we are displaying "01", we want to display "0.1".

image

## Review apps

https://cosmetics-pr-2735-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2735-search-web.london.cloudapps.digital/


nasirkhanbeis (approved):
Tested this on the review app, looking good as expected.

PR #2738, opened 2023-01-17
Refactor user roles

slawosz:
This introduces the idea of `privilege`, something that a given user can do. The implementation is focused on readability and ease of use. Entry point to the feature is `Privileges::AbstractConcern`.


cosmetics-web/app/models/concerns/privileges/abstract_concern.rb
```diff
@@ -0,0 +1,23 @@
+# Main module for defining user priviliges
```
scruti, line 1:
```suggestion
# Main module for defining user privileges.
```

cosmetics-web/app/models/concerns/privileges/abstract_concern.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# All of the priviliges definition should go here for the clarity and then should be reimplemented
```
scruti, line 3:
```suggestion
# All of the privileges definitions should go here for clarity and then should be reimplemented
```

cosmetics-web/app/models/concerns/privileges/abstract_concern.rb
```diff
@@ -0,0 +1,23 @@
...
+# with concrete code in Submit or Search Concerns.
+#
+# It can be noticed that some of the priviliges are applicable only to one domain, and the design
```
scruti, line 6:
```suggestion
# It can be noticed that some of the privileges are applicable only to one domain, and the design
```

cosmetics-web/app/models/concerns/privileges/submit_concern.rb
```diff
@@ -0,0 +1,29 @@
...
+      false
+    end
+
+    def msa_user?
+      false
+    end
+
+    def opss_science_user?
+      false
+    end
```
scruti, line 15: What do you think about defaulting all these role methods as `false` in the `AbstractConcern` (move these methods there)?

This way, we override them in the `SearchConcern` based on the SearchUser role, while `SubmitConcern` don't have to worry about role definitions as it will include the defaults for any new role.

cosmetics-web/docs/ROLES.md
```diff
@@ -0,0 +1,3 @@
+# User Roles
+
+The role system is uncomplicated. Entry point for implementation is concern in models: `Priviliges::AbstractConcern`
```
scruti, line 3:
```suggestion
The role system is uncomplicated. The entry point for implementation is a concern in models: `Privileges::AbstractConcern`
```

scruti (approved):
Left a bunch of typo fixes/grammarly suggestions.
Also, a code change suggestion.

Anyways, LGTM!

nasirkhanbeis (approved):
Regression tested this on review app, looking good as expected.

PR #2745, opened 2023-01-20
Add Github Action to generate service ERD

scruti:
[Jira ticket](https://regulatorydelivery.atlassian.net/browse/COSBETA-1744)

With every update on master (or manual action trigger), the action will generate an Entity Relationship Diagram that will be available to download as an artifact.

[Example run with generated output](https://github.com/OfficeForProductSafetyAndStandards/cosmetic-product-notifications/actions/runs/3987437648)


.github/workflows/erd.yml
```diff
@@ -0,0 +1,114 @@
+name: Generate ERD
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+  pull_request: # For testing
```
scruti, line 8: remove this line before merging the branch
```suggestion
```

PR #2760, opened 2023-01-30
Add Responsible Persons and Notifications pages

rubenarakelyan:
JIRA links: https://regulatorydelivery.atlassian.net/browse/COSBETA-1737 and https://regulatorydelivery.atlassian.net/browse/COSBETA-1738

## Description

Adds two new pages under the ingredients list page:

* Responsible Persons - list of Responsible Persons who have created notifications of products containing a specified ingredient
* Notifications - list of notifications from a specified Responsible Person with products containing a specified ingredient

## Review apps

https://cosmetics-pr-2760-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2760-search-web.london.cloudapps.digital/

Note, this PR must be rebased and merged after https://github.com/OfficeForProductSafetyAndStandards/cosmetic-product-notifications/pull/2756 since it is based on that branch.

### DIFF
--- a/cosmetics-web/app/controllers/poison_centres/ingredients_controller.rb
+++ b/cosmetics-web/app/controllers/poison_centres/ingredients_controller.rb
@@ -1,5 +1,5 @@
 class PoisonCentres::IngredientsController < SearchApplicationController
-  PER_PAGE = 200
+  include ResponsiblePersonQueryConcern
 
   def index
     return redirect_to root_path unless helpers.can_view_ingredients_list?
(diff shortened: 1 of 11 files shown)


cosmetics-web/app/controllers/poison_centres/ingredients_controller.rb
```diff
@@ -12,6 +10,50 @@ def index
...
+    sort_by = case params[:sort_by]
+              when "most_notifications"
+                "total_notifications desc"
+              when "name_desc"
+                "responsible_persons.name desc"
+              else
+                "responsible_persons.name asc"
+              end
+
+    @responsible_persons = ResponsiblePerson
```
slawosz, line 29: Personally, only thing I would change is moving this search to the model/service object. I am not a fan of using controllers spec and operating on the strings matching not on data.

slawosz (approved):

cosmetics-web/spec/controllers/poison_centres/ingredients_controller_spec.rb
```diff
@@ -0,0 +1,136 @@
+require "rails_helper"
```
scruti, line 1: Since controller specs have been discouraged for a while in favour of request specs, and we have request specs for this area (`ingredient_list_spec` ), have you considered turning these controller specs into request specs?

cosmetics-web/spec/requests/ingredient_list_spec.rb
```diff
@@ -74,9 +74,111 @@
...
       include_examples "unique ingredient names"
 
-      it "orders ingredients by name from last to first" do
+      it "orders ingredients by name from first to last" do
         expect(response.body).to match(/Aqua.*NaCl.*Sodium/m)
       end
     end
   end
+
+  describe "GET #responsible_persons" do
```
scruti, line 83: Very good test coverage :clap:

cosmetics-web/spec/requests/ingredient_list_spec.rb
```diff
@@ -74,9 +74,111 @@
...
+      end
+    end
+
+    context "when sorting by name_asc" do
+      before do
+        get poison_centre_ingredients_responsible_persons_path(ingredient_inci_name: "Aqua", sort_by: "name_asc")
+      end
+
+      it "orders responsible persons by name from first to last" do
+        expect(response.body).to match(/#{ingredient2.component.responsible_person.name}.*#{ingredient4.component.responsible_person.name}/m)
```
slawosz, line 119: So what I would suggest, is to move queries to some method: `do_responsible_person_query`, and then having test like this:
```
it "orders responsible persons by name from first to last" do
  result = Something.do_responsible_person_query(sort: NAME_ASC, ingredient_name: "Aqua")
  
  expect(result).to eq [ingredient2.component.responsible_person.name, ingredient4.component.responsible_person.name]
end
```
where `Something` is a model or service object or something like this. I think personally I would create concern `ResponsiblePersonQueryConcern` or so.

slawosz: What I might miss, and its actually my point here, its about tests, not the queries in controller. In case of extending queries etc, those test are much harder to debug as they operate on text html instead of plain ruby arrays. So in case of using arrays, the test message will give array diff, when using controllers one have to read through lot of html.

PR #2801, opened 2023-02-20
Cleanup orphaned Trigger Question data

scruti:
## Description

When the service allowed importing/parsing data from the European CPNP service, some TriggerQuestions and TriggerQuestionElements were created without being associated with a component/notification or, in the case of the TQElements,  to any TQ.

As the ERD shows, if a TQElement is not associated with a TQ, or a TQ is not associated with a Component, there is no way they can be related with any notification, so it is orphaned invalid data.

### Production numbers:
- Orphaned Trigger Questions not associated with any component: 93.580
- Orphaned Trigger Question Elements not associated with any TQ:  12

### Solution
1. Following up on the `TODO` comment introduced in their model associations, removing the `optional` association parameter.
2. Sending an alert to Sentry anytime any of the orphaned TQ/TQE gets loaded from DB. This will help us identify if any of them gets loaded unexpectedly or if they're actually orphaned/not used.
3. Introducing a rake task to delete all those affected TQ & TQE from DB.

image

### DIFF
--- a/cosmetics-web/app/models/trigger_question.rb
+++ b/cosmetics-web/app/models/trigger_question.rb
@@ -8,6 +8,13 @@ class TriggerQuestion < ApplicationRecord
 
   validates :applicable, inclusion: { in: [true, false] }, on: :update
 
+  # TODO: remove this after executing the rake task cleaning up dangling trigger questions
+  after_find do |trigger_question|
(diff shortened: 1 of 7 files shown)


scruti:
Thanks @slorek . Reworded it in code/comments before release.

PR #2808, opened 2023-02-22
Display matched ingredients in search result list

slawosz:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1750
Displays matched ingredients:
image
On the ingredients results list:
image


cosmetics-web/app/models/notification.rb
```diff
@@ -315,6 +315,10 @@ def cloned?
     source_notification.present?
   end
 
+  def ingredients
+    components.map(&:ingredients).flatten
```
scruti, line 319: To avoid N+1 queries for multicomponent notifications:
```suggestion
    components.includes(:ingredients).map(&:ingredients).flatten
```

cosmetics-web/spec/models/notification_spec.rb
```diff
@@ -609,4 +609,26 @@
       end
     end
   end
+
+  describe "#ingredients" do
+    # create notification
```
scruti, line 614:
```suggestion
```

cosmetics-web/app/views/poison_centres/ingredients_search/_result_row.html.erb
```diff
@@ -1,6 +1,6 @@
 <%
-  cell_headers = "item-#{index} meta-#{index}"
-%>
+    cell_headers = "item-#{index} meta-#{index}"
+  %>
```
scruti, line 3:
```suggestion
%>
```

cosmetics-web/app/views/poison_centres/ingredients_search/_result_row.html.erb
```diff
@@ -9,9 +9,9 @@
     </th>
     <th id="item-<%= index %>" colspan="3" scope="colgroup" class="govuk-table__header">
       <%= link_to notification.product_name,
-                  poison_centre_notification_path(notification, form => search_params, :back_to => back_to, page: params[:page]),
-                  class: "govuk-link govuk-link--no-visited-state" %>
-    </th>
+        poison_centre_notification_path(notification, form => search_params, :back_to => back_to, page: params[:page]),
+        class: "govuk-link govuk-link--no-visited-state" %>
+  </th>
```
scruti, line 14:
```suggestion
    </th>
```

cosmetics-web/app/views/poison_centres/ingredients_search/_result_row.html.erb
```diff
@@ -27,4 +27,18 @@
       <%= notification.responsible_person.name %>
     </td>
   </tr>
+  <% matches = PoisonCentres::IngredientNameMatcher.match(@search_form.q, notification) %>
```
scruti, line 30: Since we are calling the service object from a view partial instead of from the controller, and the reason seems to be the need to be called for each one of the result notifications for ingredients match display purposes... this `IngredientNameMatcher` seems to be a helper method disguised as a service object.

slawosz: We can add a helper method, but it would be very simple delegation and would not bring any value in my opinion. The service object is not a helper method. I consider (and I think its wide approach) a helper in Rails as a way to hide complex UI operations.

scruti: So `IngredientNameMatcher` code doesn't match the purpose of a helper, nor matches the purpose of Service Objects either, as it doesn't map to any business action/logic. Having a 'service object' called from a view partial for each one of the result elements is a code smell that highlights this.

So who should respond to "return the matching ingredients to this query"?

I think `Notification`

We could call: `Notification#matching_ingredients(query)` (or array of words or whatever as param) and get a list of matches per notification.

That is it, a method. No need for a new Service Object, helper,etc. A plain simple method in Notification, answering what matching ingredients contains

rubenarakelyan (approved):

PR #2836, opened 2023-03-13
Archive notifications

rubenarakelyan:
JIRA tickets: https://regulatorydelivery.atlassian.net/browse/COSBETA-1449 and https://regulatorydelivery.atlassian.net/browse/COSBETA-1831

## Description

Adds the ability to archive and unarchive notifications from both the index and individual notification pages.

Also prepares the search index to allow filtering by archived status when searching for notifications.

## Review apps

https://cosmetics-pr-2836-submit-web.london.cloudapps.digital/
https://cosmetics-pr-2836-search-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
@@ -1,20 +1,38 @@
 class ResponsiblePersons::NotificationsController < SubmitApplicationController
   before_action :set_responsible_person
   before_action :validate_responsible_person
-  before_action :set_notification, only: %i[show]
 
   def index
     @registered_notifications = get_registered_notifications(20)
     respond_to do |format|
       format.html
       format.csv do
-        @notifications = NotificationsDecorator.new(@responsible_person.notifications.completed.order(notification_complete_at: :desc))
+        @notifications = NotificationsDecorator.new(@registered_notifications.except(:limit, :offset))
         render csv: @notifications, filename: "all-notifications-#{Time.zone.now.to_fs(:db)}"
       end
     end
   end
 
-  def show; end
+  def archived
+    @registered_notifications = get_registered_archived_notifications(20)
+    respond_to do |format|
+      format.html
+      format.csv do
+        @notifications = NotificationsDecorator.new(@registered_notifications.except(:limit, :offset))
+        render csv: @notifications, filename: "all-archived-notifications-#{Time.zone.now.to_fs(:db)}"
+      end
+    end
+  end
+
+  def show
+    @notification = Notification.where.not(state: :deleted).find_by!(reference_number: params[:reference_number])
+    authorize @notification, policy_class: ResponsiblePersonNotificationPolicy
+    @back_link = if @notification.archived?
+                   responsible_person_archived_notifications_path(@responsible_person, page: params[:page])
+                 else
+                   responsible_person_notifications_path(@responsible_person, page: params[:page])
+                 end
+  end
 
   def new
     @notification = @responsible_person.notifications.new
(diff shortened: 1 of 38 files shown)


cosmetics-web/spec/controllers/responsible_persons/notifications_controller_spec.rb
```diff
@@ -125,6 +157,36 @@
...
+      expect(response).to redirect_to("/responsible_persons/#{responsible_person.id}/notifications")
+    end
+  end
+
+  describe "GET /unarchive" do
+    let(:archived_notification) { create(:registered_notification, :archived, responsible_person:) }
+
+    it "unarchives the notification and redirects to the index page" do
+      get :unarchive, params: { responsible_person_id: responsible_person.id, notification_reference_number: archived_notification.reference_number }
+      get :index, params: { responsible_person_id: responsible_person.id, reference_number: archived_notification.reference_number }
```
slawosz, line 180: If `unarchive` redirects, do we need `get :index` here?

slawosz (approved):
Except 2 comments in tests, looks good to me.

cosmetics-web/spec/controllers/responsible_persons/notifications_controller_spec.rb
rubenarakelyan: It's so I can use `assigns(:registered_notifications)` - a redirect doesn't expose this

scruti:
Hi @rubenarakelyan. 
I have some concerns regarding how we are registering the `archived` concept for the notifications.
These changes using a DB flag, seem to move away from the current state machine approach, where `archived` would be a new notification state we transition to/from, and where we could put a clear restriction on where can be transitioned from (EG: only `notification_complete` state can be transitioned to `archived`). 

I would be happy to hold it as a flag if `archived` would be something complementary to the current notification state and could be applied to multiple states, but from what I understand here, it seems like a new different state for the notification itself.

Besides consistency, having `archived` as a state would save us from having to combine scopes like:
`Notification.non_archived.completed` and `Notification.archived.completed` into a unique `Notification.completed` or `Notification.archived` .

Same with queries like ` Notification.non_archived.where.not(state: :deleted)`.  "We want neither archived nor deleted notifications".  The fact that underneath the `archived` is a boolean flag while for `deleted`, we are checking the state seems odd.

Let me know your thoughts, I may be missing some context!

cosmetics-web/app/controllers/responsible_persons/notifications_controller.rb
```diff
@@ -14,7 +13,21 @@ def index
     end
   end
 
-  def show; end
+  def archived
+    @registered_notifications = get_registered_archived_notifications(20)
+    respond_to do |format|
+      format.html
+      format.csv do
+        @notifications = NotificationsDecorator.new(@responsible_person.notifications.archived.order(notification_complete_at: :desc))
```
scruti, line 21: Suggestion:

Noticed that the CSV format doesn't use `@registered_notifications` and that we retrieve the notifications twice:
- Once for `@registered_notifications` with pagination.
- Once with the same query but without paginating for the decorated `@notifications`.
[Kaminari offers a way to unscope a paginated query](https://regulatorydelivery.atlassian.net/browse/COSBETA-1861) that looks perfect for this use case


```suggestion
        @notifications = NotificationsDecorator.new(@registered_notifications.except(:limit, :offset))
```

cosmetics-web/app/models/notification.rb
```diff
@@ -105,6 +103,7 @@ def self.completed
       indexes :searchable_ingredients, type: "text"
       indexes :created_at, type: "date"
       indexes :notification_complete_at, type: "date", format: "strict_date_optional_time"
+      indexes :state, type: "text"
```
scruti, line 106: I cannot believe we didn't have an index here :see_no_evil: . Thanks for adding it!

scruti (approved):
LGTM! Left a small code suggestion.

nasirkhanbeis (approved):
Tested this on the review app and also latest fixes, looking good as expected.

PR #2851, opened 2023-03-21
Remove date restriction from nanomaterials journey

tobyprivett:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1822

## Description
Here we update the `NanomaterialNotification#notified_to_eu_on=` method to assign a null date if the EU has **not** been notified. This might happen if the user selects 'Yes', enters a date, then changes their answer to 'No'

Previously, this raised an error and the user would have to go back and manually remove the date.

I can see some issues with the existing code:
- We use a [complex method to assign the date](https://github.com/OfficeForProductSafetyAndStandards/cosmetic-product-notifications/blob/5860e464171ccb314a8bef4bb59d88451efeddb6/cosmetics-web/app/models/nanomaterial_notification.rb#L30). Rails should be able to handle good/bad dates.
- The  ActiveRecord `#notified_to_eu_on=` is overridden to prevent nil values. Yet, we assign a nil value if the date can't be parsed. 
- The proposed implementation in this PR will only work if `#eu_notified` is assigned before `#notified_to_eu_on`

## Review apps

https://cosmetics-pr-2851-submit-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/models/nanomaterial_notification.rb
+++ b/cosmetics-web/app/models/nanomaterial_notification.rb
@@ -44,7 +44,7 @@ class NanomaterialNotification < ApplicationRecord
       date = nil
     end
 
-    self[:notified_to_eu_on] = date
+    self[:notified_to_eu_on] = eu_notified? ? date : nil
   end
 
   def submit!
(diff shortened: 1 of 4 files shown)


cosmetics-web/app/models/nanomaterial_notification.rb
```diff
@@ -44,7 +42,7 @@ def notified_to_eu_on=(notified_to_eu_on)
       date = nil
     end
 
-    self[:notified_to_eu_on] = date
+    self[:notified_to_eu_on] = eu_notified? ? date : nil
```
scruti, line 45: Suggestion: Ass this condition as first check on the `if/else` so we avoid to parse/assign the date to then nullify it afterwards.
Something like:
```ruby
if !eu_notified?
  date = nil
elsif notified_to_eu_on.is_a?(Date)
  date = notified_to_eu_on
elsif ...
```

PR #2868, opened 2023-03-28
Remove default responses from nanomaterials

tobyprivett:
https://regulatorydelivery.atlassian.net/browse/COSBETA-1859

## Description
If a notification does *not* contain nano materials, then we remove the
`contains_nanomaterials` question from the notification's `routing_questions_answers`

Here we also fix an incorrect validation error message:
'mulit-item kit' was displayed, instead of 'nanomaterial'

Testing note: You will need to create a notification manually, and clone that.
If you clone a notification that was created with seed data, the 
`new_notification.routing_questions_answers` attribute won't be set properly,

## Review apps

https://cosmetics-pr-2868-submit-web.london.cloudapps.digital/

### DIFF
--- a/cosmetics-web/app/services/notification_cloner/base.rb
+++ b/cosmetics-web/app/services/notification_cloner/base.rb
@@ -19,6 +19,11 @@ module NotificationCloner
 
     def self.clone_notification(old_notification, new_notification)
       new_notification = clone_model(old_notification, object_to_clone_to: new_notification)
+
+      if new_notification.nano_materials.empty? && new_notification.routing_questions_answers
+        new_notification.routing_questions_answers[:contains_nanomaterials] = nil
+      end
+
       new_notification.save!
       new_notification
     end
--- a/cosmetics-web/config/locales/en.yml
+++ b/cosmetics-web/config/locales/en.yml
@@ -317,7 +317,7 @@ en:
         responsible_persons/notifications/product/contains_nanomaterials_form:
           attributes:
             contains_nanomaterials:
-              inclusion: "Select yes if the product is a multi-item kit, no if its single item"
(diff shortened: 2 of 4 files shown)


cosmetics-web/app/services/notification_cloner/base.rb
```diff
@@ -19,6 +19,11 @@ def self.clone(old_notification, new_notification)
 
     def self.clone_notification(old_notification, new_notification)
       new_notification = clone_model(old_notification, object_to_clone_to: new_notification)
+
+      if new_notification.nano_materials.empty? && new_notification.routing_questions_answers
+        new_notification.routing_questions_answers[:contains_nanomaterials] = nil
```
rubenarakelyan, line 24: We should always set this to `nil` since we always want the user to choose.

rubenarakelyan (changes requested):

cosmetics-web/app/services/notification_cloner/base.rb
tobyprivett: In this instance, they want this to only apply when the are no nanomaterials:
"Where my original notification did not contain a nanomaterial, the radio buttons will not default to my original response, and I will be required to select “Yes “ or “No” in answer to this question"
and 
"When I access SCPN and I clone a previously submitted notification for my organisation which contained a nanomaterial, current behaviour and functionality will remain as-is"
@rubenarakelyan

rubenarakelyan: Ah, I must have misunderstood

rubenarakelyan (approved):

nasirkhanbeis (approved):
Tested this along with Andre and happy. working as expected.

PR #3311, opened 2024-04-25
Remove cypress test to fix deployments

alan-at-work:
## Description
Remove cypress tests while we investigate why they are not working as they were, suspected removal of a PAT from previous developer.


alan-at-work:
No longer

PR #3317, opened 2024-05-07
[PSD-2553] - fixed the Nil class error on the object responsible_person 

meeravnagesh:
the responsible person Name shown in the Nav bar of the Project screen was throwing Nil class error, which i have fixed using the Nil check


alan-at-work (approved):
Nicely done

tahir-khalid (approved):

PR #3381, opened 2024-08-19
fix/PSD-2690_no_route_error

tahir-khalid:
## Description
Attempting to resolve `responsible_person` nil scenario when loading my account area navbar.

https://regulatorydelivery.atlassian.net/browse/PSD-2690


tahir-khalid:
Bad repo closing.

PR #3397, opened 2024-08-30
fix/PSD-2690_actionview_temp_error

tahir-khalid:
## Description

Resolved no route found error for notifications controller.

PSD-2690

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
@@ -17,6 +17,10 @@ class ResponsiblePersons::ContactPersonsController < SubmitApplicationController
       # we don't want to switch to new RP
       # See also ResponsiblePerspon::AccountWizardController
       set_current_responsible_person_from_previous
+      if current_responsible_person.nil?
+        current_responsible_person
+      end
+
       redirect_to responsible_person_path(current_responsible_person), confirmation: create_successful_message
     else
       render :new
@@ -47,7 +51,7 @@ class ResponsiblePersons::ContactPersonsController < SubmitApplicationController
 private
 
   def set_responsible_person
-    @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
+    @responsible_person = helpers.set_responsible_person
     authorize @responsible_person, :update?
   end
 
--- a/cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
@@ -6,7 +6,7 @@ class ResponsiblePersons::ContactPersonsController < SubmitApplicationController
   }.freeze
 
   skip_before_action :create_or_join_responsible_person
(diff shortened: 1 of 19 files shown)


cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
```diff
@@ -17,6 +17,10 @@ def create
       # we don't want to switch to new RP
       # See also ResponsiblePerspon::AccountWizardController
       set_current_responsible_person_from_previous
+      if current_responsible_person.nil?
```
meeravnagesh, line 20: the code inside if condition , does not do anything , is that a method

cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
```diff
@@ -409,6 +409,15 @@ def notification_summary_label_image_link(image, responsible_person, notificatio
     end
   end
 
+  def set_responsible_person
```
meeravnagesh, line 412: the same of the mthod should be get_responsible_person

cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
```diff
@@ -409,6 +409,15 @@ def notification_summary_label_image_link(image, responsible_person, notificatio
...
+    @responsible_person =
+      if session[:current_responsible_person] && session[:current_responsible_person][:id] == params[:responsible_person_id]
+        ResponsiblePerson.new(session[:current_responsible_person])
```
meeravnagesh, line 415: session[:current_r_p] is already a object , is there a reason why we are creating a new object here?

cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
```diff
@@ -47,7 +51,7 @@ def update
 private
 
   def set_responsible_person
-    @responsible_person = ResponsiblePerson.find(params[:responsible_person_id])
+    @responsible_person = helpers.set_responsible_person
```
meeravnagesh, line 54: the name of the helper is deceiving as it says set but its assigned to a variable

cosmetics-web/app/controllers/responsible_persons/contact_persons_controller.rb
tahir-khalid: It does from a certain flow and there are a few ways to hit this code if I am right but you could also be correct will make a task to test this and include or exclude that code thanks.

cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
alan-at-work: I would agree keep this in the controllers, not need to move it out really.

cosmetics-web/app/helpers/responsible_persons/notifications_helper.rb
alan-at-work: if we remove the method this comment is void ;)

PR #3485, opened 2024-12-09
fix/PSD-3947-SCPN-url_generation_error

mirotriad:
# Fix component routing issue and refactor tests

## Description
This PR addresses a production error where the notification component build path was failing due to missing component IDs. It also includes a comprehensive refactor of the associated tests to improve maintainability and follow best practices.

## Changes
- Fixed component routing in `ProductController` by ensuring component creation with `first_or_create!`
- Refactored tests to follow single expectation pattern
- Added proper test coverage for all notification states (nanomaterials, multi-component, single component)
- Updated multi-component tests to use real data setup instead of stubs

## Related Issues
Fixes Sentry: `ActionController::UrlGenerationError` in `ResponsiblePersons::Notifications::ProductController#show`

## Testing
- All existing functionality remains unchanged
- Added new test cases to cover edge cases
- All tests passing
- RuboCop compliant

### DIFF
--- a/cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
+++ b/cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
@@ -26,7 +26,8 @@ class ResponsiblePersons::Notifications::ProductController < SubmitApplicationCo
     case step
     when :completed
       @notification.set_state_on_product_wizard_completed!
-      render template: "responsible_persons/notifications/task_completed", locals: { continue_path: }
+      @continue_path = continue_path
+      render template: "responsible_persons/notifications/task_completed", locals: { continue_path: continue_path }
     when :add_product_image
       @clone_image_job = NotificationCloner::JobTracker.new(@notification.id) if @notification.cloned?
       render_wizard
@@ -67,7 +68,8 @@ private
     elsif @notification.multi_component?
       new_responsible_person_notification_product_kit_path(@notification.responsible_person, @notification)
     else
-      new_responsible_person_notification_component_build_path(@notification.responsible_person, @notification, @notification.components.first)
+      component = @notification.components.first_or_create!
+      new_responsible_person_notification_component_build_path(@notification.responsible_person, @notification, component)
     end
   end
 
--- a/cosmetics-web/spec/controllers/responsible_persons/notifications/product_controller_spec.rb
+++ b/cosmetics-web/spec/controllers/responsible_persons/notifications/product_controller_spec.rb
@@ -52,12 +52,88 @@ RSpec.describe ResponsiblePersons::Notifications::ProductController, :with_stubb
     end
 
     context "when the notification is already submitted" do
-      subject(:request) { get(:show, params: params.merge({ id: "add_internal_reference" })) }
-
       let(:notification) { create(:registered_notification, responsible_person:) }
 
       it "redirects to the notifications page" do
-        expect(request).to redirect_to(responsible_person_notification_path(responsible_person, notification))
+        get(:show, params: params.merge({ id: "add_internal_reference" }))
+        expect(response).to redirect_to(responsible_person_notification_path(responsible_person, notification))
+      end
+    end
+
+    context "when on the completed step" do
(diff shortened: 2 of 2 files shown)


cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
```diff
@@ -26,7 +26,8 @@ def show
     case step
     when :completed
       @notification.set_state_on_product_wizard_completed!
-      render template: "responsible_persons/notifications/task_completed", locals: { continue_path: }
+      @continue_path = continue_path
```
alan-at-work, line 29: I don't think `@continue_path` is needed in this case. The local variable `continue_path` is already available. And this is not passed to the view ;)

cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
```diff
@@ -26,7 +26,8 @@ def show
...
+      render template: "responsible_persons/notifications/task_completed", locals: { continue_path: continue_path }
```
tahir-khalid, line 30: The template path can this be a variable or a helper function to return that value from a constant if need be.

tahir-khalid, line 30: See previous comment could be something `def responsible_persons_notif_task_completed_path...end`

cosmetics-web/spec/controllers/responsible_persons/notifications/product_controller_spec.rb
```diff
@@ -52,12 +52,88 @@
...
+      context "when notification has nanomaterials" do
+        let(:nano_material) { create(:nano_material, notification: notification) }
+
+        before do
+          notification.nano_materials << nano_material
+        end
+
+        it "renders task completed template" do
+          get(:show, params: params_with_completed)
+          expect(response).to render_template("responsible_persons/notifications/task_completed")
```
tahir-khalid, line 75: Can this be returned from a private function (see previous comment for example) or a let so its not hardcoded and better readability thanks.

cosmetics-web/spec/controllers/responsible_persons/notifications/product_controller_spec.rb
```diff
@@ -52,12 +52,88 @@
...
+      context "when notification is multi-component" do
+        before do
+          notification.components.destroy_all
+          notification.update_column(:state, "ready_for_components")
+          create_list(:component, 2, notification: notification)
+        end
+
+        it "renders task completed template" do
+          get(:show, params: params_with_completed)
+          expect(response).to render_template("responsible_persons/notifications/task_completed")
```
tahir-khalid, line 93: Same as above can you apply that refactoring suggestion to the rest if you decide to go that path thanks.

cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
mirotriad: continue_path was being passed as a local variable to the template but not being assigned as an instance variable, so assigns(:continue_path) was returning nil

cosmetics-web/app/controllers/responsible_persons/notifications/product_controller.rb
mirotriad: I think this is not needed - the path is used only once, so for each render, for example, if there are 5 paths, we will create 5 methods, which is an overhead don't you think?

alan-at-work (approved):

cosmetics-web/spec/controllers/responsible_persons/notifications/product_controller_spec.rb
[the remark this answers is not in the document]
mirotriad: Thanks for the suggestion @alan-at-work! I've split up the specs and added the continue_path assertions. This should give us better test coverage of the navigation flow while keeping the specs clean and focused. Let me know if you'd like any adjustments to this approach.

alan-at-work (approved):

PR #3523, opened 2025-01-24
PSD-4522 - create type and role migration to roles task

alan-at-work:
## Description
We are now looking at populating the roles table for each user. 

For this we will need to run the following command. This will populate the roles table based on the `legacy_type` and `legacy_role` fields in the table. 

```ruby
rails map_legacy_roles_to_rolify:migrate
```
Once we have run this we will need to go onto the server to make sure the user has the right roles. 
So we would check a random user, and then see the roles they have. Then check using the `has_role?(:role_name)` query.

These should match up.


meeravnagesh (approved):
tested in local , working fine , and created the array of roles based on the user

PR #3528, opened 2025-01-29
PSD-4519 - Update to the codebase to use new roles setup

alan-at-work:
## Description
Working on the codebase to move across to use the new roles setup. This includes an update to the submit user where they now get the new `submit_user` role on creation while still using the old user types.


meeravnagesh (approved):
I have pulled the branch in local , tested the submit user and submitted a notification and i checked for opss general user and science user in the search app

PR #3534, opened 2025-02-05
PSD-4621 - Fix for roles not being assigned to SeachUsers

alan-at-work:
## Description
Fix for roles not being assigned to SeachUsers when they are created.

### DIFF
--- a/cosmetics-web/app/services/invite_search_user.rb
+++ b/cosmetics-web/app/services/invite_search_user.rb
@@ -22,12 +22,11 @@ private
   end
 
   def create_user
-    SearchUser.find_or_create_by!(email: email) do |user|
-      user.name = name
-      user.skip_password_validation = true
-      user.invite = true
-      user.save!
-      user.add_role(role)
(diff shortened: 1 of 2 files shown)


cosmetics-web/app/services/invite_search_user.rb
```diff
@@ -22,12 +22,11 @@ def email_taken_by_other_user_type?
...
-      user.name = name
-      user.skip_password_validation = true
-      user.invite = true
-      user.save!
-      user.add_role(role)
+    SearchUser.find_or_create_by!(email: email) do |new_user|
+      new_user.name = name
+      new_user.skip_password_validation = true
+      new_user.invite = true
+      new_user.add_role(role)
```
meeravnagesh, line 29: there is no new_user.save!?

alan-at-work: nope we are using `SearchUser.find_or_create_by!`

meeravnagesh (approved):

PR #3583, opened 2025-03-31
fix-sed-command

mirotriad:
# Fix production deployment sed command issue

Fix the failing `sed` command in the production deployment workflow by removing BSD-style empty string argument that's incompatible with GNU sed in the deployment environment.

```bash
sed -i 's/      - cosmetics-puma-env/      - cosmetics-puma-env\n      - antivirus-auth-env/g' ./cosmetics-web/manifest.yml
```
```


acodeninja (approved):
Good catch.

PR #3618, opened 2025-05-15
PSD-5541-simplify-2fa-feature-flag

mirotriad:
### Summary
This PR simplifies the two-factor authentication feature flag implementation by:

1. Using a consistent feature flag name `two_factor_authentication` to align with other services
2. Removing the initializer that caused database connection issues during deployment
3. Updating documentation to reflect these changes

### Why this change?
The previous implementation caused issues during deployment because it tried to connect to the database during asset precompilation, before the database was available.

With this new approach:
- 2FA is enabled by default (more secure) when the flag is enabled
- No database connection is needed during initialization
- Code is simpler and more maintainable
- Deployment issues are resolved
- Implementation is consistent with the Product Safety Database service

### Testing
- When the `two_factor_authentication` flag is enabled (default), 2FA works normally
- When the `two_factor_authentication` flag is disabled, users can skip 2FA

### How to test
- Log in to the application with the flag enabled - 2FA should be required
- Disable the flag (`Flipper.disable(:two_factor_authentication)`) - users should be able to log in without 2FA


meeravnagesh (changes requested):
check the migration

